PR #2, opened 2021-02-26
Remove unused dependencies

cjalmeida:
This PR removes unneeded runtime dependencies (six, numpy)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,9 @@
 from setuptools import find_packages, setup, Extension
 
-requirements = ["numpy>=1.16", "six"]
 
 setup(
     packages=find_packages(exclude=["docs"]),
     python_requires=">=3.6",
-    install_requires=requirements,
-    ext_modules=[Extension("plum.function", ["plum/function.py"])],
+    install_requires=[],
     include_package_data=True,
 )
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ venv
 *.egg-info
 dist
 pip-wheel-metadata
+build
 
 # Documentation and coverage
 docs/_build


wesselb:
Hey @cjalmeida! Thank you for your interest in the package. This PR looks good to me. You're right that both `six` and `numpy` aren't used, so can be removed as dependencies. The PR, however, also removes the compiled component, which is critical for minimising the overhead that plum brings. Shall we keep that in? I've made a suggestion.

setup.py
```diff
@@ -1,11 +1,9 @@
 from setuptools import find_packages, setup, Extension
 
-requirements = ["numpy>=1.16", "six"]
 
 setup(
     packages=find_packages(exclude=["docs"]),
     python_requires=">=3.6",
-    install_requires=requirements,
-    ext_modules=[Extension("plum.function", ["plum/function.py"])],
+    install_requires=[],
```
wesselb, line 7:
```suggestion
    ext_modules=[Extension("plum.function", ["plum/function.py"])],
```

gc-ss:
I would be interested in getting this branch merged in.

Could we add numpy as a dev dependency for test and benchmark in the poetry file?

```
[tool.poetry.dev-dependencies]
numpy = "^1.20.2"
```

jph00:
@wesselb we're thinking about using your cool-looking lib for fastcore. But we can't add a dep on numpy. Any chance you could merge this PR or even just make the edit yourself? Or would you like me to create a new clean PR?

wesselb:
@gc-ss, sorry for not having acted on this timely. I completely forgot about this.

@jph00, certainly! 

I've released a new version, `1.1.0`, which drops the dependencies on `numpy` and `six`.

jph00:
Thanks a lot!

wesselb:
No problem at all!

gc-ss:
> I've released a new version, `1.1.0`, which drops the dependencies on `numpy` and `six`.

Appreciate it Wessel!

Issue #9, opened 2021-05-19
Parametric classes usage


PhilipVinc:
Hi,
I would like to use parametric classes.
I'd like to implement something akin to (Julia)
```julia
struct DoubledSpace{T} where T
   space::T
end

double_int = DoubledSpace(1)
double_float = DoubledSpace(1.0)

# and dispatch on the parameter
size(a::DoubledSpace{Int}) = ...
size(a::DoubledSpace{Float64}) = ...
```

(The parameter class will be a custom class, but anyhow...)
I don't care about dispatch cost as all this is inside jax jitted function.

Anyhow, I tried to implement this by doing

```python
from plum impor dispatch, parametric

@parametric
class doubledspace:
   def __init__(self, space):
      print("initializing doubledspace")
      self.space = space

a = doubledspace(1.0)
```

but the init never gets called.

I also tried to dive into your code in `plum/parametric.py` and instrument it a bit:
```python
def parametric(Class):
    """A decorator for parametric classes."""
    subclasses = {}

    def __new__(cls, *ps):
        print(f"executing {cls.__name__}.__new__(*{ps})")
        # Only create new subclass if it doesn't exist already.
        print(f"  SubClasses are:")
        for p in subclasses:
            print(f"  \t{p}")
        print("  ---------------")
        if ps not in subclasses:
            print(f"  it is not in sublcasses")

            def __new__(cls, *args, **kw_args):
                print(f"    Calling the original {Class.__name__}.__new__(*{args}, **{kwargs}) but dropping them")
                return Class.__new__(cls)

            # Create subclass.
            name = Class.__name__ + "[" + ",".join(str(p) for p in ps) + "]"
            SubClass = type.__new__(
                CovariantMeta,
                name,
                (ParametricClass,),
                {"__new__": __new__, "_is_parametric": True},
            )
            SubClass._type_parameter = ps[0] if len(ps) == 1 else ps
            SubClass.__module__ = Class.__module__

            print(f"  Creating a new subclass called {name}")
            # Attempt to correct docstring.
            try:
                SubClass.__doc__ = Class.__doc__
            except AttributeError:  # pragma: no cover
                pass

            subclasses[ps] = SubClass

        print(f"  returning the subclass {subclasses[ps]}")
        return subclasses[ps]
```

and the output of the snippet before is
```python
➜ python -i test.py                    
now constructing
executing doubledspace.__new__(*(1,))
  SubClasses are:
  ---------------
  it is not in sublcasses
[...]

PR #11, opened 2021-06-18
[WIP/RFC] Make Parametric class constructor more like Julia

PhilipVinc:
I am putting this here to get some feedback...
This is a tentative implementation of what I described in #9

the objective was that given the parametric class

```python
@plum.parametric
class Test:
    def __init__(self, a, *, b=3):
        self.a = a
        self.b = b
```

I would like to replace 
```python
T = Test[list]
obj = T([1,2,3], b=4)
```
with
```python
obj = Test([1,2,3], b=4)
type(obj)  # __main__.Test[<class 'list'>]
```

The syntax `Test[par]` still works.

--

I achieve this by introducing a new MetaClass, that I call `ParametricTypeMeta`, which should be only used for parametric objects, which overrides `__getitem__` to call`__new__` instead of `__call__`, and I override `__call__` to only instantiate the object if it's a concrete subtype.

What would you think about this?

### DIFF
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -30,8 +30,42 @@ log = logging.getLogger(__name__)
 
 _dispatch = Dispatcher()
 
+class ParametricTypeMeta(TypeMeta):
+    """Types can also be instantiated with indexing."""
 
-class CovariantMeta(TypeMeta):
+    def __getitem__(self, p):
+        # Type[Par1, Par2] creates a new parametric type
+
+        cls = self
+        if isinstance(p, tuple):
+            return cls.__new__(cls, *p)
+        else:
+            # cls = p
+            return cls.__new__(cls, p)
+
+    def __call__(cls, *args, **kwargs):
+        # Type(arg1, arg2, kwargs) will first construct the
+        # parametric subtype T = Type[type(arg1), type(arg2)]
+        # and then call the equivalent of T(arg1, arg2, **kwargs)
+
+        if not cls.is_concrete:
+            argsT = tuple(type(arg) for arg in args)
+            T = cls[argsT]
+        else:
+            T = cls
+
+        # this is the standard type.__call__ implementation
+        obj = T.__new__(T, *args, **kwargs)
+        if obj is not None and isinstance(obj, T) and hasattr(obj, '__init__'):
+            obj.__init__(*args, **kwargs)
+        return obj
+
+    @property
+    def is_concrete(cls):
+        return hasattr(cls, "_is_parametric")
+
+
+class CovariantMeta(ParametricTypeMeta):
     """A metaclass that implements *covariance* of parametric types."""
 
     def __subclasscheck__(self, subclass):
@@ -99,7 +133,7 @@ def parametric(Class):
         return subclasses[ps]
 
     # Create parametric class.
-    ParametricClass = TypeMeta(Class.__name__, (Class,), {"__new__": __new__})
+    ParametricClass = ParametricTypeMeta(Class.__name__, (Class,), {"__new__": __new__})
     ParametricClass.__module__ = Class.__module__
 
     # Attempt to correct docstring.
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -39,9 +39,12 @@ class ParametricTypeMeta(TypeMeta):
     `Type[type(Arg1), type(Arg2)](Arg1, Arg2, **kw_args)`.
     """
 
-    def __getitem__(self, *ps):
+    def __getitem__(self, p):
         if not self.is_concrete:
-            return self.__new__(self, *ps)
+            if isinstance(p, tuple):
+                return self.__new__(self, *p)
+            else:
+                return self.__new__(self, p)
         else:
             raise TypeError("Cannot specify type parameters. This type is concrete.")
 
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -31,17 +31,12 @@ log = logging.getLogger(__name__)
 _dispatch = Dispatcher()
 
 class ParametricTypeMeta(TypeMeta):
-    """Types can also be instantiated with indexing."""
+    """Parametric Types must be instantiated with indexing."""
 
-    def __getitem__(self, p):
+    def __getitem__(self, *p):
         # Type[Par1, Par2] creates a new parametric type
 
-        cls = self
-        if isinstance(p, tuple):
-            return cls.__new__(cls, *p)
-        else:
-            # cls = p
-            return cls.__new__(cls, p)
+        return self.__new__(self, *p)
 
     def __call__(cls, *args, **kwargs):
         # Type(arg1, arg2, kwargs) will first construct the
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -210,7 +210,7 @@ class List(ComparableType):
         return f"List[{self._el_type}]"
 
     def get_types(self):
-        return (_ParametricList(self._el_type),)
+        return (_ParametricList[self._el_type],)
 
     @property
     def parametric(self):
@@ -246,7 +246,7 @@ class Tuple(ComparableType):
         return f'Tuple[{", ".join(map(str, self._el_types))}]'
 
     def get_types(self):
-        return (_ParametricTuple(*self._el_types),)
+        return (_ParametricTuple[self._el_types],)
 
     @property
     def parametric(self):
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -32,7 +32,7 @@ _dispatch = Dispatcher()
 
 
 class ParametricTypeMeta(TypeMeta):
-    """Parametric Types can be instantiated with indexing.
+    """Parametric types can be instantiated with indexing.
 
     A concrete parametric type can be instantiated by calling `Type[Par1, Par2]`.
     If `Type(Arg1, Arg2, **kw_args)` is called, this returns
@@ -46,19 +46,19 @@ class ParametricTypeMeta(TypeMeta):
             raise TypeError("Cannot specify type parameters. This type is concrete.")
 
     def __call__(cls, *args, **kw_args):
-        # Type(arg1, arg2, kw_args) will first construct the
-        # parametric subtype T = Type[type(arg1), type(arg2)]
-        # and then call the equivalent of T(arg1, arg2, **kw_args)
+        # `Type(arg1, arg2, **kw_args)` will first construct the
+        # parametric subtype `T = Type[type(arg1), type(arg2)]`
+        # and then call the equivalent of `T(arg1, arg2, **kw_args)`.
 
         if not cls.is_concrete:
-            argsT = tuple(type(arg) for arg in args)
-            if len(argsT) == 1:
-                argsT = argsT[0]
-            T = cls[argsT]
+            type_parameter = tuple(type(arg) for arg in args)
+            if len(type_parameter) == 1:
+                type_parameter = type_parameter[0]
+            T = cls[type_parameter]
         else:
             T = cls
 
-        # calls __new__ and __init__
+        # Calls `__new__` and `__init__`.
         return type.__call__(T, *args, **kw_args)
 
     @property
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -57,6 +57,7 @@ class ParametricTypeMeta(TypeMeta):
 
     @property
     def is_concrete(cls):
+        """"bool: Check whether the parametric type is instantiated or not."""
         return hasattr(cls, "_is_parametric")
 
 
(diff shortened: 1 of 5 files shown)


wesselb:
I think this looks great! The implementation looks sounds to me. I'd be very happy to like this merged!

Do you have any additional thoughts? Is a solution like this sufficient for your applications?

PhilipVinc:
This is sufficient for my needs for now.

It would be nice if the selection of which arguments are lifted to type parameters was overrideable (it wouldn't even be too hard. The `@parametric` decorator could check if a method like `__type_parameters(args) ->Tuple` was defined on the the class, and if so take it and use it to determine the type parameters, but... yeah. 
Let's leave this for the future.

Let me add some tests to this PR and bump the version and then I'd be super-happy if this can be merged (and a new release tagged)

wesselb:
> It would be nice if (...) Let's leave this for the future.

I had very similar thoughts, but agree that this is perhaps currently best left for the future.

> Let me add some tests to this PR and bump the version and then I'd be super-happy if this can be merged (and a new release tagged)

That sounds like a plan! I'll leave some comments on the PR. Moreover, should we adjust the README to explain this behaviour?

plum/parametric.py
```diff
@@ -30,8 +30,37 @@
...
+        # parametric subtype T = Type[type(arg1), type(arg2)]
+        # and then call the equivalent of T(arg1, arg2, **kwargs)
+
+        if not cls.is_concrete:
+            argsT = tuple(type(arg) for arg in args)
+            T = cls[argsT]
+        else:
+            T = cls
+
+        # this is the standard type.__call__ implementation
```
wesselb, line 52: Shall we simplify this to `return type.__call__(T, *args, **kwargs)`?

plum/parametric.py
```diff
@@ -30,8 +30,37 @@
...
+class ParametricTypeMeta(TypeMeta):
+    """Parametric Types must be instantiated with indexing."""
 
-class CovariantMeta(TypeMeta):
+    def __getitem__(self, *p):
+        # Type[Par1, Par2] creates a new parametric type
+
+        return self.__new__(self, *p)
+
+    def __call__(cls, *args, **kwargs):
```
wesselb, line 41: The convention I've been using is to use `kw_args` rather than `kwargs`.

plum/parametric.py
```diff
@@ -30,8 +30,37 @@
 
 _dispatch = Dispatcher()
 
+class ParametricTypeMeta(TypeMeta):
+    """Parametric Types must be instantiated with indexing."""
 
-class CovariantMeta(TypeMeta):
+    def __getitem__(self, *p):
```
wesselb, line 36: Shall we name `p` to `ps` to indicate plurality?

plum/parametric.py
```diff
@@ -30,8 +30,37 @@
...
+
+        return self.__new__(self, *p)
+
+    def __call__(cls, *args, **kwargs):
+        # Type(arg1, arg2, kwargs) will first construct the
+        # parametric subtype T = Type[type(arg1), type(arg2)]
+        # and then call the equivalent of T(arg1, arg2, **kwargs)
+
+        if not cls.is_concrete:
+            argsT = tuple(type(arg) for arg in args)
```
wesselb, line 47: If `len(args) == 1`, is `argsT = args[0]` a sensible simplification? Moreover, shall we name `argsT` to `type_parameters` to slightly clarify?

plum/parametric.py
```diff
@@ -30,8 +30,37 @@
 
 _dispatch = Dispatcher()
 
+class ParametricTypeMeta(TypeMeta):
+    """Parametric Types must be instantiated with indexing."""
 
-class CovariantMeta(TypeMeta):
+    def __getitem__(self, *p):
+        # Type[Par1, Par2] creates a new parametric type
```
wesselb, line 37: Would it make sense to move this comment and the below comment to the docstring of the class, so that the generated docs also include them?

plum/parametric.py
```diff
@@ -30,8 +30,37 @@
...
+            T = cls
+
+        # this is the standard type.__call__ implementation
+        obj = T.__new__(T, *args, **kwargs)
+        if obj is not None and isinstance(obj, T) and hasattr(obj, '__init__'):
+            obj.__init__(*args, **kwargs)
+        return obj
+
+    @property
+    def is_concrete(cls):
```
wesselb, line 59:
```suggestion
    def is_concrete(cls):
        """"bool: Check whether the parametric type is instantiated or not."""
```

wesselb (changes requested):
I've left some comments. Would you mind also running it through the `black` formatter?

plum/parametric.py
PhilipVinc: I agree with the renaming.
However I don't understand the semplification you are proposing. This code works for an arbitrary number of parameters

wesselb: Currently, this is what happens:

```python
In [1]: from plum import parametric

In [2]: @parametric
   ...: class A:
   ...:     def __init__(self, *xs):
   ...:         pass
   ...:

In [3]: A(1)
Out[3]: <__main__.A[(<class 'int'>,)] at 0x7fbfa809bf60>

In [4]: A[int](1)
Out[4]: <__main__.A[<class 'int'>] at 0x7fbfa80b4a20>
```

Note that `A(1)` has `(int,)` rather than `int` as its type parameter. It's not necessarily an issue, but it may be nice to simplify `(int,)` to `int`.

plum/parametric.py
PhilipVinc: works a charm.

plum/parametric.py
PhilipVinc: I thought it would work for the `()` constructor too. But evidently I was wrong..

I addressed the issue.

PhilipVinc:
I addressed all review comments and added tests.

I had to go and change 2 tests in https://github.com/wesselb/plum/pull/11/commits/4200fcc5c34195be2eb2284a8fab6e0b33c45a74 because they where using the `()` syntax to instantiate the parametrised type, otherwise everything was smooth.

I also added a feature so that the following now throws an error
```
>>> Test[int]
<class '__main__.Test[<class 'int'>]'>
>>> Test[int][float]
  File "<stdin>", line 1, in <module>
  File "/Users/filippovicentini/Dropbox/Ricerca/Codes/Python/plum/plum/parametric.py", line 46, in __getitem__
    raise TypeError("Cannot specify type parameters. "
TypeError: Cannot specify type parameters. This type is concrete.
```

as opposed to current master where there is a weird behaviour
```python
>>> import plum
>>> @plum.parametric
... class Test:
...   pass
...
>>> Test[1]
<class '__main__.Test[1]'>
>>> Test[1][2]
<__main__.Test[1] object at 0x106f12d90>
>>> Test[1][2][3]
  File "<stdin>", line 1, in <module>
TypeError: 'Test[1]' object is not subscriptable
```

PhilipVinc:
I added a test for the branch raising an error so that coverage is back 100%

plum/parametric.py
```diff
@@ -31,7 +31,43 @@
...
+    def __call__(cls, *args, **kw_args):
+        # Type(arg1, arg2, kw_args) will first construct the
+        # parametric subtype T = Type[type(arg1), type(arg2)]
+        # and then call the equivalent of T(arg1, arg2, **kw_args)
+
+        if not cls.is_concrete:
+            argsT = tuple(type(arg) for arg in args)
+            if len(argsT) == 1:
+                argsT = argsT[0]
+            T = cls[argsT]
```
wesselb, line 57:
```suggestion
            type_parameter = tuple(type(arg) for arg in args)
            if len(type_parameter) == 1:
                type_parameter = type_parameter[0]
            T = cls[type_parameter]
```

plum/parametric.py
```diff
@@ -31,7 +31,43 @@
...
+        else:
+            T = cls
+
+        # calls __new__ and __init__
```
wesselb, line 61:
```suggestion
        # Calls `__new__` and `__init__`.
```

plum/parametric.py
```diff
@@ -31,7 +31,43 @@
...
+    def __getitem__(self, *ps):
+        if not self.is_concrete:
+            return self.__new__(self, *ps)
+        else:
+            raise TypeError("Cannot specify type parameters. This type is concrete.")
+
+    def __call__(cls, *args, **kw_args):
+        # Type(arg1, arg2, kw_args) will first construct the
+        # parametric subtype T = Type[type(arg1), type(arg2)]
+        # and then call the equivalent of T(arg1, arg2, **kw_args)
```
wesselb, line 51:
```suggestion
        # `Type(arg1, arg2, **kw_args)` will first construct the
        # parametric subtype `T = Type[type(arg1), type(arg2)]`
        # and then call the equivalent of `T(arg1, arg2, **kw_args)`.
```

plum/parametric.py
```diff
@@ -31,7 +31,43 @@
 _dispatch = Dispatcher()
 
 
-class CovariantMeta(TypeMeta):
+class ParametricTypeMeta(TypeMeta):
+    """Parametric Types can be instantiated with indexing.
```
wesselb, line 35:
```suggestion
    """Parametric types can be instantiated with indexing.
```

tests/test_parametric.py
```diff
@@ -75,16 +75,48 @@ def tuple_elements_are_identical(tup1, tup2):
     assert tuple_elements_are_identical(type_parameter(A[a1, a2]()), (a1, a2))
     assert tuple_elements_are_identical(type_parameter(A[1, a2]()), (1, a2))
 
+    # test raise error if specify parameters twice
```
wesselb, line 78:
```suggestion
    # Test that an error is raised if type parameters are specified twice.
```

wesselb (changes requested):
This looks great. Thanks for making the changes, adding a description to the README, and fixing up the coverage! I've added some minor style suggestions mostly  for consistency with the rest of the code base. Besides those, I'm happy to merge.

Perhaps one other feature which we could include is to simplify the printing of `a` in 

```python
In [1]: from plum import parametric

In [2]: @parametric
   ...: class A:
   ...:     def __init__(self, *args):
   ...:         pass
   ...:

In [3]:

In [3]: a = A(1, 2)

In [4]: a
Out[4]: <__main__.A[(<class 'int'>, <class 'int'>)] at 0x7f89b050c8d0>
```
to `<__main__.A[<class 'int'>, <class 'int'>] at 0x7f89b050c8d0>`, so without the surrounding `(` and `)`, which would amount to a small tweaking of [this line](https://github.com/wesselb/plum/blob/962b47f077c486e63467c841b29c2c3d281fb09b/plum/parametric.py#L82), but this can also be addressed later.

PhilipVinc:
I applied your suggestions and fixed the issue you raised about the name.
It was due to  `__getitem__` wrapping everything inside a tuple when it was already a tuple, before calling `__new__`.
It's fixed now (I pasted your code, which i thought i could get by without...)

tests/test_parametric.py
```diff
@@ -75,16 +75,48 @@ def tuple_elements_are_identical(tup1, tup2):
...
+    a2 = A(5.0)
+
+    assert a1.x == 5.0
+    assert a2.x == 5.0
+    assert a1.y == 3
+    assert a2.y == 3
+
+    assert type_parameter(a1) == float
+    assert type_parameter(a2) == float
+    assert type(a1) == type(a2)
```
wesselb, line 101:
```suggestion
    assert type(a1) == type(a2)
    assert type(a1).__name__ == type(a2).__name__ == f"A[{float}]"
```

tests/test_parametric.py
```diff
@@ -75,16 +75,48 @@ def tuple_elements_are_identical(tup1, tup2):
...
-    a = A[1](5.0)
+    assert b1.x == 5.0
+    assert b2.x == 5.0
+    assert b1.y == 3
+    assert b2.y == 3
 
-    assert a.x == 5.0
+    assert type_parameter(b1) == (float, int)
+    assert type_parameter(b2) == (float, int)
+    assert type(b1) == type(b2)
```
wesselb, line 119:
```suggestion
    assert type(b1) == type(b2)
    assert type(b1).__name__ == type(b2).__name__ == f"B[{float},{int}]"
```

wesselb (changes requested):
Ah, I see!

I've just suggested two more tests which test the printing, and then this looks good to go. :)

PhilipVinc:
good idea.
Sorry for being lazy.

wesselb (approved):

wesselb:
No problem at all. This looks ready to merge then!

wesselb:
I've released the new version `v1.2.0` to PyPI. Thanks @PhilipVinc for the contribution!

PhilipVinc:
thanks a lot!

PR #14, opened 2021-06-28
Accept *None* as valid type annotation

### LINKED ISSUE #13: Accept `None` as type annotation
Hi! Thank you so much for this library, I've been looking for something like this for a long time!

I would like to suggest accepting `None` as a valid type in Plum. For instance, this snippet does not work:
```py
from plum import dispatch

def foo(x: None) -> None:
    return x

print(foo(None)) # raises: RuntimeError: Could not convert "None" to a type.
```
whereas this one works as expected:
```py
from plum import dispatch

NoneType = type(None)

@dispatch
def foo(x: NoneType) -> NoneType:
    return x

print(foo(None)) # prints: None
```
Since type-annotating `None`'s type as `None` is the standard way to do it, I think this simple enhancement would make dispatching a lot more intuitive in this case.

I would also love to submit a PR if you agree, I'm just not familiar with Plum's implementation. It seems to me that adding
```py
if obj is None:
    return Type(type(None))
```
just [before the RuntimeError](https://github.com/wesselb/plum/blob/516e4aa75910abcfd7a510726c586600b9edfd6e/plum/type.py#L364) should fix this, am I correct?

ruancomelli:
Accept `None` as valid type annotations, so that the following code snippet works:
```py
from plum import dispatch


def foo(x: None) -> None:
  return x


assert foo(None) is None
```

The solution simply checks if functions contain `None` as type annotation and, if so, automatically converts them to `builtins.NoneType`.

Closes #13

### DIFF
--- a/plum/type.py
+++ b/plum/type.py
@@ -362,6 +362,9 @@ def ptype(obj):
     if isinstance(obj, type):
         return Type(obj)
 
+    if obj is None:
+        return Type(type(None))
+
     raise RuntimeError(f'Could not convert "{obj}" to a type.')
 
 
--- a/tests/dispatcher/test_cases.py
+++ b/tests/dispatcher/test_cases.py
@@ -341,3 +341,13 @@ def test_property():
     a.name = "another name"
     with pytest.raises(NotFoundLookupError):
         a.name = 1
+
+
+def test_none():
+    dispatch = Dispatcher()
+
+    @dispatch
+    def foo(x: None) -> None:
+        return x
+
+    assert foo(None) is None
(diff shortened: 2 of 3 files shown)


wesselb:
This looks good! Thanks. :)

Could you perhaps also add some along the lines of

```python
def test_none():
    dispatch = Dispatcher()

    @dispatch
    def foo(x: None) -> None:
        return x

    assert foo(None) is None
```

as a test case at the end of `tests/dispatcher/test_cases.py`?

ruancomelli:
Sure! I don't know how I missed that 😅

wesselb:
This looks great. :) Thank you for the contribution, @ruancomelli!

wesselb:
I've pushed a new release `v1.3.1` to PyPI which includes this PR.

PR #15, opened 2021-07-06
[WIP] cythonize plum.parametric

PhilipVinc:
I experimented with cythonizing the plum.parametric module to speed it up a bit (for my own personal peace of mind).

Other than slightly modifying the test because now `typeof` is a function instead of a `function`, there is a change needed:

In Cython, type hints are all strings. So the `type_of` dispatches of `typeof(obj: list)` and `typeof(obj: tuple)` in `plum.parametric` will have a string and type instead of a real class, thus in type.py they will be forward references/promises which are never _fullfilled_.

Apparently plum breaks down when those two promises are not fulfilled.
A fix was to hardcode those two cases in ptype:
```python
def ptype(obj):

    if hasattr(obj, "__module__") and obj.__module__ == "typing":
        ...
    if isinstance(obj, str):
        if obj == "list":
            return ptype(list)
        elif obj == "tuple":
            return ptype(tuple)
        else:
            return get_forward_reference(obj)
```

there's probably a better way to address this, but I can't see it now.
The PromisedType mechanism is a bit confusing...

### DIFF
--- a/plum/type.py
+++ b/plum/type.py
@@ -356,7 +356,12 @@ def ptype(obj):
 
     # Strings are forward references.
     if isinstance(obj, str):
-        return get_forward_reference(obj)
+        if obj == "list":
+            return ptype(list)
+        elif obj == "tuple":
+            return ptype(tuple)
+        else:
+            return get_forward_reference(obj)
 
     # If `obj` is a `type`, wrap it in a `Type`.
     if isinstance(obj, type):
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,9 @@ setup(
     packages=find_packages(exclude=["docs"]),
     python_requires=">=3.6",
     install_requires=requirements,
-    ext_modules=[Extension("plum.function", ["plum/function.py"]), 
-                 Extension("plum.parametric", ["plum/parametric.py"])],
+    ext_modules=[
+        Extension("plum.function", ["plum/function.py"]),
+        Extension("plum.parametric", ["plum/parametric.py"]),
+    ],
     include_package_data=True,
 )
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ setup(
     packages=find_packages(exclude=["docs"]),
     python_requires=">=3.6",
     install_requires=requirements,
-    ext_modules=[Extension("plum.function", ["plum/function.py"])],
+    ext_modules=[Extension("plum.function", ["plum/function.py"]), 
+                 Extension("plum.parametric", ["plum/parametric.py"])],
     include_package_data=True,
 )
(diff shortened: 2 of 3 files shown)


PhilipVinc:
@wesselb do you have benchmarks i can try ?

wesselb:
Another way of addressing the type-hints-become-strings issue is to use `dispatch.multi`, which doesn't use the type hints:
```
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -280,12 +280,12 @@ def type_of(obj):
     return ptype(type(obj))


-@_dispatch
+@_dispatch.multi((list,))
 def type_of(obj: list):
     return List(_types_of_iterable(obj))


-@_dispatch
+@_dispatch.multi((tuple,))
 def type_of(obj: tuple):
     return Tuple(*(type_of(x) for x in obj))
```

I agree that the promise system is a confusing, but unfortunately it's necessary because there are some circular dependencies between the files.

There is a [small benchmark](https://github.com/wesselb/plum/blob/master/benchmark.py) in the root of the repo. Unfortunately, it looks like Cythonising `parametric.py` does not make much of a difference.

Without Cythonising:

```
# Function Calls
Native call:       0.08 us (1.0 x)
First plum call: 122.31 us (1538.0 x)
Plum call:         0.33 us (4.0 x)

# Parametric Function Calls
Native call:       0.07 us (1.0 x)
First plum call: 450.13 us (6565.0 x)
Plum call:        10.80 us (157.0 x)

# Class Calls
Native call:       0.09 us (1.0 x)
First plum call: 114.20 us (1269.0 x)
Plum call:         1.05 us (12.0 x)

# Class Attribute Calls
Native call:       0.14 us (1.0 x)
First plum call: 165.22 us (1178.0 x)
Plum call:         1.14 us (8.0 x)
```

With Cythonising:
```
# Function Calls
Native call:       0.09 us (1.0 x)
First plum call: 132.80 us (1517.0 x)
Plum call:         0.32 us (4.0 x)

# Parametric Function Calls
Native call:       0.07 us (1.0 x)
First plum call: 474.93 us (6593.0 x)
Plum call:         8.98 us (125.0 x)

# Class Calls
Native call:       0.09 us (1.0 x)
First plum call: 112.77 us (1230.0 x)
Plum call:         1.06 us (12.0 x)

# Class Attribute Calls
Native call:       0.14 us (1.0 x)
First plum call: 114.92 us (826.0 x)
Plum call:         1.15 us (8.0 x)
```

The big performance bottleneck is `type_of`. Currently, [there is a performance optimisation in place](https://github.com/wesselb/plum/blob/master/plum/function.py#L452) which only calls `type_of` if parametric types are being used. But this is too cautious: `type` will work just fine with `@parametric` classes. I can see a few scenarios where a mechanism like `type_of` cannot be avoided:

* parametric lists (`List[int]`), because `type([1, 1])` will always give just `list`;
* parametric tuples (`Tuple[int, int]`), for similar reasons; and
* constructions [like this](https://github.com/wesselb/plum/#hooking-into-type-inference).

What we probably need is a better algorithm to determine if `type_of` if necessary, not just use it whenever a parametric types are being used. What are your thoughts?

wesselb:
Closing this since we've decided to go for a pure-Python package without Cythonisation.

PR #17, opened 2021-07-12
make resolution errors nicer

PhilipVinc:
This PR changes `ForwardReferenceType` to have a better `__repr__` method, that only calls `ptype` on itself it the promise has been kept, otherwise it properly prints it's name.

That way, when a signature check fails because a type has not been defined, a clear error message is visualized.
Before, one had to guess which type was causing the issue


plum/resolvable.py
```diff
@@ -41,6 +41,15 @@ def deliver(self, obj):
 
     def resolve(self):
         if self._obj is None:
-            raise ResolutionError("Promise was not kept.")
+            raise ResolutionError("Promise `{}` was not kept.".format(repr(self)))
```
wesselb, line 44:
```suggestion
            raise ResolutionError(f"Promise `{self!r}` was not kept.")
```

plum/resolvable.py
```diff
@@ -41,6 +41,15 @@ def deliver(self, obj):
...
             return self._obj
+
+    @property
+    def resolved(self):
+        """bool: True if the Promise has been kept."""
+        return self._obj is not None
+
+    def __repr__(self):
+        # fallback in case superclasses don't overload repr
+        return "Promise(obj=self._obj)"
```
wesselb, line 55:
```suggestion
        # Fallback in case subclasses don't overload `__repr__`.
        return f"Promise(obj={self._obj!r})"
```

Should this also check whether the promise has already been resolved, because `self._obj` might not exist yet? I.e., maybe give `<Promise: unfulfilled>` back if it isn't yet resolved?

plum/type.py
```diff
@@ -229,7 +229,10 @@ def __hash__(self):
         return hash(ptype(self.resolve()))
 
     def __repr__(self):
-        return repr(ptype(self.resolve()))
+        if self.resolved:
+            return repr(ptype(self.resolve()))
+        else:
+            return "ResolvableType(unresolved)"
```
wesselb, line 235:
```suggestion
            return "<ResolvableType: unresolved>"
```

I like to give an angle-bracket representation back if the output of `__repr__` isn't able to fully reconstruct the object.

plum/type.py
```diff
@@ -289,6 +292,12 @@ def __init__(self, name):
         self.name = name
         PromisedType.__init__(self)
 
+    def __repr__(self):
+        if self.resolved:
+            return super().__repr__()
+        else:
+            return "ForwardReferencedType(name={})".format(self.name)
```
wesselb, line 299:
```suggestion
            return f"ForwardReferencedType(name={self.name!r})"
```

wesselb (changes requested):
This looks great! Thanks for putting this together. :) I've left a few comments!

plum/resolvable.py
PhilipVinc: I'm ok with this, but i think the f-string syntax has only been around since python 3.6.
I thouht you were trying to support older versions?

plum/resolvable.py
PhilipVinc: that is a good idea, indeed. I implemented it

PhilipVinc:
I implemented all your review comments.

plum/resolvable.py
```diff
@@ -41,6 +41,18 @@ def deliver(self, obj):
...
     def resolve(self):
         if self._obj is None:
-            raise ResolutionError("Promise was not kept.")
+            raise ResolutionError(f"Promise `{self!r}` was not kept.")
         else:
             return self._obj
+
+    @property
+    def resolved(self):
+        """bool: True if the Promise has been kept."""
```
wesselb, line 50:
```suggestion
        """bool: `True` if the promise has been kept."""
```

plum/resolvable.py
```diff
@@ -41,6 +41,18 @@ def deliver(self, obj):
...
+        return self._obj is not None
+
+    def __repr__(self):
+        # fallback in case superclasses don't overload repr
```
wesselb, line 54:
```suggestion
        # Fallback in case subclasses don't overload `__repr__`.
```

(I'm assuming you mean classes that inherit `Promise`.)

plum/resolvable.py
```diff
@@ -41,6 +41,18 @@ def deliver(self, obj):
...
+        if self.resolved:
+            return f"Promise(obj={self._obj!r})"  # pragma: no cover
```
wesselb, line 56: Would you able to test the printing of `Promise` in e.g. a new test `tests/test_resolvable.py::test_promise_repr`?

wesselb: Moreover, should we change the constructor to

```python
    def __init__(self, obj=None):
        self._obj = obj
```

so that the output of `__repr__` can be `eval`d?

wesselb (changes requested):
Just a few more very tiny nitpicks. :) This looks essentially good to go!

plum/resolvable.py
PhilipVinc: Done

wesselb (approved):
This looks ready to go now. Thanks, @PhilipVinc!

wesselb:
The latest release `v1.3.3` includes this PR.

PR #18, opened 2021-07-12
Add Val Type

PhilipVinc:
This PR adds a `Val` type (similar to julia's Val type).
To do so, I add a hook to override the deduction of the type-parameters when calling `parametrictype(args...)`.

The hook is simply a function, provided to the `@parametric` decorator, that takes the arguments as input and must return the type parameters as output.
The default mechanism is now switched to a default implementation of this function.

I need to document what is added in this PR, but i'd like to know if you'd accept this first.

--

The motivation for this is that i have some bool values in netket that i'd like to use to dispatch on the right implementation.
As those values are jax-jit time constants, it makes sense to use them for dispatch to the right implementation...

### DIFF
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -330,8 +330,7 @@ promised_type_of2.deliver(type_of)
 
 @parametric
 class Val:
-    """
-    A parametric type used to move information from the value-domain to the type-domain.
+    """A parametric type used to move information from the value domain to the type domain.
     """
 
     @classmethod
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -132,7 +132,7 @@ def parametric(Class):
 
     ```python
     @classmethod
-    __infer_type_parameter__(cls, *vals, **kw_args) -> Tuple:
+    __infer_type_parameter__(cls, *args, **kw_args) -> Tuple:
         return tuple(type(arg) for arg in args)
     ```
     """
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -56,7 +56,7 @@ class ParametricTypeMeta(TypeMeta):
         # and then call the equivalent of `T(arg1, arg2, **kw_args)`.
 
         if not cls.is_concrete:
-            type_parameter = cls._init_args_types(*args)
+            type_parameter = cls.__infer_type_parameter__(*args, **kw_args)
             T = cls[type_parameter]
         else:
             T = cls
@@ -64,7 +64,7 @@ class ParametricTypeMeta(TypeMeta):
         # Calls `__new__` and `__init__`.
         return type.__call__(T, *args, **kw_args)
 
-    def _init_args_types(cls, *args):
+    def __infer_type_parameter__(cls, *args, **kw_args):
         """Function called when the constructor of this parametric type is called
         before the parameters have been specified.
 
@@ -78,9 +78,6 @@ class ParametricTypeMeta(TypeMeta):
         Returns:
             A type or tuple of types.
         """
-        print("cls", cls)
-        print("args", args)
-
         type_parameter = tuple(type(arg) for arg in args)
         if len(type_parameter) == 1:
             type_parameter = type_parameter[0]
@@ -123,22 +120,21 @@ class CovariantMeta(ParametricTypeMeta):
         return type.__subclasscheck__(self, subclass)
 
 
-def parametric(Class=None, *, init_args_types=None):
+def parametric(Class):
     """A decorator for parametric classes.
+    
+    When the constructor of this parametric type is called before the type parameter
+    has been specified, the type parameters are inferred from the arguments of the 
+    constructor by calling the following function.
 
-    Optional keyword arguments:
-        init_args_types: Function called when the constructor of this parametric type is
-            called before the parameters have been specified. This function must have
-            the signature `f(cls, *args) -> Union[Any,Tuple]` and it must return an
-            hashable object or tuple that will be used as type parameters of the
-            resulting types. The default implementation of this function will return
-            the types of all arguments.
+    The default implementation is shown here, but it is possible to override it.
+
+    ```python 
+    @classmethod
+    __infer_type_parameter__(cls, *vals, **kw_args) -> Tuple:
+        return tuple(type(arg) for arg in args)    
+    ```
     """
-
-    # allow the kwargs to be passed in without using functools.partial explicitly
-    if Class is None:
-        return partial(parametric, init_args_types=init_args_types)
-
     subclasses = {}
 
     if not issubclass(Class, object):  # pragma: no cover
@@ -176,12 +172,6 @@ def parametric(Class=None, *, init_args_types=None):
     # Create parametric class.
     ParametricClass = ParametricTypeMeta(Class.__name__, (Class,), {"__new__": __new__})
     ParametricClass.__module__ = Class.__module__
-
-    if init_args_types is not None:
-        # bind method to class
-        # https://stackoverflow.com/questions/1015307/python-bind-an-unbound-method
-        bound_method = init_args_types.__get__(ParametricClass)
-        setattr(ParametricClass, "_init_args_types", bound_method)
 
     # Attempt to correct docstring.
     try:
@@ -337,22 +327,22 @@ promised_type_of1.deliver(type_of)
 promised_type_of2.deliver(type_of)
 
 
-def _Val_type_param(cls, *arg):
-    """Function called when the constructor of `Val` is called to determine the type
-    parameters.
-    """
-    if len(arg) == 0:
-        raise ValueError("The value must be specified.")
-    elif len(arg) > 1:
-        raise ValueError("Too many values. Val accepts only one argument.")
-    return arg[0]
-
-
-@parametric(init_args_types=_Val_type_param)
+@parametric
 class Val:
     """
     A parametric type used to move information from the value-domain to the type-domain.
     """
+
+    @classmethod
+    def __infer_type_parameter__(cls, *arg, **kwargs):
+        """Function called when the constructor of `Val` is called to determine the type
+        parameters.
+        """
+        if len(arg) == 0:
+            raise ValueError("The value must be specified.")
+        elif len(arg) > 1:
+            raise ValueError("Too many values. Val accepts only one argument.")
+        return arg[0]
 
     def __init__(self, arg=None):
         """
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -79,6 +79,8 @@ class ParametricTypeMeta(TypeMeta):
         Returns:
             type or tuple[type]: A type or tuple of types.
         """
+
+        # TODO: Use `type_of` instead of `type`
         type_parameter = tuple(type(arg) for arg in args)
         if len(type_parameter) == 1:
             type_parameter = type_parameter[0]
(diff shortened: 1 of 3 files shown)


wesselb:
As with your other PRs, also this looks great!!

Instead of adding a keyword `init_args_types` to `parametric`, how would you feel about instead checking for a magic method `__infer_type_parameter__`, in a way that the below would be the implementation of `NTuple`?

```python
from typing import Union
from plum import parametric, type_parameter

@parametric
class NTuple
    def __init__(self, *vals):
        n, val_type = type_parameter(self)
        assert all([isinstance(val, val_type) for val in vals])
        assert len(vals) == n
        self.vals = vals

    def __infer_type_parameter__(self, *vals):
        return Union[tuple(type(val) for val in vals)], len(vals)

```

If the magic method doesn't exist, then the fallback should be the current behaviour.

PhilipVinc:
I had thought about that at first, and indeed from one point of view it feels cleaner.

However i had thought about the following counter-argument:
 - This method should live in the metaclass, not the class itself. The first argument shouldn't be `self` but `cls`, and it should be marked as a `static` method (not even, because it's called before the class is actually created) so it felt a bit weird to put it in the class.
- if i slightly change the signature of `init_args_types` (or `infer_type_parameters`, which is much better), in order not to be `infer_type_parameters(cls, val1, val2=3)...` one can use `@dispatch` on the infer_type function itself. Maybe this is useless, but maybe it might be interesting. In any case it's more general.

Point 1 is a bit moot, so could be ignored with a bit of metaprogramming in order to extract the underlying function from the method.
Point 2 could also be ignored, because it's possible to declare dispatch rules also on class methods but i'm unsure of how it will work together with the workaround to point 1.

wesselb:
Those are two good points!

Regarding the first point, we could require `__infer_type_parameter__` to be a `@classmethod`. Moreover, regarding the second point, the user would be able to also use `@dispatch` within the class:

```python
from typing import Union
from numbers import Number
from plum import dispatch, parametric, type_parameter

@parametric
class NTupleForNumbers
    def __init__(self, *vals):
        n, val_type = type_parameter(self)
        assert all([isinstance(val, val_type) for val in vals])
        assert len(vals) == n
        self.vals = vals

    @dispatch
    @classmethod
    def __infer_type_parameter__(cls, *vals: Number):
        return Union[tuple(type(val) for val in vals)], len(vals)
```

(Currently, `@dispatch` appears to not work with `@staticmethod` and `@classmethod`, but that should be easy to fix.)

What do you think? It feels kinda nice to bundle `__infer_type_parameter__` together with the class and to keep `parametric` just a simple decorator.

Just letting you know that I'm currently on holiday for a few days. I'll get back to this PR and your other PRs early next week!

PhilipVinc:
That's a nice idea!
I implemented it (classmethods still dont work with dispatch, but that can be addressed in a separate PR).

By the way, i just noticed that maybe you already had something like `Val` type? 
Is it what `parametric.Kind` is supposed to be?

PhilipVinc:
bumpety bump @wesselb , unless you're still on holiday

plum/parametric.py
```diff
@@ -54,16 +56,33 @@ def __call__(cls, *args, **kw_args):
...
+    def __infer_type_parameter__(cls, *args, **kw_args):
+        """Function called when the constructor of this parametric type is called
+        before the parameters have been specified.
+
+        The default behaviour is to take as parameters the type of every argument,
+        but this behaviour can be overridden by redefining this function on the
+        metaclass.
+
+        Args:
+            *args: the argument-values passed to the __init__ method.
```
wesselb, line 76:
```suggestion
            *args: Positional arguments passed to the `__init__` method.
            *kw_args: Keyword arguments passed to the `__init__` method.
```

plum/parametric.py
```diff
@@ -54,16 +56,33 @@ def __call__(cls, *args, **kw_args):
...
+
+        Returns:
+            A type or tuple of types.
```
wesselb, line 79:
```suggestion
            type or tuple[type]: A type or tuple of types.
```

plum/parametric.py
```diff
@@ -102,7 +121,20 @@ def __subclasscheck__(self, subclass):
...
+
+    When the constructor of this parametric type is called before the type parameter
+    has been specified, the type parameters are inferred from the arguments of the
+    constructor by calling the following function.
+
+    The default implementation is shown here, but it is possible to override it.
+
+    ```python
+    @classmethod
+    __infer_type_parameter__(cls, *vals, **kw_args) -> Tuple:
```
wesselb, line 134:
```suggestion
    __infer_type_parameter__(cls, *args, **kw_args) -> Tuple:
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
 # Deliver `type_of`.
 promised_type_of1.deliver(type_of)
 promised_type_of2.deliver(type_of)
+
+
+@parametric
+class Val:
+    """
+    A parametric type used to move information from the value-domain to the type-domain.
```
wesselb, line 333:
```suggestion
    """A parametric type used to move information from the value domain to the type domain.
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+    """
+
+    @classmethod
+    def __infer_type_parameter__(cls, *arg, **kwargs):
```
wesselb, line 337:
```suggestion
    def __infer_type_parameter__(cls, *arg, **kw_args):
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+        """Function called when the constructor of `Val` is called to determine the type
+        parameters.
+        """
+        if len(arg) == 0:
+            raise ValueError("The value must be specified.")
+        elif len(arg) > 1:
+            raise ValueError("Too many values. Val accepts only one argument.")
```
wesselb, line 344:
```suggestion
            raise ValueError("Too many values. `Val` accepts only one argument.")
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+        return arg[0]
+
+    def __init__(self, arg=None):
+        """
+        Construct a Value object with type `Val(arg)` that cna be used to dispatch
```
wesselb, line 349:
```suggestion
        Construct a value object with type `Val(arg)` that can be used to dispatch
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+        based on values.
+
+        Args:
+            arg: the value to be moved to the type domain
```
wesselb, line 353:
```suggestion
            arg (object): The value to be moved to the type domain.
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+        """
+        if type(self).is_concrete:
+            return
+        else:
+            raise ValueError("The value must be specified.")  # pragma: no cover
```
wesselb, line 358:
```suggestion
            raise ValueError("The value must be specified.")
```

(I think this is tested.)

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+
+    def __repr__(self):
+        return "{}()".format(repr(type(self)))
```
wesselb, line 361:
```suggestion
        return f"{repr(type(self))}()"
```

plum/parametric.py
```diff
@@ -293,3 +325,40 @@ def type_of(obj: tuple):
...
+
+    def __eq__(self, other):
```
wesselb, line 363:
```suggestion
    @_dispatch
    def __eq__(self, other: "Val"):
```

Maybe add this for safety? Perhaps also add a quick test to test this?

tests/test_parametric.py
```diff
@@ -121,6 +122,24 @@ def __init__(self, x, y):
     assert type(b1).__name__ == type(b2).__name__ == f"B[{float},{int}]"
 
 
+def test_override_type_parameters():
+    @parametric
+    class NTuple:
+        @classmethod
+        def __infer_type_parameter__(self, *vals, **kwargs):
+            # mimicks the type parameters of an NTuple
```
wesselb, line 130:
```suggestion
        def __infer_type_parameter__(self, *vals, **kw_args):
            # Mimicks the type parameters of an `NTuple`.
```

tests/test_parametric.py
```diff
@@ -302,3 +321,25 @@ def f(x: NPArray[2]):
     assert f(np.random.randn(10, 10)) == "matrix"
     with pytest.raises(NotFoundLookupError):
         f(np.random.randn(10, 10, 10))
+
+
+def test_Val():
```
wesselb, line 326:
```suggestion
def test_val():
```

plum/parametric.py
```diff
@@ -54,16 +56,33 @@ def __call__(cls, *args, **kw_args):
...
+        but this behaviour can be overridden by redefining this function on the
+        metaclass.
+
+        Args:
+            *args: the argument-values passed to the __init__ method.
+
+        Returns:
+            A type or tuple of types.
+        """
+        type_parameter = tuple(type(arg) for arg in args)
```
wesselb, line 81: Should we consider using `type_of` here?

wesselb (changes requested):
Hey! I've been back for a few days. Thanks for the bump!

I really like this pattern with `__infer_type_parameter__`. This looks great.

I've left some minor comments.

wesselb:
Also, would you also like to add an entry to the README? I'm also happy to do that, if you like.

plum/parametric.py
PhilipVinc: Why? 
I think this is already pretty safe, and avoids an extra dispatch...
Unless we want to make this extensible (but do we?)

plum/parametric.py
PhilipVinc: fine by me

PhilipVinc:
I added to the readme.

wrt using `type_of`: i reverted the commit.
It turns out that problems arise in the tests because you cannot really compare plum types with standard python types:

```python
>>> plum.type_of(float) == float
  File "<stdin>", line 1, in <module>
  File "/home/filippovicentini/Dropbox/Ricerca/Codes/Python/plum/plum/util.py", line 30, in __eq__
    return self <= other <= self
  File "/home/filippovicentini/Dropbox/Ricerca/Codes/Python/plum/plum/type.py", line 103, in __le__
    
TypeError: issubclass() arg 1 must be a class
```

I think this PR could be merged as-is, and eventually this can be left as an improvement for the future if this issue is resolved

plum/parametric.py
wesselb: Hmm, yeah, I was worried about the case where `other` isn't a `Val`, but checking the types should catch that.

PhilipVinc: What do you mean by checking the types?

wesselb: The implementation  of `__eq__` is `type(self) == type(other)`, so that can never be true if `other` isn't a `Val`, because then the type of `other` will be different. Think this looks good. :)

wesselb:
> I think this PR could be merged as-is, and eventually this can be left as an improvement for the future if this issue is resolved

Ah, that's unfortunate. I agree: let's leave it for a future issue.

Think this is ready to be merged then! As many times before, thanks @PhilipVinc for putting together another great PR :)

wesselb:
This is included in the latest release `v1.4.0`. Thanks again :)

PR #19, opened 2021-07-12
Improve dev QOL

PhilipVinc:
Hi Wessel,
this is a small QOL improvement for working on plum. Feel free to ignore it if you don't like this.

This PR inserts `black` into a pre-commit hook (so that if you do `pre-commit install`, every time you commit the code will be formatted correctly with black).

It also moves the black format check to a separate CI runner and uses a flag to show what caused the failure.


wesselb:
Nice! That's certaintly a QOL improvement. Thanks! :)

PR #20, opened 2021-07-13
[WIP] Support default arguments as additional signatures

### LINKED ISSUE #7: [BUG] Default arguments are not supported with Dispatch (regression from 0.2.5)
This was working in 0.2.5,

```python
import plum
from plum import dispatch

@dispatch
def test(a:int, b: int = 1):
	print('int, int')

@dispatch
def test(a:float, b: int = 1):
	print('float, int')

In [33]: 

In [33]: test(1,1)
NotFoundLookupError                       Traceback (most recent call last)
<ipython-input-33-9b86683066a3> in <module>
----> 1 test(1,1)

~/Documents/pythonenvs/netket_env/lib64/python3.8/site-packages/plum/function.cpython-38-x86_64-linux-gnu.so in plum.function.Function.__call__()

~/Documents/pythonenvs/netket_env/lib64/python3.8/site-packages/plum/function.cpython-38-x86_64-linux-gnu.so in plum.function.Function.resolve_method()

~/Documents/pythonenvs/netket_env/lib64/python3.8/site-packages/plum/function.cpython-38-x86_64-linux-gnu.so in plum.function.Function.resolve_signature()

NotFoundLookupError: For function "test", signature Signature(builtins.int, builtins.int) could not be resolved.

In [34]: test(1)
int, int

```

PhilipVinc:
Closes #7

It's a simple solution though maybe not the most elegant.
When a signature is to be registered, it checks if some arguments have a default.
If so, it creates a new signature without those arguments (incrementally, so if you have 4 arguments of which the last 3 have defaults, you create 4 signatures).

To be elegantly correct, it should also check that the type of a default argument matches the type of the type annotation.
Though to do that correctly i need to check if the function is parametric or not
so i need to have already registered some signatures...

therefore i'll probably need to rewrite this.
But i'm just putting it here for anyone to see in the meantime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,17 +304,17 @@ For more advanced use cases of forward references, you can use `plum.type.Promis
 
 Default arguments can be used. The type annotation must match the default value otherwise 
 an error is thrown. 
-Different default values can be used in different dispatch rules.
+As the example below illustrates, different default values can be used for different methods:
 
 ```python
 from plum import dispatch
 
 @dispatch
-def f(x:int, y:int=3):
+def f(x: int, y: int = 3):
     return f"Value for 2nd y: {y}"
 
 @dispatch
-def f(x:float, y:float=3.0):
+def f(x: float, y: float = 3.0):
     return f"Value for 2nd y: {y}"
 ```
 
--- a/README.md
+++ b/README.md
@@ -302,9 +302,34 @@ For more advanced use cases of forward references, you can use `plum.type.Promis
 
 ## Keyword Arguments and Default Values
 
-Keyword arguments can be used, but are *not* dispatched on.
-As a convention, arguments with default values should always be called as keyword 
-arguments.
+Default arguments can be used. The type annotation must match the default value otherwise 
+an error is thrown. 
+Different default values can be used in different dispatch rules.
+
+```python
+from plum import dispatch
+
+@dispatch
+def f(x:int, y:int=3):
+    return f"Value for 2nd y: {y}"
+
+@dispatch
+def f(x:float, y:float=3.0):
+    return f"Value for 2nd y: {y}"
+```
+
+```python
+>>> f(1)
+'Value for 2nd y: 3'
+
+>>> f(1.0)
+'Value for 2nd y: 3.0'
+
+>>> f(1.0, 4.0)
+'Value for 2nd y: 4.0'
+```
+
+Keyword-only arguments, separated by an asterisk from the other arguments, can also be used, but are *not* dispatched on.
 
 Example:
 
@@ -312,7 +337,7 @@ Example:
 from plum import dispatch
 
 @dispatch
-def f(x, option="a"):
+def f(x, *, option="a"):
     return f"Value for option: {option}"
 ```
 
@@ -327,33 +352,6 @@ def f(x, option="a"):
 NotFoundLookupError: For function "f", signature Signature(builtins.int, builtins.str) could not be resolved.
 ```
 
-If you want to use a default value for a positional argument, use the following pattern
-instead:
-
-
-```python
-from plum import dispatch
-
-@dispatch
-def f(x, option):
-    return f"Value for option: {option}"
-
-
-@dispatch
-def f(x):
-    return f(x, "a")  # Use default value for `option`.
-```
-
-```python
->>> f(1)              # This is fine.
-'Value for option: a'
-
->>> f(1, "b")         # And this will work!
-'Value for option: b'
-
->>> f(1, option="b")  # But this won't.
-TypeError: f() got an unexpected keyword argument 'option'
-```
 
 ## Comparison with `multipledispatch`
 
--- a/plum/function.py
+++ b/plum/function.py
@@ -60,21 +60,31 @@ def extract_signature(f):
         p = sig.parameters[arg]
 
         # Stop once we have seen all positional parameter without a default value.
-        if _is_not_empty(p.default) or p.kind in {p.KEYWORD_ONLY, p.VAR_KEYWORD}:
+        if p.kind in {p.KEYWORD_ONLY, p.VAR_KEYWORD}:
             break
 
         if p.kind == p.VAR_POSITIONAL:
             # Parameter is varargs.
             if _is_not_empty(p.annotation):
-                types.append(VarArgs(ptype(p.annotation)))
+                T = VarArgs(ptype(p.annotation))
             else:
-                types.append(VarArgs(default_obj_type))
+                T = VarArgs(default_obj_type)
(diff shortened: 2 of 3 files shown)


wesselb:
Hey @PhilipVinc, I think this concept takes the right approach to solve this. The next step, which is beyond the scope of this proposal, would be to add a mechanism that checks the types of the keyword arguments.

wesselb:
> I'm not sure it makes sense to dispatch on keyword arguments?

Sorry, I didn't mean to propose that we should dispatch on keyword arguments. I just meant that we might want to check that, when a method is chosen according to dispatch, the types of the keyword arguments match the initial type annotations. This is what you currently do for the default values, but we could also do that for the values at runtime.

README.md
```diff
@@ -302,59 +302,57 @@ For more advanced use cases of forward references, you can use `plum.type.Promis
...
+an error is thrown. 
+Different default values can be used in different dispatch rules.
 
 ```python
 from plum import dispatch
 
 @dispatch
-def f(x, option="a"):
-    return f"Value for option: {option}"
+def f(x:int, y:int=3):
```
wesselb, line 313:
```suggestion
def f(x: int, y: int = 3):
```

(I think this is what Black prescribes.)

README.md
```diff
@@ -302,59 +302,57 @@ For more advanced use cases of forward references, you can use `plum.type.Promis
...
+    return f"Value for 2nd y: {y}"
+
+@dispatch
+def f(x:float, y:float=3.0):
```
wesselb, line 317:
```suggestion
def f(x: float, y: float = 3.0):
```

Idem.

README.md
```diff
@@ -302,59 +302,57 @@ For more advanced use cases of forward references, you can use `plum.type.Promis
...
 ## Keyword Arguments and Default Values
 
-Keyword arguments can be used, but are *not* dispatched on.
-As a convention, arguments with default values should always be called as keyword 
-arguments.
-
-Example:
+Default arguments can be used. The type annotation must match the default value otherwise 
+an error is thrown. 
+Different default values can be used in different dispatch rules.
```
wesselb, line 307:
```suggestion
As the example below illustrates, different default values can be used for different methods:
```

plum/function.py
```diff
@@ -88,6 +98,35 @@ def extract_signature(f):
...
+
+    arg_names = list(sig.parameters.keys())
+    arg_names.reverse()
+    n_args = len(arg_names)
+
+    deleted_args = 0
+    for i, arg in enumerate(arg_names):
+        p = sig.parameters[arg]
+
+        # ignore keyword arguments
```
wesselb, line 115:
```suggestion
        # Ignore keyword arguments.
```

plum/function.py
```diff
@@ -88,6 +98,35 @@ def extract_signature(f):
...
+        if p.kind in {p.KEYWORD_ONLY, p.VAR_KEYWORD}:
+            continue
+
+        # stop when reacing non default args
```
wesselb, line 119:
```suggestion
        # Stop when non-default arguments are reached.
```

plum/function.py
```diff
@@ -88,6 +98,35 @@ def extract_signature(f):
...
 
 
+def extract_default_args(signature, f):
+    # Extract specification.
+    sig = inspect.signature(f)
+
+    signatures = [signature]
+
+    arg_names = list(sig.parameters.keys())
+    arg_names.reverse()
```
wesselb, line 108:
```suggestion
    # We start at the end and, once we reach non-keyword-only arguments, delete the
    # argument with defaults values one by one. This generates a sequence of signatures,
    # which we return.
    arg_names.reverse()
```

plum/function.py
```diff
@@ -88,6 +98,35 @@ def extract_signature(f):
...
+    n_args = len(arg_names)
+
+    deleted_args = 0
+    for i, arg in enumerate(arg_names):
```
wesselb, line 112:
```suggestion
    for arg in arg_names:
```

plum/function.py
```diff
@@ -88,6 +98,35 @@ def extract_signature(f):
     return signature, return_type
 
 
+def extract_default_args(signature, f):
```
wesselb, line 101: Would you mind adding a docstring for this function?

tests/dispatcher/test_dispatcher.py
```diff
@@ -78,6 +78,53 @@ def f(x: int, option=None):
...
+    assert f(2.0) == y_default ** 2
+    assert f(2.0, y=4) == 4 ** 2
+
+    with pytest.raises(NotFoundLookupError):
+        f(2, 4.0)
+
+    with pytest.raises(NotFoundLookupError):
+        f(2, 4.0, option=2)
+
+    # wrong default type
```
wesselb, line 109:
```suggestion
    # Wrong default type
```

tests/dispatcher/test_dispatcher.py
```diff
@@ -78,6 +78,53 @@ def f(x: int, option=None):
...
+    with pytest.raises(TypeError):
+
+        @dispatch
+        def f(x: int, y: float = y_default):
+            return y
+
+    # ignore wrong type annotations on keyword arguments
```
wesselb, line 116:
```suggestion
    # Ignore wrong type annotations on keyword arguments.
```

wesselb (changes requested):
This looks good! No substantial comments, just minor stylistic changes.

wesselb:
I'm going to be away until the end of the week, but this looks basically ready to go!

Thanks again @PhilipVinc for your contributions.

README.md
PhilipVinc: yeah, ideally black could be run on readme.md... 
I'm very bad at enforcing style manually

plum/function.py
PhilipVinc: done. I also renamed it

wesselb:
This looks fantastic. Merging this now! Thanks, @PhilipVinc, this is a substantial improvement in how default arguments are handled. :)

PhilipVinc:
If you tag a new version now from here https://github.com/wesselb/plum/releases then I can try to fix conda immediately.

wesselb:
Sweet, thanks! I've tagged a new version, which appears to have successfully been submitted to PyPI.

PhilipVinc:
Yep! I'm glad everything was smooth.

I updated https://github.com/conda-forge/plum-dispatch-feedstock/pull/19 with the right SHA and version. If it builds then I'll merge it and the automatic version bump will work again.

wesselb:
Fantastic! Thanks for handling all the Conda stuff.

PR #23, opened 2021-07-22
Speedup parametric classes dispatch

PhilipVinc:
This PR implements what we had talked about a few times.
Namely, `@parametric` classes that don't need a custom `type_of` implementation that checks the type at _runtime_ can still benefit from using python's builtin type, which is much faster.

I have added a new argument to the decorator, called `runtime_typeof` (defaults to `False`).
When this is `False` the same dispatch logic as non-parametric classes is used.
If it is `True` then dispatch will use `plum.type_of`.

Ideally this could be picked up automatically, but the implementation would be relatively more complicated.

--

By the way, while going around the codebase I now realised that a paramtric type is an instance of `ParametricMeta`, which is a metaclass, but it is not an `AbstractType`.
This maybe makes sense, because `AbstractType` is a concrete type used internally in plum, but it's a bit confusing...
Probably it is not important, though

### DIFF
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -1,4 +1,5 @@
 import logging
+from functools import partial
 
 from .dispatcher import Dispatcher
 from .function import (
@@ -69,6 +70,10 @@ class ParametricTypeMeta(TypeMeta):
         """bool: Check whether the parametric type is instantiated or not."""
         return hasattr(cls, "_is_parametric")
 
+    @property
+    def runtime_typeof(cls):
+        return cls._runtime_typeof
+
 
 class CovariantMeta(ParametricTypeMeta):
     """A metaclass that implements *covariance* of parametric types."""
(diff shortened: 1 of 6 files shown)


plum/type.py
```diff
@@ -221,6 +225,11 @@ def parametric(self):
         # `_is_parametric` with the value `True`.
         return hasattr(self._type, "_is_parametric") and self._type._is_parametric
 
+    @property
+    def runtime_typeof(self):
+        # Types don-t specify runtime_typeof except for custom Plum Types
```
wesselb, line 230:
```suggestion
        # Types don't specify `runtime_typeof` except for custom Plum types.
```

plum/parametric.py
```diff
@@ -101,8 +106,21 @@ def __subclasscheck__(self, subclass):
...
-    """A decorator for parametric classes."""
+def parametric(Class=None, runtime_typeof=False):
+    """A decorator for parametric classes.
+
+    Optional keyword arguments:
+        runtime_typeof: If this type cannot be inferred by python's built-in`type`,
+            but you need to specialize `plum.typeof` (for example to inspect runtime)
+            values, then this must be set to True.
+            Functions that have this class in one of its method's signature will be
+            noticeably slower on dispatch.
```
wesselb, line 117:
```suggestion
    Args:
        runtime_typeof (bool, optional): If this type cannot be inferred by Python's
            built-in `type` and you need to specialise `plum.type_of`, for example to
            inspect runtime values, then this must be set to `True`. Functions that have
            this class in one of its method's signature will be noticeably slower on
            dispatch.
```

wesselb (changes requested):
Yes, this is much more sensible than the current approach.

Note, though, that the current name of the `type_of` function is `plum.type_of` rather than `plum.typeof`. Shall we name the current method `runtime_type_of` instead? Alternatively, we can globally rename `type_of` to `typeof`, but perhaps that's an undesirable change.

Also, does this change obsolete the `parametric` property of Plum classes? IIRC, that was only used to see if `type_of` needed to be sure. Since that's now done with  `runtime_type_of`/`runtime_typeof`, perhaps we can remove the whole `parametric` property.

wesselb:
> Though in practice it might be usefull in the future. Maybe let's just open an issue about deprecating .parametric if it turns out to be useless in the future?

That sounds reasonable. Let's do it that way.

Regarding your earlier

> This maybe makes sense, because AbstractType is a concrete type used internally in plum, but it's a bit confusing...
Probably it is not important, though

I agree. The distinction between internal Plum types and normal types is a bit of a mess... :(

For now, I think this PR looks good! Let's get this merged.

wesselb:
This is included in the latest release `v1.4.0`. Thanks! :)

PR #25, opened 2021-08-11
Fix bug introduced by #23

PhilipVinc:
In #23 I forgot to properly forward `runtime_type_of` in `Union` and `VarArgs` types, which was caught in [netket's tests](https://github.com/netket/netket/pull/855/checks?check_run_id=3296989199#step:7:4848).

Essentially, signatures like `def f(a: Union[Tuple[int]])` are not tagged as `runtime_type_of=True` and therefore dispatch fails.

This PR fixes the bug i introduced and adds relevant tests.

### DIFF
--- a/plum/type.py
+++ b/plum/type.py
@@ -59,8 +59,10 @@ class AbstractType(metaclass=TypeMeta):
         type."""
 
     @property
+    @abc.abstractmethod
     def runtime_type_of(self):
-        return False
+        """Boolean that indicates whether this type can only be determined at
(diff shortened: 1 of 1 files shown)


PhilipVinc:
@wesselb sorry for causing this

PhilipVinc:
If you could tag i'd love it!

wesselb:
This is included in the latest release `v1.5.1`. Thanks again, @PhilipVinc :)

PR #26, opened 2021-08-13
Autoversion

PhilipVinc:
I don't know if you'd like this (maybe not, so just close this down).

i wanted to add a `__version__` field to plum, but then thought to implement a more standard worflow...

This pr uses [setuptools_scm](https://github.com/pypa/setuptools_scm/) to autogenerate the version of a package from the latest git tag.
The version is added to files read by pypi as well as to a `plum/_version.py` file that is untracked by git.

It comes with a github action that autopublishes to PyPi as soon as you tag a release on github. To make it work you'd need to add your pypi username and password as secrets (see https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

The workflow is then the following:
 - if you `git clone` the repo, `pip install -e .` will create the `_version.py` file and install it.
 - if you want to tag a new release, go on github and tag it and the new release is automatically uploaded to pypi with the version corresponding to the tag.


PhilipVinc:
This would also require a minor change to the conda recipe to pick up the autogenerated version.

wesselb:
This is fantastic. I'm currently running some shell script which automates the steps, but this is much, much better. Thanks @PhilipVinc for constantly improving my developer quality of life!

wesselb:
What would need to change for the recipe? Is it just that the build requirements would need to be updated [here](https://github.com/conda-forge/plum-dispatch-feedstock/blob/master/recipe/meta.yaml)?

PhilipVinc:
I'll show you tomorrow . 
I can open the PR with the changes myself.

Just don't tag a version in the meantime otherwise the Fonda version gets a 00.0.0 version...

PR #31, opened 2021-11-01
[DIST] Build wheels for all platforms on GitHub actions

PhilipVinc:
Colleagues reported from time to time problems with install plum: sometimes you need to have python installed before doing `pip install plum-dispatch` otherwise pip fails with weird errors, if you have intel compilers compilation will fail because it does not accept some flags given to python...

distributing wheels might make it easier.

My plan is to first test this action (which is also used by sklearn and matplotlib, so it should be reliable) and check that the produced wheels work correctly.

Once this is done, I want to auto-upload the wheels when we publish to `pypi`

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -1,0 +1,24 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, windows-2019, macos-10.15]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.2.2
+        # to supply options, put them in 'env', like:
+        # env:
+        #   CIBW_SOME_OPTION: value
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./wheelhouse/*.whl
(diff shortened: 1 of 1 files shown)


PhilipVinc:
@wesselb would you be in favour of this?

wesselb:
@PhilipVinc, most certainly! This is fantastic. Thanks yet again for the QOL improvement. :)

Is this ready to go? (Should the `on: [push, ...` be something like `on: release: types: [published]`?)

Once this is ready to go, shall I merge it and then release a new version so the wheels will be built and uploaded?

PhilipVinc:
It's now ready to go.

I set it to run after we publish the source distribution, so yes, to trigger it you need to tag a new version (even if nothing changed).

It will take an 1h30 to run to generate all wheels for linux

wesselb:
Excellent, I'll now merge it and publish a new version.

.github/workflows/build-wheels.yml
```diff
@@ -0,0 +1,56 @@
+name: Build
+
+# Only trigger, when the build workflow succeeded
+on:
+  workflow_run:
+    workflows: ["CI build"]
```
PhilipVinc, line 6: Ops, this should have been
Publish to PyPI
`    workflows: ["Publish to PyPI"]`

wesselb:
@PhilipVinc, how is this triggered by the source publishing workflow? Does the below do that?

```
on:
  workflow_run:
    workflows: ["CI build"]
    types:
      - completed
```

(It's not clear to me what "CI build" refers to.)

.github/workflows/build-wheels.yml
wesselb: Gotcha, let me fix that.

PR #33, opened 2021-11-29
build wheels for macOS M1 too

PhilipVinc:


wesselb:
Very sweet. Thank you @PhilipVinc, as so many times before!

PR #34, opened 2021-12-04
make @parametric work with classes with arbitrary metaclasses

PhilipVinc:
I am experimenting with https://github.com/cgarciae/treeo and plum, and found out that since both inject meta classes they don't work together.

This PR makes plum work with classes which already have a meta class by injecting `CovariantMeta` as a superclass of the existing meta class, and by redefining `__call__` (essentially the constructor) to infer the parameters before calling into the existing class.

I'm surprised this was so easy to implement!

### DIFF
--- a/README.md
+++ b/README.md
@@ -978,6 +978,14 @@ class NTuple:
 __main__.NTuple[3, <class 'int'>]
 ```
 
+**Note:** The `@parametric` decorator injects a metaclass (`plum.parametric.CovariantMeta`) into the 
+decorated class. If the decorated class already has a metaclass, it creates a new metaclass which
+inherits from both the existing metaclass and `plum.parametric.CovariantMeta`.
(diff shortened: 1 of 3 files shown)


wesselb:
Thank you, @PhilipVinc!! This really is excellent—a strict improvement. Merging this now and releasing a new version. :)

wesselb:
Actually, you've still marked this as a draft. The PR looks complete to me. Are you happy to have this merged?

PhilipVinc:
Yeah, I think I wanted to add a test to check for what the `fix issue in __init_subclass__` commit fixed.
I found out about it in a very exotic case that arises when you combine `@parametric` with classes that auto-register themselves with jax, but it's a real issue of plum,.

I'll add a test soon

wesselb:
The contents of this PR will be included in #68, so closing this.

Issue #38, opened 2022-02-08
FR: Make plum work with ipython's autoreload


PhilipVinc:
The two don't work well together. 
I think it's because autoreload changes the classes of objects, but for some reason forgets to update the Types in plum's dispatch tables.

Should be investigated...

wesselb:
Also tagging @seeM here.

This issue is unfortunately a tricky one to fix. Just to document what's going wrong, here's a minimal example.

Consider the minimal module

```python
from __future__ import annotations


class A:
    def test(self) -> A:
        # Hello
        return self
```

We now have the following behaviour:

```python
In [1]: from core import A

In [2]: from typing import get_type_hints

In [3]: get_type_hints(A.test)["return"] == A
Out[3]: True

In [4]: # Make changes and trigger autoreload

In [5]: get_type_hints(A.test)["return"] == A
Out[5]: False
```

(This is on Python 3.9.)

I think it's necessary to carefully pin down how `get_type_hints` interacts with `autoreload`, especially with regards to forwards references.

If this particular case can be fixed, I think good support for `autoreload` in the case of `from __future__ import annotations` should be possible.

PhilipVinc:
I had investigated the issue a bit, and eventually realized that the best course of action would be for someone to contact the developers of iPython to ask them to add an hook to auto reload so that when he reloads a class/files plugins can ask to be notified about it.

wesselb:
@PhilipVinc That's interesting. I suppose hooking into autoreload would be the way go about this: upon every autoreload, clear all cache.

I'm wondering, is the above example expected behaviour? Should `get_type_hints(A.test)["return"] == A` after autoreloading? The locals in an iPython sessions should be updated with every autoreload, right?

Perhaps not relevant to this issue, but I'm tempted to strip out all current functionality to handle forward references from the package and just rely on `typing.get_type_hints` to handle forward references. What are your thoughts on this?

wesselb:
Experimental support thanks to #44 by @PhilipVinc is included in the latest release `v1.6`. :)

PR #44, opened 2022-06-06
autoreload support

PhilipVinc:
Addresses #38

This is a hack implementation monkey patching IPython autoreload to make it work (partly) with plum.

Example: first create a module
```
echo "class A:
    sz = 1" >>ar.py
```

then run IPython
```python
%load_ext autoreload
%autoreload 2

import plum
# load explicitly this monkey patching module
from plum import autoreload

import ar

@plum.dispatch
def test(a: ar.A):
	return 1

a = ar.A()

print(test(a))
```

now modify `ar.py` in any way, and run
```python
print(test(a))
```
and it will work.

--

## Implementation:

Autoreeload uses `gc.get_referrers(class)` to get all objects referring to a class, which includes all its instances, and updates their class using the function `update_instance`. 

```python
def _update_instances(old, new):
    refs = gc.get_referrers(old)
    for ref in refs:
        if type(ref) is old:
            ref.__class__ = new
```

To make this work with plum's custom types I needed to do 3 things: 
 - Make `gc.get_referrers(class)` show up all `plum.types.Type` objects holding a reference to the given class. This was tricky: `Type(class)` holds a reference to `class` in a subfield `_type` constructed at `__init__` time. But what actually python does is that it stores the field as a key in the dictionary of the instance `Type(class).__dict__["_type"] = class`, so `get_referrers` actually returns the dictionary and it's hard to tell that this dictionary belongs to Type.
Instead, by declaring the field a `slot` (which is an helpful performance enhancement anyway) the field is stored inline and get referrers actually return the Type instance. Unfortunately, while this trick works well for `_type`, this does not work for `plum.types.Union` types because they stored a tuple holding all classes, and therefore the `gc` would return that tuple.
 - Make `update_instance` update the plum types. This is achieved by brutally monkey patching the method.
 - Clear the cache. This is also done by pulling off the same trick using the gc. Note that this is probably inefficient but works...


PhilipVinc:
To avoid monkey patching, we should ask IPython developers to provide us some hooks

wesselb:
This is really great! I've tested it out and it indeed fixes autoreloading for me locally. :) 

For the union type, we could potentially subtype `tuple` and store the types with that?

PhilipVinc:
>For the union type, we could potentially subtype tuple and store the types with that?

Actually I just realised there's no need for that: Union types contain a tuple of Plum Types, and by updating the plum types everything's fixed.

>This is really great! I've tested it out and it indeed fixes autoreloading for me locally. :)

Now the underlying question is how to turn this on?
Should we 
 - require users to explicitly import `from plum import autoreload` or implement a magic like `% autoreload_plum` 
 - should we automatically do this if an user imports plum *After* loading autoreload
 - should we do this always unlesss an ENV variable is specified
 - should we never do this unless an ENV variable is specified.

By checking the history of autoreload, it seems it has been quite stable and has not changed much over the years, but this is still monkey patching so... what do you think would be the best way to go?

Maybe @seeM and/or @gasparyanartur would be interested in commenting on this?

I would be naturally tempted to say "what this is doing is morally wrong and pirating" so we should hide this behind explicit user-action (`from plum import autoreload`) but as a library author, my users are not so-well-aware that we're using `plum` under the hood, so I would rather prefer to have it always be enabled by default, but I am unsure if that would be a great idea..

seeM:
Very neat! That’s a tricky choice. I personally wouldn’t mind this being enabled by default unless an env variable is set.

I only use autoreload when developing, never when something is deployed or when running tests. Given the dynamic nature of Python, it’s not unusual that autoreload gets things into a weird state and I need to restart my kernel, so I wouldn’t mind if plum complicated it a little more.

That said, I’d personally try this out for a bit to test for stability and any weird behaviour before making it the default.

Keen to hear more peoples thoughts. I could ask in the fastai Discord as well.

gasparyanartur:
As a disclaimer, I have only used iPython for course work, so I can only offer my layman perspective. 

The situations is a bit of a dilemma. On the one hand, users of a library should ideally not be bothered by implementation details from another, but on the other hand, this could potentially introduce bugs and make it harder to maintain, since there is no guarantee that iPython's implementation will remain the same. If style is of concern, the Zen of Python seems to prefer explicit over implicit. 

Personally, if I was to use a library, I would rather explicitly import something than have it potentially break existing functionality. However, as a compromise, I do like the idea of it being on by default but with a possibility to disable it with ENV-variables. But do take my perspective with a grain of salt, since I haven't used autoreload very much.

wesselb:
This is fantastic! Thanks @PhilipVinc for this!!

I think the two ways of activating it are very sensible choices. Merging this right away. :)

wesselb:
This is included in the latest release `v1.6`.

PR #60, opened 2022-08-24
support int|float union syntax

### LINKED ISSUE #59: Add support for "|" operator in union types
PEP 604 allowed for the writing of union types as  `x  |  y`
but this break in plum.dispatch:
```
@plum.dispatch
def foo(x: int | float):
    return x + 1
```
```
>>> foo(3.14)
RuntimeError: Could not convert "int | float" to a type.
```

PhilipVinc:
Closes #59 

cc @alex-janss

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.10]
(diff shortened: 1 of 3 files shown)


PhilipVinc:
@wesselb ready to go.
I can't get coverage to 100% on py<3.10 but overall it is 100%

wesselb:
This looks great!! As any times before, thanks @PhilipVinc!

PR #61, opened 2022-08-24
fix pyproject.toml build settings

PhilipVinc:
@wesselb when you rewrote the pyproject.toml you removed too many things required by the build system. Those are often installed but I test from time to time plum in sandboxed environments where setuptools is not present unless you require it.


wesselb:
Ah, damn it. I wasn't sure whether I could remove those. Thanks for catching this, and sorry for the screw up!

PR #63, opened 2022-09-26
Fix typo in comparison.md

PabloLION:
Nice lib! Would be amazing if it works as docs!
I think here was a typo when I read the docs.
Thanks for building this again!

### DIFF
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -50,7 +50,7 @@ class A:
         return "fallback"
         
 
-class B:
+class B(A):
     @dispatch(int)
     def f(self, x):
         return x


wesselb:
Hey @PabloLION! First of all, I'm very sorry for the radio silence.

Thanks for catching the typo. :) You're 100% right, and I'm merging this right away!

I hope the package is useful to you! Please do open an issue if something doesn't work as expected.

PabloLION:
@wesselb I tried it out and it's amazing, best function overloading package in Python!

wesselb:
I’m very glad to hear that! 😊 Thank you!!

PR #67, opened 2022-10-21
Update setup.py

PhilipVinc:
@wesselb  Sorry, that was totally my fault...
I think this is a better way to do it, because even if we add folders or files later on they are ignored.


wesselb:
Not a problem at all!! Thanks for the PR. This looks like a much better solution. :)

PR #73, opened 2023-02-19
[WIP] Plum 2

### LINKED ISSUE #10: Hack some way to dispatch on number of dimensions of numpy/jax arrays
It would be nice (even though admittedly hacky) if we could dispatch on the number of dimensions of a jax/numpy array, which can be probed with `.ndim` on an instance.

This is admittedly not part of the type information in python, but maybe we can hack it in?

I was thinking of creating a custom parametric type for signatures, but then the problem is resolving the call signature and bringing this information from the value domain to the type domain.

This happens in `parametric.py` if I am not mistaken, and would require changing this function to allow hacking in some custom types... 
Do you have any idea on what would be the best way to implement this?

```python
def type_of(obj):
    """Get the Plum type of an object.

    Args:
        obj (object): Object to get type of.

    Returns
        ptype: Plum type of `obj`.
    """
    if isinstance(obj, list):
        return List(_types_of_iterable(obj))

    if isinstance(obj, tuple):
        return Tuple(*(type_of(x) for x in obj))

    return ptype(type(obj))
```

### LINKED ISSUE #28: Support for `Protocol`s
[`Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol)s were introduced in Python 3.8 by [PEP 544](https://www.python.org/dev/peps/pep-0544/), and it would be awesome if Plum could handle them.

A simple use case would look like this:
```py
from typing import Any, Protocol
from plum import dispatch

class Named(Protocol):
    name: str

class Person:
    def __init__(self, name: str, age: int) -> None:
        self.name: str = name
        self.age: int = age

@dispatch
def greet(anyone: Any) -> None:
    print('Hi!')

@greet.dispatch
def _greet_named(named: Named) -> None:
    print(f'Hello, {named.name}')

me = Person('Ruan', 24)
greet(me)
# should print "Hello, Ruan"
# currently throws the following error:
# TypeError: Instance and class checks can only be used with @runtime_checkable protocols
```

Another use case is what I am actually interested in: using [dataclasses](https://docs.python.org/3/library/dataclasses.html). Dataclasses can be identified by a special attribute: `__dataclass_fields__` (see python/mypy#8578), so that I could write
```py
class Dataclass(Protocol):
    __dataclass_fields__: Dict[str, Any]

@dispatch
def do_something(dataclass_instance: Dataclass):
    ...
```

I have no idea how to proceed to implement this, but I'm interested in seeing this in Plum and would love to help in any way I can.
Just for the record, [Dry Python Classes](https://github.com/dry-python/classes) has this feature, we could take some inspiration from them.

### LINKED ISSUE #48: Issue when the output is the a Generator
It seems that the plum has issue in dispatch function when the output is Generator

I have the sample code as
```
from typing import Generator, List
from plum import dispatch
class Convert():
    @dispatch
    def run(self, input: List[int, ]) -> Generator[int, None, None]:
        for item in input:
            yield int(item)

    @dispatch
    def run(self, input: List[float, ]) -> Generator[float, None, None]:
        for item in input:
            yield float(item)
```

I got
```
  File "/Users/zwu/miniconda3_x86/envs/exs-orionbssfloes/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3552, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-26-2a8843a2d5ae>", line 1, in <module>
    class Convert():
  File "<ipython-input-26-2a8843a2d5ae>", line 3, in Convert
    def run(self, input: List[int, ]) -> Generator[int, None, None]:
  File "/Users/zwu/miniconda3_x86/envs/exs-orionbssfloes/lib/python3.7/site-packages/plum/dispatcher.py", line 43, in __call__
    signature, return_type = extract_signature(method)
  File "plum/function.py", line 120, in plum.function.extract_signature
  File "/Users/zwu/miniconda3_x86/envs/exs-orionbssfloes/lib/python3.7/site-packages/plum/type.py", line 409, in ptype
    f'There is currently no support for "{obj}". '
NotImplementedError: There is currently no support for "typing.Generator[int, NoneType, NoneType]". Please open an issue at https://github.com/wesselb/plum/issues
```

### LINKED ISSUE #53: The `type_of` could be very slow
Hi, thanks for the nice project! 

I found that the [_types_of_iterable](https://github.com/wesselb/plum/blob/fb1ecf1733147d23fbb08d3f37b65425d76e9465/plum/parametric.py#L613) could be very slow if the iterable is very large. This could be a big problem for practical application. Is there any solution for this?

### LINKED ISSUE #54: How to handle doc strings
How do doc strings work with several methods? It would be nice to put different doc strings in different methods. I don't think this works.

Is there a recommended way then ? Maybe put one big doc string in the first method?

### LINKED ISSUE #66: [FeatReq] support for literal types
This seems to be mentioned in the discuss of https://github.com/wesselb/plum/issues/53#issuecomment-1220323266
While I was trying to dispatch with `typing.Literal[True]` it fails.
Can we add this functionality?
TIA.

### LINKED ISSUE #68: Plum 2
### What is This?
This issue proposes possible major improvements for a version two of the package.  The goal is to start a discussion around these proposals. If you agree or disagree with some or all of changes, or think that other improvements are missing, then please do share your thoughts by commenting below. :) Tagging @PhilipVinc, @tbsexton, @leycec, and @seeM here since they have been involved in recent relevant discussions.

None of the proposed changes should change current behaviour, thus maintaining backwards compatibility. I'll be continuously updating and adding to this issue over time.

The list of proposed improvements is as follows:

### Instance-Check-Based Dispatch
Instead of performing dispatch based on the types of the arguments, perform dispatch entirely using `isinstance`, hence fully eliminating the need to use `type_of`. [See this proposal by @tbsexton.](https://github.com/wesselb/plum/issues/53#issuecomment-1274235670) The massive benefit of doing things this way is it becomes _much_ easier to support types like `Literal`s (#66), `Protocol`s (#28), `numpy` arrays with a specific number of dimensions (#10), and `PyTree`s ([Stheno #21](https://github.com/wesselb/stheno/issues/21)).

The big downside of this change is that performance will take a hit, since instance checks will need to happen every time dispatch happens. However, by using @beartype (@leycec), instance checks will be blazingly fast, so the performance hit should be minimal. Moreover, caching is still possible for functions with a so-called _faithful dispatch tree_ ([see here](https://github.com/wesselb/plum/issues/53#issuecomment-1288103843)), so we should get the best of both worlds!

Although dispatch will not rely on `type_of` anymore, other functionality of the package still does, such as [promotion](https://github.com/wesselb/plum/blob/master/plum/promotion.py). Hence, a `type_of` function is still necessary.

### Postponed Evaluation of Types
[@beartype](https://github.com/beartype/beartype/issues/7#issuecomment-1251978495) now includes `beartype.peps.resolve_pep538()`, which should handle PEP538-style postponed types for us. A massive thanks to @leycec for this!! This allows us to get rid of all use of `inspect`, which I believe currently is causing performance issues ([Stheno #25](https://github.com/wesselb/stheno/issues/25)). Deferring the responsibility of handling postponed evaluation of types to @beartype should _significantly_ simplify the internals of the package.

### Wrapping of Types
Internally Plum currently wraps types in its own types. We should get rid of those entirely and use types from `typing`. This does pose a challenge for some more advanced custom types, such as `PromisedType`.

### Compatibility with Modern Tooling
Whatever Plum does should be compatible with modern tooling. Think `mypy`, `pylance` (#51), and possible others.

wesselb:
Status: done! Ready to go?!

_Massive_ simplification of the internal design. Closes #68. Gets rid of wrapping types into "Plum types" and other unnecessary complexity.

Breaking changes:
* Hopefully, nearly none. If everything goes according to plan, this release should be nearly backwards compatible.
* Since `type_of` is now gone (see below), `runtime_type_of` is gone too from the API of `parametric`.
* Renames `autoreload.activate` to `activate_autoreload` and `autoreload.deactivate` to `deactivate_autoreload`.

Changes:
* The tests are now setup in a more systematic way. This caught a few subtle bugs.
* Fully depends on @beartype to handle the hard stuff, which means that we may now enjoy the fruits of @leycec's hard work!
    * Fixes #10. This could be implemented simply by using types such as from [`nptyping`](https://github.com/ramonhagenaars/nptyping). The problem is that these types currently do not yet support subclass checks, which means that they are not usable as is. One solution is to extend `nptyping`; another is to hack something together. See the example below.
    * Fixes #28. Not actually fixed, but we now support `Protocol`s to the extent @beartype does.
    * Fixes #48. As above.
    * Fixes #66. Fully supported.
* Gets rid of `type_of` and therefore all associated issues (@tbsexton). 
   * Fixes #53.
* Merges #34, with some changes. (Closes #34.)

New features:
* Union aliases, which must be enabled explicitly. See below.
* Better support docstrings. (Fixes #54.)

Outstanding issues:
* Plum still cannot really conveniently be used with tooling like `mypy`. I'm still not sure what the best way to address this is.

## Show me Something Cool!

It will finally be possible to properly dispatch on NumPy array shapes and sizes without making major concessions. The below example will be functional:

```python
from plum import dispatch, parametric
from typing import Any, Optional, Tuple, Union

import numpy as np


class NDArrayMeta(type):
    def __instancecheck__(self, x):
        if self.concrete:
            shape, dtype = self.type_parameter
        else:
            shape, dtype = None, None
        return (
            isinstance(x, np.ndarray)
            and (shape is None or x.shape == shape)
            and (dtype is None or x.dtype == dtype)
        )


@parametric
class NDArray(np.ndarray, metaclass=NDArrayMeta):
    @classmethod
    @dispatch
    def __init_type_parameter__(
        cls,
        shape: Optional[Tuple[int, ...]],
        dtype: Optional[Any],
    ):
        """Validate the type parameter."""
        return shape, dtype

    @classmethod
    @dispatch
    def __le_type_parameter__(
        cls,
        left: Tuple[Optional[Tuple[int, ...]], Optional[Any]],
        right: Tuple[Optional[Tuple[int, ...]], Optional[Any]],
    ):
        """Define an order on type parameters. That is, check whether
        `left <= right` or not."""
        shape_left, dtype_left = left
        shape_right, dtype_right = right
        return (
            (shape_right is None or shape_left == shape_right)
            and (dtype_right is None or dtype_left == dtype_right)
        )


@dispatch
def f(x: np.ndarray):
    print("Any NP array!")


@dispatch
def f(x: NDArray[(2, 2), None]):
    print("A 2x2 array!")


@dispatch
def f(x: NDArray[None, int]):
    print("An int array!")


@dispatch
def f(x: NDArray[(2, 2), int]):
    print("A 2x2 int array!")


f(np.ones((3, 3)))       # Any NP array!
f(np.ones((3, 3), int))  # An int array!
f(np.ones((2, 2)))       # A 2x2 array!
f(np.ones((2, 2), int))  # A 2x2 int array!
```

## Union Aliases

I've been wanting this feature for a long time. It must be enabled explicitly, since it messes with `typing.Union.__repr__`:

```python
from typing import Union
from plum import activate_union_aliases, set_union_alias

activate_union_aliases()
set_union_alias(Union[int, str], alias="IntStr")
```

```python
>>> Union[int, complex]
typing.Union[int, complex]

>>> Union[int, complex, str]
typing.Union[IntStr, complex]
```

One quirk is that is aliases will always be printed within a `typing.Union`:

```python
>>> Union[int, str]
typing.Union[IntStr]
```

This is delibrate and is to prevent code which parses how unions print from breaking. E.g., just printing `IntStr` would eliminate the `[` and `]`.

## Better Support for Docstrings

```python
from numbers import Number

from plum import dispatch

@dispatch.abstract
def f(x: Number, y: Number):
    """Multiply two numbers."""


@dispatch
def f(x: float, y: float):
    """Multiply two floats."""
    return x * y


@dispatch
def f(x: int, y: int):
    """Multiply two ints."""
    return x * y

```

Then calling `help(f)` produces

```
Help on Function in module __main__:

f(x: numbers.Number, y: numbers.Number)
    Multiply two numbers.

    This function has further implementations documented below.

    f(x: float, y: float)
        Multiply two floats.

    f(x: int, y: int)
        Multiply two ints.
```


wesselb:
@PhilipVinc At last, I've managed to put together the changes I've had in mind for a long time. Would you mind checking whether this release is really backwards compatible on your end?

leycec:
**Phenomenal.** It's fascinating that you dramatically simplified the codebase, resolved outstanding issues, *and* implemented a swath of new features. Usually, those are competing design goals. But you had your plum cake and ate it, too! :cake: 

The NumPy example is so elegantly concise. I'm not even sure what's going on there, frankly... but I like it.

## Now I'm Gonna Bore Everyone (Including Myself)

As an ignorable nitpick, `dtype: Optional[type]` is *probably* a bit too permissive. @beartype itself struggled with proper `dtype` detection until [I implemented a painfully non-trivial `numpy.typing.NDArray` handler](https://github.com/beartype/beartype/blob/398823663d8fa0df7874d7b19060a8a8003345cf/beartype/_util/hint/pep/mod/utilmodnumpy.py#L48). The effort was so exhausting that I simply gave up on supporting any *other* non-standard NumPy type hints. To summarize matters:

* `typing.Any` is a valid dtype unsurprisingly matching all possible dtypes. Don't ask. I didn't. What we *should* ask is: "Uhm. How can I annotate a type hint as accepting `Any`?" I... don't actually know the answer to that. `typing.Literal` only supports six types. `Any` is not one of them. Moreover, the runtime implementation of `Any` varies wildly across Python versions. Under new Python versions, it's a standard class; under older Python versions, it's a non-standard `_SpecialForm` singleton. In other words... let's pretend that `typing.Any` is *not* a valid dtype.
* Ignoring `typing.Any`, it's still excrutiatingly non-trivial to match the set of all valid dtypes at runtime. As far as I can tell, a `dtype` is any object that either:
  * Can be passed as is to the `numpy.dtype.__init__()` constructor *without* raising an exception.
  * Has a `type` attribute whose value is a subclass of one or more abstract base classes (ABCs) in the set `NUMPY_DTYPE_TYPE_ABCS` (defined below):

```python
from numpy import (  # pyright: ignore[reportMissingImports]
    character,
    complexfloating,
    flexible,
    floating,
    generic,
    inexact,
    integer,
    number,
    signedinteger,
    unsignedinteger,
)

# Frozen set of all NumPy scalar data type abstract base classes.
NUMPY_DTYPE_TYPE_ABCS = ((
    character,
    complexfloating,
    flexible,
    floating,
    generic,
    inexact,
    integer,
    number,
    signedinteger,
    unsignedinteger,
))
```

Basically, it's madness. But `dtype: Optional[type]` sadly is *not* right, now that I've delved beyond the Mountains of Madness above. Although it just reduces to a noop, `dtype: Any` is probably the best that can be done at the moment.

Of course, NumPy *also* provides a [`numpy.typing.DtypeLike`](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.DTypeLike) type hint. If that hint actually worked at runtime, that would solve all your problems. You could just write `dtype: Optional[numpy.typing.DtypeLike]`. But last I checked, that hint did *not* work at runtime. Maybe it does now? It's feels like forever since I submitted a NumPy issue about that.

I exhale with exhaustion. :face_exhaling:

patrick-kidger:
This is cool to see! (Thanks to @mthiboust for pointing me at the fact that this I have a whole pile of questions.

---

Regarding array types.

> This could be implemented simply by using types such as from nptyping. The problem is that these types currently do not yet support subclass checks, which means that they are not usable as is.

So [jaxtyping](https://github.com/google/jaxtyping) provides array types, parameterised by shape and dtype, that *do* support `isinstance` checks. Despite the name it also supports NumPy/PyTorch/etc, and works with both runtime and static type checking out-of-the-box, so it's now my recommended approach for all kinds of array typing.

I'm a big fan of Plum. Based on #68 (that all checks happen via `isinstance`), my hope was that jaxtyping and Plum v2 would work smoothly with each other out-of-the-box!

I can see here that you mention subclass checks. Do I understand correctly here that it would also be helpful if e.g. `issubclass(jax.Array, jaxtyping.Float[jax.Array, "size1 size2"]) == True`? (Morally speaking that's seems backward, of course?) I'd be happy to tweak jaxtyping if it is needed for compatibility.

---

Union aliases: what's the use case for this, out of curiosity?

---

A question for @leycec: last I checked, beartype checked random container elements. In order to perform reliable dispatch here, does beartype now provide a mode for comprehensive checking? (A la typeguard.)

If so then that'd be great. My personal use cases involve a tracing JIT compiler, so until now I've been preferring the reliablity of typeguard -- the overhead doesn't affect me since the checking only happens at trace (compile) time without impacting runtime.

---

Dtype handling: blech, I've bumped into this problem too. I sympathise.

For that matter:
```python
import numpy as np
array = np.array(1, dtype=np.float32)
print(array.dtype)        # dtype('float32')
print(array.dtype.type)   # <class 'numpy.float32'>
print(type(array.dtype))  # <class 'numpy.dtype[float32]'>
```
For some reason numpy has three distinct objects representing the same concept!

tbsexton:
Chiming in here since I was involved in #68 discussion a bit; @patrick-kidger and @wesselb I may be remembering wrong but I'm pretty sure the new system is using something _like_ `isinstance`, but is actually casting to beartype `TypeHints` to reason about things... a la the `beartype.is_bearable()` function. They're all instance checks, yes, but the logi behind them bypasses native `isinstance` (because, as @leycec would say, that is slow :sweat_smile:)

FOr `issubclass(jax.Array, jaxtyping.Float[jax.Array, "size1 size2"]) == True` being useful, you'd probably get a kick out of reading #53 ! THere's a...um... a *lot* there, but fundamentally I think @wesselb has a caching-based speedup if a type is "faithful", i.e. that 

```python 
isinstance(x, Type) == issubclass(type(x), Type)
```

This is complicated by the fact that beartype handles waaaaay more cases and odd situations at runtime than those functions do, so really I think you should check in beartype that 

```python 
is_bearable(x, Type) == TypeHint(type(x))<=Type
```

So now the key issue is when `type(x)` lies to us, which is _often_ (again, that thread has loads of examples), and this is why I'm pretty sure Plum is relying on `is_bearable` much more directly. (Totally correct me if I'm wrong here, guys!)

I think you can guarantee that `jaxtyping` provides faithful types by using beartype's [upcoming plugin api](https://github.com/beartype/beartype/issues/192) that gives us a protocol to allow user-created types to report their own runtime `TypeHints` that plum can then check against a `Type`, which will be "faithful" by design. @leycec ?

wesselb:
@leycec 

> Although it just reduces to a noop, dtype: Any is probably the best that can be done at the moment.

I was aware typing for data types was a mess, but I wasn't aware that it was this big of a mess. 😩 `Any` it is! `numpy.typing.DTypeLike` indeed does not support instance/subclass checks. If only life were so simple...

> I'd like to help boost Plum's visibility. This is an open-ended offer that you may graciously refuse at any time – especially including now. But... would you like to join the https://github.com/beartype organization?

_Whoa!_ This is an _incredibly_ kind offer, which I very much appreciate. I've sent you an email to discuss the matter on a more private medium. :)

---

@patrick-kidger 

> I'm a big fan of Plum. Based on https://github.com/wesselb/plum/issues/68 (that all checks happen via isinstance), my hope was that jaxtyping and Plum v2 would work smoothly with each other out-of-the-box!

I believe that it should! If a type correctly implements `isinstance` and `issubclass`, then, from now on, that type should be compatible with Plum. I will try this myself shortly, but it would be _amazing_ if `jaxtyping` were supported!

> I can see here that you mention subclass checks.

Subclass checks are used to determine whether a type is a subtype of another type. I.e., if `issubclass(t1, t2)` then `t1` is assumed to be a subtype of `t2`. (Actually, most type hints don't support subclass checks, and that's where @beartype comes in! If, however, a type supports subclass checks, then that should suffice.) In your example, I believe we would want

```python
assert issubclass(jaxtyping.Float[jax.Array, "size1 size2"], jax.Array) 
assert not issubclass(jax.Array, jaxtyping.Float[jax.Array, "size1 size2"]) 
```

since a `jaxtyping.Float[jax.Array, "size1 size2"]` is a `jax.Array`, but not the other way around. As alluded to above, under the hood Plum uses @beartype, so what would actually be required is

```python
from beartype.door import TypeHint

assert TypeHint(jaxtyping.Float[jax.Array, "size1 size2"]) <= TypeHint(jax.Array) 
assert not TypeHint(jax.Array) <= TypeHint(jaxtyping.Float[jax.Array, "size1 size2"]) 
```

> Union aliases: what's the use case for this, out of curiosity?

Another project of mine is [LAB](https://github.com/wesselb/lab), where I try to define types like `B.Numeric` and `B.Int`, which would be types for general numerical objects and integers. (I think this pursues similar goals as `jaxtyping`!) Under the hood, `B.Numeric` and `B.Int` will be some very large union. When printing `B.Int` or when showing them in a rendered version of a docstring, instead of showing that whole union, I would just like to see `B.Int`.

> A question for @leycec: last I checked, beartype checked random container elements. In order to perform reliable dispatch here, does beartype now provide a mode for comprehensive checking? (A la typeguard.)

I believe there is such a mode! See [this part of the @beartype docs.](https://beartype.readthedocs.io/en/latest/api_decor/#beartype.BeartypeStrategy)

---

@tbsexton 

Yes, you are completely right!

Under the hood, the spirit of what's happening is a lot of `isinstance` and `issubclass` calls, but @beartype is used to make these calls work for type hints as well:

```
isinstance(x, t)   ->  is_bearable(x, t)
issubclass(t1, t2) -> TypeHint(t1) <= TypeHint(t2)
```

As you very correctly explain, the PR also implements the concept of "faithful" types, which are types `t` such that, for all `x`,

```python
isinstance(x, t) == issubclass(type(x), t)
```

An example of a faithful type is `int`. An example of an unfaithful type is `Literal[1]`.

Regardless of whether a type is faithful or not, dispatch should _always_ function correctly. If, however, a type is faithful, and Plum knows it is, clever caching can be used to also make dispatch super fast.

> I think you can guarantee that jaxtyping provides faithful

Oof, that would be _super_ nice, but might be difficult to achieve. For example, `type(np.array([1, 2])) == np.ndarray`, which loses shape and data type information. I'm not sure how you would work around this issue.

patrick-kidger:
Marvellous. Thanks for the all the answers. I've just updated jaxtyping; the latest HEAD should now support `issubclass`. I'll also do a new release in a few days once I've checked that nothing has broken internally.

Let me know if this seems to work with Plum v2.

Agreed that I don't think there's any way to make jaxtyping provide faithful types, unfortunately.

wesselb:
@patrick-kidger, that's fantastic! It seems to work perfectly. :)

```python
import jax
import jaxtyping as jt

from plum import dispatch


@dispatch
def f(x: jax.Array):
    return "An array!"


@dispatch
def f(x: jt.Float[jax.Array, "size1 size2"]):
    return "A 2D array of floats!"
```

```python
>>> import jax.numpy as jnp

>>> f(jnp.ones(5))
'An array!'

>>> f(jnp.ones((5, 5)))
'A 2D array of floats!'

>>> f(jnp.ones((5, 5), dtype=int))
'An array!'
```

wesselb:
Alright, I believe this PR is now ready to go. Hurray! :)

I may give it another week or so to double check compatibility with existing code. If there are no objections by then, I will proceed to merging this and tagging a new release.

wesselb:
@leycec I'm considering making the `O(n)` the default which is used in Plum and to let the users opt into `O(1)` or `O(0)`. I know you've got more than enough on your plate already, but I was wondering whether the `O(n)` strategy is on your radar. Please don't consider this a high priority request! The plan is to move forward with the `O(1)` strategy in the current release (but already opt into the `O(n)` strategy).

Issue #78, opened 2023-03-13
Missing `plum.__version__` attribute


patrick-kidger:
This isn't technically necessary, but it's a common convention to define a `__version__` attribute on Python packages, that can be checked at runtime. Currently Plum doesn't define one.

pabloferz:
AFAICT, the only way to access this for now is like this:
```python
from plum._version import __version__ as __plum_version__
```

wesselb:
I believe #79 by @PhilipVinc should have addressed this. :)

PR #79, opened 2023-03-16
Move to PyProject.toml, autogenerate version string

PhilipVinc:
Addresses #78, and moves towards the `pyproject.toml` removing the `setup.py` file which is now deprecated.. 

This removes the `requirements.txt` file which is also non standard by now. extra dependencies can be installed by specifying `pip install 'plum-dispatch[dev]'` or `pip install -e '.[dev]'`.

Uses hatch/hatchling for declaring the build steps (not that there's any build step) as it's recommend by the official python docs, and is much more lightweight than poetry..


wesselb:
Ahhh, this is fantastic! Thanks, @PhilipVinc. :) I’ll be merging this right away. Thank you for yet again making our lives a bit better!

PR #91, opened 2023-08-10
Add a few type hints to the codebase

gabrieldemarmiesse:
I added some type hints, not all. We can add other type hints later on in other pull requests. I recommend sqash and merge as the merging method as I'm not fond of rewriting my branch history (a bit dangerous).

### DIFF
--- a/plum/dispatcher.py
+++ b/plum/dispatcher.py
@@ -2,7 +2,7 @@ from typing import Callable, Dict, Optional, Tuple, Union
 
 from .function import Function
 from .signature import Signature
-from .util import get_class, is_in_class
+from .util import TypeHint, get_class, is_in_class
 
 __all__ = ["Dispatcher", "dispatch", "clear_all_cache"]
 
@@ -36,7 +36,7 @@ class Dispatcher:
         # set the signature argument to `None`.
         return self._add_method(method, None, precedence=precedence)
 
-    def multi(self, *signatures: Union[Signature, Tuple[type, ...]]) -> Callable:
+    def multi(self, *signatures: Union[Signature, Tuple[TypeHint, ...]]) -> Callable:
         """Decorator to register multiple signatures at once.
 
         Args:
--- a/plum/dispatcher.py
+++ b/plum/dispatcher.py
@@ -1,4 +1,4 @@
-from typing import Callable, Optional, Union
+from typing import Callable, Dict, Optional, Tuple, Union
 
 from .function import Function
 from .signature import Signature
@@ -17,8 +17,8 @@ class Dispatcher:
     """
 
     def __init__(self):
-        self.functions: dict[str, Function] = {}
-        self.classes: dict[str, dict[str, Function]] = {}
+        self.functions: Dict[str, Function] = {}
+        self.classes: Dict[str, Dict[str, Function]] = {}
 
     def __call__(self, method: Optional[Callable] = None, precedence: int = 0):
         """Decorator to register for a particular signature.
@@ -36,7 +36,7 @@ class Dispatcher:
         # set the signature argument to `None`.
         return self._add_method(method, None, precedence=precedence)
 
-    def multi(self, *signatures: Union[Signature, tuple[type, ...]]) -> Callable:
+    def multi(self, *signatures: Union[Signature, Tuple[type, ...]]) -> Callable:
         """Decorator to register multiple signatures at once.
 
         Args:
--- a/plum/dispatcher.py
+++ b/plum/dispatcher.py
@@ -1,4 +1,4 @@
-from typing import Optional, Callable, Union
+from typing import Callable, Optional, Union
 
 from .function import Function
 from .signature import Signature
@@ -88,7 +88,12 @@ class Dispatcher:
 
         return namespace[name]
 
-    def _add_method(self, method: Callable, *signatures: Optional[Signature], precedence: Optional[int]) -> Function:
+    def _add_method(
+        self,
+        method: Callable,
+        *signatures: Optional[Signature],
+        precedence: Optional[int],
+    ) -> Function:
         f = self._get_function(method)
         for signature in signatures:
             f.register(method, signature, precedence)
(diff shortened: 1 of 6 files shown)


plum/function.py
```diff
@@ -14,8 +14,13 @@
 _promised_convert = None
 """function or None: This will be set to :func:`.parametric.convert`."""
 
+# replace this by from typing import Self in python 3.11
+Self = TypeVar("Self")
```
wesselb, line 18:
```suggestion
# `typing.Self` is available for Python 3.11 and higher.
try:  # pragma: specific no cover 3.11
    from typing import Self
except ImportError:  # pragma: specific no cover 3.7 3.8 3.9 3.10
    Self = TypeVar("Self", bound="Function")
```

How about this?

plum/function.py
```diff
@@ -14,8 +14,13 @@
...
 
-def _convert(obj, target_type):
+SomeExceptionType = TypeVar("SomeExceptionType", bound=Exception)
+
+
+def _convert(obj: Any, target_type: type):
```
wesselb, line 23:
```suggestion
def _convert(obj: Any, target_type: type) -> Any:
```

plum/function.py
```diff
@@ -391,7 +404,7 @@ def __call__(self, *args, **kw_args):
 
         return _convert(method(*args, **kw_args), return_type)
 
-    def invoke(self, *types):
+    def invoke(self, *types: type):
```
wesselb, line 407:
```suggestion
    def invoke(self, *types: type) -> Callable:
```

plum/parametric.py
```diff
@@ -114,7 +116,7 @@ def type_parameter(cls):
             )
 
 
-def _default_le_type_par(p_left, p_right):
+def _default_le_type_par(p_left: Union[type, object], p_right: Union[type, object]):
```
wesselb, line 119:
```suggestion
def _default_le_type_par(p_left: Union[type, object], p_right: Union[type, object]) -> bool:
```

plum/resolver.py
```diff
@@ -95,7 +98,7 @@ def doc(self, exclude=None):
         # a newline.
         return "\n\n".join(unique_docs)
 
-    def register(self, signature):
+    def register(self, signature: Signature):
```
wesselb, line 101:
```suggestion
    def register(self, signature: Signature) -> None:
```

wesselb (commented):
This is incredibly helpful, @gabrieldemarmiesse! Thank you. :) I've added a few suggestions to add return types wherever the arguments have types.

Let me know when you think this is ready to merge.

gabrieldemarmiesse:
@wesselb you can re-review, I had the time to also change some stuff after running mypy on a bit of it.

type hints cannot be typed at "type", we have to use "object" it's the only public type hint available for it :(

plum/function.py
gabrieldemarmiesse: Yes, done

plum/resolver.py
gabrieldemarmiesse: done

plum/function.py
gabrieldemarmiesse: Done

plum/parametric.py
gabrieldemarmiesse: Done

plum/function.py
gabrieldemarmiesse: done

plum/dispatcher.py
```diff
@@ -15,10 +17,10 @@ class Dispatcher:
     """
 
     def __init__(self):
-        self.functions = {}
-        self.classes = {}
+        self.functions: Dict[str, Function] = {}
+        self.classes: Dict[str, Dict[str, Function]] = {}
 
-    def __call__(self, method=None, precedence=0):
+    def __call__(self, method: Optional[Callable] = None, precedence: int = 0):
```
wesselb, line 23:
```suggestion
    def __call__(self, method: Optional[Callable] = None, precedence: int = 0) -> Callable:
```

plum/function.py
```diff
@@ -80,22 +88,22 @@ class Function(metaclass=_FunctionMeta):
 
     _instances = []
 
-    def __init__(self, f, owner=None):
+    def __init__(self, f: Callable, owner: Optional[str] = None):
```
wesselb, line 91:
```suggestion
    def __init__(self, f: Callable, owner: Optional[str] = None) -> None:
```

plum/function.py
```diff
@@ -160,17 +168,19 @@ def __doc__(self):
         return doc if doc else None
 
     @__doc__.setter
-    def __doc__(self, value):
+    def __doc__(self, value: str):
```
wesselb, line 171:
```suggestion
    def __doc__(self, value: str) -> None:
```

wesselb:
@gabrieldemarmiesse ah, yes, I think you're right that the best we can do is use `object` for type hints. :( That's unfortunate. I think your solution with `TypeHint` is great!!

How would you feel about renaming `TypeHint` to `TypeHintType`, to avoid confusing with `beartype`'s `TypeHint`?

I've added a few more suggestions for return types. (`mypy` only checks functions with returns types, which is why I'm checking them all. :) )

gabrieldemarmiesse:
I'm unsure. TypeHint (the type) will be used a lot more in the codebase than beartype's TypeHint. And TypeHintType has Type in it twice, making it quite hard to understand at first glance and can hinder readability. I would suggest to take one of those three paths:
1) leave the PR untouched and merge it
2) Rename the import `from beartype.door import TypeHint as BeartypeTypeHint`
3) Find a new name for TypeHintType that is very easy to understand and read

wesselb (approved):
Alright, that's a fair point. I think the current solution is quite reasonable, so let's go with that then!

gabrieldemarmiesse:
Thanks, I'll make more PRs for type hints :)

wesselb:
Thank you, @gabrieldemarmiesse, that's much appreciated. :)

Issue #89, opened 2023-07-21
Is there a way to make pycharm understand what plum does?


gabrieldemarmiesse:
From the readme's example:

```python
from numbers import Number

from plum import dispatch


@dispatch
def f(x: str):
    return "This is a string!"


@dispatch
def f(x: int):
    return "This is an integer!"


@dispatch
def f(x: Number):
    return "This is a general number, but I don't know which type."
```

Pycharm doesn't understand the signature of the function f. It also can't jump to source according to the correct type. Is this possible to make pycharm work better with plum?

gabrieldemarmiesse:
No worries, I like plum a lot, but IDE integration prevents us from using it in our company, so yeah. But I can see the potential of this package. I think this was a missing feature of python

wesselb:
It's a shame, albeit completely reasonable of course, to hear that the lack of IDE integration is what's preventing adoption. I've been meaning to look into how difficult some reasonable degree of integration would be. Once I find some time to look into this, I'll update you in this issue!

machow:
I don't know how nuts this is in practice, but what about something like this?

```python
from typing import TYPE_CHECKING
from plum import dispatch

if TYPE_CHECKING:
    from typing import overload
else:
    def overload(f):
        return dispatch(f)


@overload
def f(x: int) -> int:
    return x

@overload
def f(x: str) -> str:
    return x
```

If plum defined an overload like above, I wonder if a user could do...

```python
from plum import overload

@overload
def f(x: int) -> int:
    return x

@overload
def f(x: str) -> str:
    return x
```

And then use it not just as a type stub, but also to implement the dispatch. I tested a bit in VS Code and was surprised it showed the overloaded signatures. One downside is that the signature for overload is the one from `typing.overload`.

machow:
@wch gave some really helpful feedback on this approach. In python 3.11, `overload` [sets the wrapped function in a registry](https://github.com/python/cpython/blob/3.11/Lib/typing.py#L2573-L2611). In this case, plum could use `typing.get_overloads()` to get all the original functions back...

```python
from typing import overloads
from plum import dispatch

@overload
def f(x: int) -> int:
    return x

@overload
def f(x: str) -> str:
    return x

# here, dispatch could use typing.get_overloads to retrieve the functions above
@dispatch
def f(x):
   raise NotImplementedError()
```

wesselb:
[...]

PR #93, opened 2023-08-13
Add `plum.overload`

wesselb:
@gabrieldemarmiesse, @machow, and @wch, I've taken a stab at adding `plum.overload` as described in #89. With this PR, the following would be functional and `mypy`-compliant:

```python
from plum.overload import dispatch, overload


@overload
def f(x: int) -> int:
    return x


@overload
def f(x: str) -> str:
    return x


@dispatch
def f(x):
    pass
```

How would something like this look?


wch:
I have two suggestions:


- make the `typing-extensions` dependency a hard dependency (not in `dev`), but conditional on Python version.
- make the `typing_extensions` import conditional on Python version.


For the first one, the pyproject.toml would have something like this:

```
dependencies = [
    ...,
    "typing-extensions; python_version<='3.10'"
]
```

(Note that I've done this kind of thing with setup.cfg, but I'm not 100% sure this is the correct syntax for pyproject.toml.)

If you do this, then the `try` wrapping the import won't be necessary.


Next, to make the `typing_extensions` import conditional on Python version, the code should look like this:

```py
import sys

if sys.version_info >= (3, 11):
    from typing import get_overloads, overload
else:
    from typing_extensions import get_overloads, overload
```

With this change, Python 3.11+ won't require `typing-extensions`. Also, as time goes on and Python 3.10 reaches end of life, then it will be clear in the code that you can you remove the `if sys.version_info >= (3, 11)` and always do `from typing import ...`.


***** 

One more thing that would be helpful for pyright users (like me and my team) is to add pyright tests in addition to mypy.

plum/overload.py
```diff
@@ -0,0 +1,21 @@
...
+    from typing_extensions import get_overloads, overload
+
+from .function import Function
+
+__all__ = ["overload", "dispatch"]
+
+T = TypeVar("T", bound=Callable)
+
+
+def dispatch(f: T) -> T:
```
gabrieldemarmiesse, line 16:
```suggestion
def dispatch(f: Callable) -> Function:
```
Maybe mypy didn't catch this? I'm unsure

plum/overload.pyi
gabrieldemarmiesse, line 1: I'm not sure I understand why this file is needed since the type hints are already in the implementation. If it's necessary, maybe let's add a docstring on top to explain why this file is needed?

gabrieldemarmiesse (changes requested):

plum/overload.pyi
wesselb: [`mypy` e.g. uses files like this](https://mypy.readthedocs.io/en/stable/stubs.html), called type stubs, to determine the types of implemenations. See also [PEP 561](https://peps.python.org/pep-0561/).

If you just `pip install plum-dispatch`, IIRC some linters wont recursive into the source to get the types of things. They will specifically look for type stubs.

Hence, even though the information in this file is also contained in the additional source, it is a necessary addition to make all tooling work correctly.

plum/overload.py
wesselb: Although this is technically correct, I believe this won't have the intended effect. Specifically, I tried changing the type stub to the following:

```python
from typing import Callable, TypeVar, overload

from plum import Function

__all__ = ["overload", "dispatch"]

T = TypeVar("T", bound=Callable)

def dispatch(f: Callable) -> Function: ...
```

Then `mypy` is happy with the following:


```python
from plum.overload import dispatch, overload


@overload
def f(x: int) -> int:
    return x


@overload
def f(x: str) -> str:
    return x


@dispatch
def f(x):
    pass


f(1)
f("1")
f(1.0)  # Should complain about this!!
```

Note that this is wrong, because it should complain about `f(1.0)`.

I believe that's what happening is the following. If you say that `dispatch` takes in `Callable` and gives back a `Function`, then `f(1.0)` will be matched against `Function.__call__`. `Function.__call__` takes in any arguments, so `f(1.0)` check out.

However, if you say that `dispatch` takes in `T` and gives back `T`, then `mypy` knows that it must correspond to the implementation in the file. The `@overload`s in the file correspond to the implementation of `f` in the file, so `mypy` knows that `f(1.0)` is wrong and then correctly complains.

plum/overload.py
```diff
@@ -0,0 +1,21 @@
...
+if sys.version_info >= (3, 11):  # pragma: specific no cover 3.7 3.8 3.9 3.10
+    from typing import get_overloads, overload
+else:  # pragma: specific no cover 3.11
+    from typing_extensions import get_overloads, overload
+
+from .function import Function
+
+__all__ = ["overload", "dispatch"]
+
+T = TypeVar("T", bound=Callable)
```
wch, line 13: With pyright (and by extension, pylance) in strict mode, `Callable` with no parameters will be flagged as an error.

image

I believe that for mypy, `Callable` by itself is shorthand for `Callable[..., Any]`, so it should be like this:

```suggestion
T = TypeVar("T", bound=Callable[..., Any])
```

plum/overload.py
wesselb: Hold on, I screwed up the test on my part! `mypy` actually does correctly complain about `f(1.0)` with the corrected type stub. I'm now a bit confused, because I don't exactly understand what's going on.

plum/overload.py
wesselb: Thanks for the suggestion, @wch! I've committed it. :)

plum/overload.pyi
wch: For type hints to be respected, a `py.typed` file is required -- when the package is installed normally. (Note that when working on the package code itself, or if the package is installed with `pip install -e`, then the `py.typed` file might not be necessary for the tools to respect type hints.)

You may need to add something to the `pyproject.toml`. For example, this is what we use in our `setup.cfg` file:

https://github.com/rstudio/py-shiny/blob/4bcd5296e36438928e793b068c4aa5b79dd4f134/setup.cfg#L108-L109

```
[options.package_data]
shiny = py.typed
```

According to [this](https://stackoverflow.com/a/76214408/412655), you can use something similar in pyproject.toml:

```
[tool.setuptools.package-data]
"pkgname" = ["py.typed"]
```

And you can test by building a wheel and installing that.

wesselb: Thanks @wch, @gabrieldemarmiesse! You're entirely right. I've removed the `.pyi` file and can confirm that things work.

@wch, after building and installing from the wheel locally, I can confirm that there are no additions necessary to `pyproject.toml`. Hatchling correctly bundles the file.

Thanks both for the input. :)

plum/overload.py
```diff
@@ -0,0 +1,21 @@
...
+
+T = TypeVar("T", bound=Callable[..., Any])
+
+
+def dispatch(f: Callable) -> Function:
+    """Decorator to register a particular signature."""
+    f_plum = Function(f)
+    for method in get_overloads(f):
+        f_plum.dispatch(method)
+    return f_plum
```
gabrieldemarmiesse, line 21: I think we can do an even better UX. What if, instead of using `from plum.overload import overload, dispatch` we could do `from plum import overload, dispatch`? This would mean that the `dispatch` function automatically detech overloads, which is possible in theory.

Can this logic be put in there: https://github.com/beartype/plum/blob/master/plum/dispatcher.py#L35 ?

The function signature is nearly the same, so it should work with mypy.

Something like:

```python

    def __call__(
        self,
        method: Optional[Callable] = None,
        precedence: int = 0,
    ) -> Callable:
        """Decorator to register for a particular signature.

        Args:
            precedence (int, optional): Precedence of the signature. Defaults to `0`.

        Returns:
            function: Decorator.
        """
        if method is None:
            return lambda m: self(m, precedence=precedence)

        overloads = get_overloads(method)
        if overloads != []:
            for overload_function in overloads:
                # the plum_function is always the same in this loop
                plum_function = self(overload_function)
            # method is never registered because it has no implementation
            return plum_function

        # The signature will be automatically derived from `method`, so we can safely
        # set the signature argument to `None`.
        return self._add_method(method, None, precedence=precedence)
```

gabrieldemarmiesse (changes requested):

plum/overload.py
wesselb: I've now added tests for both `mypy` and `pyright`. It seems that `pyright` does indeed require the `T -> T` construction. If you'd like to experiment with it, you can run the linter assertion tests with

```bash
python check_linter_assertions.py tests/typechecked
```

from the repo root.

plum/overload.py
```diff
@@ -0,0 +1,21 @@
...
+if sys.version_info >= (3, 11):  # pragma: specific no cover 3.7 3.8 3.9 3.10
+    from typing import get_overloads, overload
+else:  # pragma: specific no cover 3.11
+    from typing_extensions import get_overloads, overload
+
+from .function import Function
+
+__all__ = ["overload", "dispatch"]
+
+T = TypeVar("T", bound=Callable[..., Any])
```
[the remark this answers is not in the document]
wesselb: I've added this back in, because it seems that `pyright` does require the `T -> T` construction. :(

docs/integration.md
```diff
@@ -0,0 +1,33 @@
...
+    return x
+
+
+@dispatch
+def f(x):
+    pass
+```
+
+In the above, `plum.overload.overload` is `typing_extensions.overload`.
+For this pattern to work in all Python versions, you must use `typing_extensions.overload`, not `typing.overload`.
```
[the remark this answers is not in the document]
wesselb: You're right! First line is corrected now.

With the second line I intended to explain that things might break if you try to `from typing import overload`. I've added a brief second sentence to elaborate.

plum/overload.py
wesselb: @gabrieldemarmiesse Ahhh this is clever. I've implemented it, and it works! Thank you for the awesome suggestion. :)

wesselb:
@wch, I've integrated `pyright` into the unit test. Whereas the code is not fully `pyright`-compliant, `pyright` does pick up the dispatch pattern and complains whenever `f` is given an argument of the wrong type, so that's at least something! This is asserted by the comments in `tests/typechecked/test_overload.py` in combination with `check_linter_assertions.py`. It's a little hacked together, but it does the job.

docs/api.rst
```diff
@@ -43,6 +43,10 @@ iPython AutoReload
 .. automodule:: plum.autoreload
     :members:
 
+Overload Support
+----------------
+.. autofunction:: plum.overload.dispatch
```
gabrieldemarmiesse, line 48: are you sure this function still exists?

gabrieldemarmiesse (changes requested):

docs/api.rst
wesselb: Should be gone now. :) Ty!

wesselb:
Unless there are any objections, I think that I'll be merging this. :)

Let's address further improvements in separate PRs!

Thanks @gabrieldemarmiesse and @wch for your suggested improvements and feedback!!

gabrieldemarmiesse:
LGTM!

PhilipVinc:
I've played a bit around with this, and judging from my personal use case, where dispatch is used to multiple expand a base library, I can say that it's not easy to ensure that the function defined with `@dispatch` be the last one defined. 
I think this interface is a bit... not easy to use.

Also, we had added a while back `@abstract` to be used as a first 'abstract' definition. Ideally one would define the `@abstract` function once and then define all implementations with `@overload` and plum would automatically register those at call time (though it might not be easy to detect when one has to re-register).

Can't we make `dispatch(method, ...)` automatically call `_ = typing.overload(method)` inside once so that (hopefully) type-checkers are happy?

wesselb:
@PhilipVinc That's an interesting idea!

I've hacked around a bit to see if I could come up with a `mypy`-compliant way where a function is defined in one file and extended in another, but I've had no success. Some problems are that (1) type checkers assume that all overloads _must_ come before the implementation, (2) you cannot import a function and then add more overloads, even if the imported function only has overloads defined, and (3) overloaded functions must have an implementation in the same file.

I'm almost thinking that some kind of rewriting of the source before running `mypy` or a linter would be the simplest way to make this work...

PR #94, opened 2023-08-16
:up: Set minimum Python version for plum to Python 3.8

gabrieldemarmiesse:
Python 3.7 end of support for security bugs was the 5th of june 2023.

Some notes here:

* This allows the python version in pre-commit to be the same as the one installed for testing of plum (3.8) because before, the pre-commit python version was 3.8 and the minimum python version for plum was 3.7.
* Since `typing.Literal` is available in python 3.8, I think `plum.Val` can be now deprecated. I made the related changes.
* `get_args` and `get_origin` can now be imported normally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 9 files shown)


wesselb:
This is great! Removes a whole bunch of annoying boilerplate. Thanks!! :)

Issue #92, opened 2023-08-10
Would you be open to a bit of refactoring?


gabrieldemarmiesse:
To make it easier to add support for keyword arguments #40 , I would like to do some light refactoring. For instance, hide the cache and `Function._pending` behind getters, so that it's impossible to forget to update it. It would also go in the Resolver class since it's not really the job of Function to cache the result of the resolver.

That would also potentially require some change in some public methods of those classes defined in this part of the doc: https://beartype.github.io/plum/api.html I wonder if we can change the api of those function without requiring a deprecation procedure (warning, backward compatibility, etc). Of course methods and functions described in "basic usage" and "advanced usage" should be considered intouchable.

If you're ok with some light refactoring I'll do several small PRs to make it easier to discuss the changes

wesselb:
Hey @gabrieldemarmiesse! As long as the API changes are mostly changes to internals, not API changes to the more commonly used functions, then I'd definitely be open to refactoring. :) With regards to moving things around, e.g. moving `_pending` to `Resolver`, I'd be open to that too, as long as the changes don't significantly increase coupling: the resolver current doesn't hold a reference to the function, and it would be nice to keep it that way.

gabrieldemarmiesse:
Sure thing, that would be possible

gabrieldemarmiesse:
Hi @wesselb, I wanted to give you an update on my work. I made the refactoring I had in mind and it helped me a lot to focus on the issue of keyword arguments. Now that it's done, it's actually hard to make the keyword argument system work, because of how flexible and fast plum is (caching is important here). 

I kind of gave up at this point (sorry I was young and full of dreams).

I tried to focus more on the `@overload` use case and forget about performance, and I tried to make a lib that does the dispatch only on implementations supported by `@overload`. Suprisingly, I had quite some success by matching the behavior of mypy and pyright.

So I kinda made a competitor of plum (but not really since my lib does not support crazy stuff) in the process, sorry again :( I hope that both projects can learn from each other. 

If it goes well, my goal is to make a PEP so that `@overload` can support multiple dispatch via third party libraries. Currently, we're in a grey area, hopefully we can do something like https://peps.python.org/pep-0681/ where we declare a "multiple dispatch" library and the type checkers will behave accordingly.

wesselb:
Hey @gabrieldemarmiesse! Sorry for the delay in replying. I've just got back to work. :)

[...]

Issue #40, opened 2022-05-19
Keyword args not supported


DonkeyKong1:
First, I really like this. I was getting pretty confused when I kept getting "plum.function.NotFoundLookupError" and couldn't find issue with the simple class I'd made. Randomly, I forgot to provide a keyword for the arg and it worked. Am I correct that keyword args are not supported?

wesselb:
Hey @DonkeyKong1! You're right that dispatch interacts a bit strangely with the keyword arguments. In particular, arguments which are not given a default value must also be given in non-keyword form. E.g., consider

```python
@dispatch
def f(x: int):
     return x
```

Then `f(2)` will work, whereas `f(x=2)` breaks.

You can use default values for arguments, and, in that case, you can give these arguments in keyword form. See [https://github.com/wesselb/plum#keyword-arguments-and-default-values](here).

I hope this clarifies! :) Please let me know if you have any other questions.

astanziola:
Hi! Sorry to post on an old issue, but is there any fundamental reason for not supporting keyword arguments without default values?
I'm using plum to dispatch over operators with many type combinations of mandatory inputs, and it would make the code much clearer to be able to call them using keyword arguments.

wesselb:
Hey @astanziola!

The design of Plum is heavily inspired by how multiple dispatch works in the Julia programming language. In Julia, keyword arguments are also not considered in the dispatch process. I believe you can find some discussion online around this point.

More fundamentally, multiple dispatch is based on the idea that arguments are identified by _position_ and _type_. Keyword arguments, on the other hand, are identified by _name_. I agree that one could consider a dispatch process which also considers keyword arguments, but, up to this point, we've tried to follow the Julia programming language.

astanziola:
Thank you for clarifying @wesselb , I had no idea that Julia was following this rule for the multiple dispatch system!

teffland:
Hi, are there any plans in the future to support this? I really like dispatch, but this is a bit of an issue for downstream usability, since in python its very common to call functions with named args for both readability and to stay robust to irrelevant upstream signature updates.

Is there something inherently preventing this behavior? Would be interested in making a contribution to allow this behavior or at least display a more useful error message to the user.

wesselb:
@teffland Currently, there are no plans to dispatch on keyword arguments, but obviously future plans can change. :) I agree that calling a function with named arguments is very common in Python. It would be very nice to support this.

> Is there something inherently preventing this behavior? 

[...]

PR #96, opened 2023-08-17
:wrench: Group logic concerning the cache in one single method

gabrieldemarmiesse:
* The cache is read and updated every time the method `Function.resolve_method` is called. This was done in various places in the code, making it hard to understand and easy to get wrong. Now one single method is in charge of reading and updating the cache: `_resolve_method_with_cache`.
* This function also ensures there are no pending registration, so it's not necessary to manually call `self._resolve_pending_registrations()` every time before resolving the method.
* The `resolved_method` has been made shorter by extracting the logic concerning the `__mro__` in another method. This allows the method `resolved_method` to have less local variables overall and less variables that are only used in specific situations.

Later on, it will be possible to move the cache into the resolver, but we need to start somewhere, the aim of this PR is just to make the code clearer, not to have the perfect structure

Related to #92  the goal overall is to have a good structure to add support for keyword arguments, as mentionned in #40

### DIFF
--- a/plum/function.py
+++ b/plum/function.py
@@ -316,14 +316,44 @@ class Function(metaclass=_FunctionMeta):
         message = str(e)
         return type(e)(prefix + message[0].lower() + message[1:])
 
+    def _resolve_method_with_cache(
+        self,
+        args: Union[Tuple[object, ...], Signature, None] = None,
+        types: Optional[Tuple[TypeHint, ...]] = None,
+    ) -> Tuple[Callable, TypeHint]:
+        if args is None and types is None:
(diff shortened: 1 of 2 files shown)


PhilipVinc (approved):
@wesselb This looks like a nice refactoring, and I support it.

(Wessel has been trying to keep perfect coverage, so maybe you should fix that one line that is no longer covered by tests?)

wesselb:
@gabrieldemarmiesse Thanks for the refactoring. I think this is a lot more clear and a definite improvement. Merging right away :)

PR #97, opened 2023-08-19
Simplify updates of resolver

gabrieldemarmiesse:


wesselb:
Hey @gabrieldemarmiesse! Thanks for submitting yet another PR. :)

Whereas I think your proposed changes are eminently sensible from a design point-of-view, I’m slightly hesitant to change the point at which methods are resolved. 

To properly deal with forward references, [promised types](https://beartype.github.io/plum/types.html#promisedtype), and other sorts of type constructions that one preferable should not be doing, it is important that the methods are always resolved at the very last moment, i.e. when the function is called for the first time. I’m worried that calling the resolution process in getters might inadvertently change the order of operations in more complex designs. Therefore, for the particular operation of registering the methods, I think I might prefer the design without getters where the registration process is called explicitly.

I’m aware that the unit tests still pass with this change, so there shouldn’t be a problem there. I’m more talking about other packages I’m working on that very intensely use Plum, where I’ve encountered a fair share of dispatch-related problems. It has taken quite some effort to get the order of operations just right.

In addition, I think being able to see how many methods are registered and pending is quite a useful feature, especially in dispatch-related problems.

gabrieldemarmiesse:
Thanks for the feedback, I'll close this pull request then

wesselb:
Thanks for your understanding! I do really appreciate your efforts in trying to improve the internal design. :)

PR #100, opened 2023-09-01
remove deprecation warning of Val

PhilipVinc:


wesselb:
Thanks, @PhilipVinc! Change makes total sense. :)

PR #102, opened 2023-09-04
Refactoring: Clarifying the responsabilities of each component

gabrieldemarmiesse:
* The cache is maintained by the resolver now
* The resolver is immutable, making it simpler to reason about. Only the cache can change inside the resolver.
* A method registry is created. Functions can be added to it, but can never be removed, making it simpler to reason about.

Thus we simplify the whole logic of list swapping, cache invalidation and such. The resolver is fully in charge of doing whatever it sees fit with the inputs, without having other concepts getting in the way.

If one is brave enough, by changing only the Signature class + the resolver, it should be enough to support keyword arguments.


plum/function.py
```diff
@@ -125,21 +259,6 @@ def __doc__(self) -> Optional[str]:
...
-            # `staticmethod` will attempt to inherit `__doc__` (see
-            # https://docs.python.org/3/library/functions.html#staticmethod). Since
-            # we are still in class construction, forward references are not yet
-            # defined, so attempting to resolve all pending methods might fail with a
-            # `NameError`. This is fine, because later calling `__doc__` on the
-            # `staticmethod` will again call this `__doc__`, at which point all methods
-            # will resolve properly. For now, we just ignore the error and undo the
-            # partially completed :meth:`Function._resolve_pending_registrations` by
-            # clearing the cache.
-            self.clear_cache(reregister=False)
```
wesselb, line 142: Could you explain why you've removed this? Do you think it's not at all necessary to resolve pending registration when getting the owner? I should've documentation why that line was there in the first place... I think removing it might make sense, but I'm worried about subtly breaking something that's currently not tested for, so I just want to double check with you.

tests/test_function.py
```diff
@@ -284,9 +284,8 @@ def f(x: int):
     g = Function(f)
     g.register(f)
 
-    assert g._pending == [(f, None, 0)]
-    assert g._resolved == []
-    assert len(g._resolver) == 0
+    assert len(g.methods) == 1
+    assert g.methods[0].implementation == f
```
wesselb, line 288:
```suggestion
    assert len(g.methods) == 1
    assert g.methods[0].implementation == f
    assert g.methods[0].precedence == 0
```

tests/test_function.py
```diff
@@ -301,19 +300,16 @@ def f(x: int):
...
     def f(x: str):
         pass
 
     # Now there is something to register. A call should clear the cache.
-    assert len(f._pending) == 1
-    f._resolve_pending_registrations()
-    assert len(f._pending) == 0
-    assert len(f._cache) == 0
+    f._resolver
+    assert f._methods_registry._resolver._cache == {}
```
wesselb, line 312:
```suggestion
    assert not f._methods_registry._resolver._cache
```

(slightly more Pythonic)

plum/resolver.py
```diff
@@ -66,62 +23,62 @@ class Resolver:
         is_faithful (bool): Whether all signatures are faithful or not.
     """
 
-    def __init__(self):
-        self.signatures: List[Signature] = []
-        self.is_faithful: bool = True
+    def __init__(self, signatures: Iterable[Signature]):
+        signatures_dict = {hash(s): s for s in signatures}
```
wesselb, line 27: Ahhh, I now understand that this is used to remove duplicate signatures. What I'm worried about is specifically which duplicate is kept. Specifically, if `signatures` implements the same method twice, then the _later_ one should be kept. Right now, which implementation is kept is up to the implementation of the dictionary. I think it's safest to implement this in way that very explicitly retains the later implementations.

plum/resolver.py
```diff
@@ -66,62 +23,62 @@ class Resolver:
...
-        # Use a double negation for slightly better performance.
-        self.is_faithful = not any(not s.is_faithful for s in self.signatures)
-
-    def __len__(self) -> int:
-        return len(self.signatures)
+        signature = self._resolve(target)
+        return signature.implementation, signature.return_type
 
-    def resolve(self, target: Union[Tuple[object, ...], Signature]) -> Signature:
+    def _resolve(self, target: Union[Tuple[object, ...], Signature]) -> Signature:
```
wesselb, line 81: Why make this a private method?

tests/typechecked/test_overload.py
```diff
@@ -6,24 +6,24 @@
...
 def test_overload() -> None:
-    assert f(1) == 1
-    assert f("1") == "1"
+    assert f(1, 2) == 1
+    assert f("1", "2") == "1"
     with pytest.raises(NotFoundLookupError):
         # E: pyright(argument of type "float" cannot be assigned to parameter "x")
         # E: mypy(no overload variant of "f" matches argument type "float")
-        f(1.0)
+        f(1.0, 2.0)
```
wesselb, line 29:
```suggestion
        f(1.0, 2.0)
        
  def test_methods() -> None:
      assert f.methods
```

How about adding in a unit test like this? That would test whether the typing for `Dispatcher` is right.

wesselb (commented):
Hey @gabrieldemarmiesse! Thanks for this PR!! Your changes make a lot of sense. I've left a few comments and suggestions. :)

plum/resolver.py
wesselb: We should also have this tested by a unit test!

PR #104, opened 2023-09-18
Set minimum bear type version to 0.16

PhilipVinc:
It fixes a few bugs with `Literal`and `Union`, and so I would like to argue that they are 'required' for plum to work properly


wesselb:
Thanks @PhilipVinc! Makes total sense. :) Merging this right away, but perhaps I'll wait with a new release until we have a better understanding of the one test that is failing.

PR #106, opened 2023-09-18
Fix #105

### LINKED ISSUE #105: `parametric` conflicts with certain usages of customized `__init_subclass__`
Hi!

I'm using JAX, and also using `plum` -- in my library, I've define a mixin class called `Pytree` which automatically implements the Pytree interface for classes which mix it in.

It's quite simple:

```python
class Pytree:
    def __init_subclass__(cls, **kwargs):
        jtu.register_pytree_node(
            cls,
            cls.flatten,
            cls.unflatten,
        )

```

If I wish to use this mixin, and `parametric` -- I'm in for problems, I get duplicate registration:

```
ERROR ... ValueError: Duplicate custom PyTreeDef type registration for <class...>
```

I'm not exactly sure why this occurs, but I'm hoping to find a fix -- because I'd like to use parametric classes to guide some of the dispatch in my library functions.

PhilipVinc:
We were providing the wrong class to the `__init_subclass__`called in our initialisation of parametric subclasses. This fixes it and adds a test.

### DIFF
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -240,7 +240,7 @@ def parametric(original_class=None):
                 return original_class.__new__(cls)
 
             cls.__new__ = class_new
-        original_class.__init_subclass__(**kw_args)
+        super(original_class, cls).__init_subclass__(**kw_args)
 
     # Create parametric class.
     parametric_class = meta(
--- a/tests/test_parametric.py
+++ b/tests/test_parametric.py
@@ -507,3 +507,25 @@ def test_val():
         Val[1].__init__(MockVal())
 
     assert repr(Val[1]()) == "plum.parametric.Val[1]()"
+
+
+def test_init_subclass_correct_args():
+    # See issue https://github.com/beartype/plum/issues/105
+
+    from plum import parametric
+
+    register = set()
+
+    class Pytree:
+        def __init_subclass__(cls, **kwargs):
+            if cls in register:
+                raise ValueError("duplicate")
+            else:
+                register.add(cls)
+
+    @parametric
+    class Wrapper(Pytree):
+        pass
+
+    Wrapper[int]
+    assert Wrapper[int] in register


PhilipVinc:
@femtomc CI is broken by the latest release of bear type, but this should work. Can you test it on your code and let me know if it solves it even in the more realistic/complex scenario?

femtomc:
Will do, one moment.

femtomc:
@PhilipVinc That fix works to get past the first error! -- but I'm actually finding now that some of my methods (defined via `dispatch` inside of classes) are not correctly dispatching based on _covariance_.

E.g. in a base class (call it `Model`) which I mix in, I define a method with signature:

```
    @dispatch
    def update(
        self,
        key: PRNGKey,
        prev: Trace,
        new_constraints: Mask[ChoiceMap],
        argdiffs: Tuple,
    ) -> Tuple[Any, FloatArray, Trace, Mask[ChoiceMap]]:
```

Now, in a subclass which mixes in `Model`, I tried and invoke `update` with a `Mask[T1]`, where `T1 <: ChoiceMap`.

However, this fails on the resolver:

```
plum.resolver.NotFoundLookupError: For function `update` of `Model`, `(Normal(), Array([4171552086, 1934896579], dtype=uint32), DistributionTrace(gen_fn=Normal(), args=(0.0, 1.0), value=Array(-0.10823099, dtype=float32), score=Array(-0.9247955, dtype=float32)), <Mask[ValueChoiceMap] object at 0x7fac1c6527d0>, (Diff(primal=0.0, tangent=_NoChange()), Diff(primal=1.0, tangent=_NoChange())))` could not be resolved.
```

I can pull this into a separate issue.

PhilipVinc:
If you can provide a MWE that does not work I can give a look. 
I suspect some bugs might be in the underlying bear type, which might not correctly treat type parameters correctly.

wesselb:
There is one failing test, but that failure seems unrelated to this, so I'm happy to merge this if you are too. :)

@philipvinc and @femtoc, I believe this might be the MWE that you are looking for:

```python
from numbers import Number
from plum import parametric


@parametric
class A:
    def __init__(self, x):
        self.x = x
```

Then

```python
>>> issubclass(A[int], A[int])  # Just a check
True

>>> issubclass(A[int], A[Number])  # Covariance!
True

>>> isinstance(A[int](None), A[Number])  # Here is where it goes wrong :((( This breaks dispatch.
False
```

I've been vaguely aware of this for a while, but never really took the time to dig into it. The problem is that we have not implemented `CovariantMeta.__instancecheck__`. I'm not 100% sure what exactly the right way to implement `CovariantMeta.__instancecheck__` is. My sense is that the the default `__instancecheck__` works fine, _except_ when the other argument is another instance of the parametric class. Hence, once option would be to handle this special case.

Here's how that would work in the above example:


```python
from numbers import Number
from plum import parametric


class AMeta(type):
    def __instancecheck__(self, x):
        return issubclass(type(x), self)

@parametric
class A(metaclass=AMeta):
    def __init__(self, x):
        self.x = x
```

Then
```python
>>> issubclass(A[int], A[int])
True

>>> issubclass(A[int], A[Number])
True

>>> isinstance(A[int](None), A[Number]) 
True
```

so we're all good.

I have two concerns about this:

1. Is `isinstance(type(x), self)` truly correct here? Can this give `False` in situations where it should give `True`? (This is related to our made-up notion of faithfulness.) Is there a way to implement this without appealing to `type(x)`?

2. Are there any performance implications for reducing instance checks to subclass checks? `parametric` is already pretty slow, so maybe this is not a problem.

PhilipVinc:
yes the failure is unrelated and due to the resolve pep _bug_ in beartype. I'll merge this fix and we can experiment with fixing the other issue raised in a separate PR.

PR #110, opened 2023-09-28
[3/3] Improve Signature and method table printing

PhilipVinc:
very very WIP. 
I would also like to experiment with [rich](https://rich.readthedocs.io/en/stable/) to print more coloured, more readable stuff, but this is already more readable...

```python
In [1]: f.methods
Out[1]: 
Method List with # 7 methods:
 [0] f(x: str)
        <function f at 0x103a63060> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:6
 [1] f(x: int)
        <function f at 0x103c811c0> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:36
 [2] f(x: numbers.Number)
        <function f at 0x103c80ea0> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:16
 [3] f(x: numbers.Number, k: float, *, kk, j, ksa, **povolo)
        <function f at 0x103c80fe0> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:28
 [4] f(x: float, y: numbers.Number) -> int
        precedence=10
        <function f at 0x103c81080> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:24
 [5] f(x: numbers.Number, y: float, *z: tuple[int, ...])
        <function f at 0x103c81120> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:32
 [6] f(x: int, y: int, *z: tuple[int, ...])
        <function f at 0x103c811c0> @ /Users/filippo.vicentini/Dropbox/Ricerca/Codes/Python/plum/prova.py:36
```

It refactors a bit the internals. Need to find a better way to do it..

### DIFF
--- a/plum/method.py
+++ b/plum/method.py
@@ -1,6 +1,6 @@
 import inspect
 import typing
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional, Set, Tuple
 
 from rich.segment import Segment
 from rich.text import Text
@@ -90,70 +90,54 @@ class Method:
         return f"Method({function_name=}, {signature=}, {return_type=}, {impl=})"
 
     def __rich_console__(self, console, options):
-        arg_names, kw_names, kw_var_name = extract_arg_names(self.implementation)
+        yield self.repr_mismatch()
 
-        sig = self.signature
-        parts = []
-        if sig.types:
-            for nm, t in zip(arg_names, sig.types):
-                parts.append(Text(f"{nm}: ") + repr_type(t))
-        if sig.varargs != Signature._default_varargs:
-            parts.append(Text(f"*{arg_names[-1]}: ") + repr_type(sig.varargs))
-
-        if len(kw_names) > 0 or kw_var_name is not None:
-            parts.append(Text("*"))
-        for kwnm in kw_names:
-            parts.append(Text(f"{kwnm}"))
-        if kw_var_name is not None:
-            parts.append(Text(f"**{kw_var_name}"))
-
-        res = Text(self.function_name) + Text("(") + Text(", ").join(parts) + Text(")")
-        if self.return_type != Method._default_return_type:
-            res.append(" -> ")
-            res.append(repr_type(self.return_type))
-        if sig.precedence != Signature._default_precedence:
-            res.append(f"\n    precedence={sig.precedence}")
-
-        yield res
-        yield Segment("    ")
-        yield repr_pyfunction(self.implementation)
-
-    def _repr_signature_mismatch(self, args_ok: List[bool]) -> str:
-        """Special version of __repr__ but used when
-        printing args mismatch (mainly in hints to possible
-        similar signatures).
+    def repr_mismatch(
+        self,
+        mismatches: Set[int] = frozenset(),
+        varargs_matched: bool = True,
+    ) -> str:
+        """Version of `__repr__` that can print which arguments are mismatched. This
+        is mainly used in hints.
 
         Args:
-            args_ok: a list of which arguments match this signature
-                and which don't according to the resolver.
+            mismatches (set[int], optional): Indices of the positional arguments which
+                are mismatched. Defaults to no mismatched arguments.
+            varargs_matched (bool, optional): Whether the varargs are matched. Defaults
+                to `True`.
+
+        Returns:
+            list: :mod:`rich` representation of the method showing which arguments
+                are mismatched.
         """
         sig = self.signature
-
         arg_names, kw_names, kw_var_name = extract_arg_names(self.implementation)
-        varargs_ok = all(args_ok[len(sig.types) :])
 
         parts = []
+
+        # Walk through the positional arguments.
         if sig.types:
-            for i, (nm, t) in enumerate(zip(arg_names, sig.types)):
-                is_ok = args_ok[i] if i < len(args_ok) else False
-                arg_txt = Text(f"{nm}: ")
+            for i, (arg_name, t) in enumerate(zip(arg_names, sig.types)):
+                arg_txt = Text(f"{arg_name}: ")
                 type_txt = repr_type(t)
-                if not is_ok:
+                if i in mismatches:
                     type_txt.stylize("red")
                 arg_txt.append(type_txt)
                 parts.append(arg_txt)
+
+        # Print the varargs.
         if sig.varargs != Signature._default_varargs:
             arg_txt = Text(f"*{arg_names[-1]}: ")
             type_txt = repr_type(sig.varargs)
-            if not varargs_ok:
+            if not varargs_matched:
                 type_txt.stylize("red")
             arg_txt.append(type_txt)
             parts.append(arg_txt)
 
         if len(kw_names) > 0 or kw_var_name is not None:
             parts.append(Text("*"))
-        for kwnm in kw_names:
-            parts.append(Text(f"{kwnm}"))
+        for kw_name in kw_names:
+            parts.append(Text(f"{kw_name}"))
         if kw_var_name is not None:
             parts.append(Text(f"**{kw_var_name}"))
 
@@ -166,6 +150,7 @@ class Method:
 
         res.append("\n    ")
         res.append_text(repr_pyfunction(self.implementation))
+
         return res
 
 
--- a/plum/method.py
+++ b/plum/method.py
@@ -1,6 +1,6 @@
 import inspect
 import typing
-from typing import Callable, Optional
+from typing import Callable, List, Optional
 
 from .repr import color, colored, link, repr_type
 from .signature import Signature, inspect_signature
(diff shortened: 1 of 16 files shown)


wesselb:
Ah, that looks beautiful! Love the idea. :)

Let me know when you'd like me to review the PR or parts of it.

PhilipVinc:
I would love some insight on how to compute the distance between (invalid) signatures!

wesselb:
Ah, let me reply here instead of in the other issue. :)

Hmmm, I'm not exactly sure how a solution should look like, but maybe something simple like this could work:

```python
def signature_distance(candidate: Signature, reference: Signature):
    distance = 0
    for t1, t2 in zip(candidate.types, reference.types):
        if not issubclass(t1, t2):
            distance += 1
    return distance
```

This isn't yet quite right, since it doesn't appropriately consider the number of arguments (`len(candidate.types)`) and varags.

Do you have any thoughts or ideas? The above is just the first thing that comes to mind. We could likely do something better.

One approach would be to proceed with something simple like this and later improve the distance function.

PhilipVinc:
https://github.com/beartype/plum/assets/2407108/526c0878-a430-4614-8da2-ac571b933126

Ok, more updates... What do you think about this?

I think the main thing I would like to do is to replace the values printed in the resolution error with the types, at least for those that are not parametric... but that's not easily done...

Because it's hard to understand if the values printed there match or are related to the types below. maybe not hard, but not straightforward

wesselb:
> Ok, more updates... What do you think about this?

I absolutely love it!! That looks _so_ good. This is awesome work! :)

> I think the main thing I would like to do is to replace the values printed in the resolution error with the types, 

This would be really nice, but seems like it would require a `type_of` function. I agree with you that it would make it easier to understand the list of closest methods.

PhilipVinc:
@wesselb Ok, I cleaned up the PR. 

This is now based on top of #112 . 

It should be possible to implement this logic without the changes of #112 , so if you don't like those changes I could try to remove them from the PR, but in my opinion they modularise a bit the internals of plum (and It made it simpler for me to understand what I was working with.)

The major missing things from this PR is now: 
 - ANSI detection: Printing colors/bolds/hyperlinks in terminal is all nice and great because most terminals support them. Theroretically one should check that the terminal supports ANSI codes (which is what is used to display colours in terminals) but we could relatively safely assume they are supported. However...
    - When one uses output redirection, for example when running a batch job on a cluster, you don't want to print them because they show up as noise in the logs.  Detecting this reliably is... a mess. See for example [how complicated it is to detect if one is running on Jupyter (which is not a terminal)](https://github.com/Textualize/rich/blob/e9f75c9912ed25b9777bc0257853370951220b17/rich/console.py#L518). 
    - To do this... the most maintainable approach would be to add a dependency, rich, which is a no-dependency library for pretty-printing stuff, and use it. The library is solid (pip uses it as well since a year or so). Then, rich automatically detects and disables colors when they should not be 
    - Alternatively we could roll a simplified version of this, which would surely support fewer edge cases and be more prone to breaking with the benefit of not having the dependency.

I would frankly go for rich, but what do you think?

PhilipVinc:
@wesselb I don't want to push, but in case you like/accept the approach and the few internal refactors, I can work my way to add tests and increase coverage again. (I don't want to do the work if you don't like the structure as it would be wasted effort).

wesselb:
@PhilipVinc sorry for the delay! Yes, I absolutely love this!! It would be awesome to merge this. :) I think the changes in #112 are sensible too. I'll review those first, and then we can merge this.

With regards to `rich`, since it's a no-dependency library which is incredibly popular and well battle tested, I'd be happy to add it as a dependency. Another option would be to use `rich` if it is installed and otherwise print without colour. I see you've got a fairly concise custom implementation already in `repr.py`, so that would work too.

I'm slightly leaning towards just adding `rich` as a dependency for the reasons that you mentioned. What do you think?

PhilipVinc:
@wesselb The PR is done and the tests have been updated, so I think it is time for a review. Sorry for taking all your time!

As you agreed to include `rich`, that's what I used, because it makes output over dumb terminals/mpi/output pipes clean and is a bit easier to use than alternatives. 

Most functionality necessary for converting to displayable objects are in `plum/repr.py` with a lot of docstring. An important function there is `rich_repr`which is a decorator for classes to make them use `rich` repr handling.

Other minor changes are...
 - Resolution error printing is now delegated to the error types themselves instead of the Resolver. This simplifies the logic handling and removes the need for that `_enhance_error` function in `Function`.
 - Added a `MethodList` which is a thin wrapper over `list` that pretty prints methods.
 - Added some logic to detect which arguments are a mismatch and which are not, in order to colorise in red only those that mismatch.

Overall... this logic is ugly as hell, but I don't think it's possible to simplify those repr methods much..

wesselb:
Hey @PhilipVinc! This PR has been waiting for a long time now. I'm very sorry about that. :( I sometimes struggle to find the time for this.

I will prioritise this PR next week!

PhilipVinc:
bumpety bump

plum/method.py
```diff
@@ -77,13 +82,131 @@ def __hash__(self):
...
+    Returns:
+        :class:`.Signature`: Signature.
+    """
+    # Extract specification.
+    sig = inspect_signature(f)
+
+    # Get types of arguments.
+    argnames = []
+    kwnames = []
+    kwvar_name = None
```
wesselb, line 197:
```suggestion
    regular_args = []
    kw_only_args = []
    var_kw_name = None
```

where `var_kw_name` stands for `variable_keykword_name`.

Would u be ok with these variable names? Trying to make things as clear as possible.

plum/method.py
```diff
@@ -77,13 +82,131 @@ def __hash__(self):
...
+        if sig.precedence != Signature._default_precedence:
+            res.append(f"\n    precedence={sig.precedence}")
+
+        res.append("\n    ")
+        res.append_text(repr_pyfunction(self.implementation))
+        return res
+
+
+@rich_repr
+class MethodList(list):
```
wesselb, line 173:
```suggestion
class MethodList(list):
    "A list of :class:`Method`s which is nicely printed by :mod:`rich`.
```

plum/method.py
```diff
@@ -77,13 +82,131 @@ def __hash__(self):
...
 
-    def __repr__(self):
+    def __str__(self):
         function_name = self.function_name
         signature = self.signature
         return_type = self.return_type
         impl = self.implementation
         return f"Method({function_name=}, {signature=}, {return_type=}, {impl=})"
 
+    def __rich_console__(self, console, options):
```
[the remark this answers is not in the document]
PhilipVinc: IIRC `POSITIONAL_ONLY` arguments are handled as 'regular' arguments by `extract_arg_names`. There's no need to special case them.

We can special case them for better readability, but can we move this to a later PR? I can open an issue to remind us, but would like to get this merged.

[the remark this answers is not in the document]
wesselb: @PhilipVinc sounds entirely reasonable!

plum/method.py
```diff
@@ -77,13 +82,131 @@ def __hash__(self):
...
+            res.append(" -> ")
+            res.append(repr_type(self.return_type))
+        if sig.precedence != Signature._default_precedence:
+            res.append(f"\n    precedence={sig.precedence}")
+
+        yield res
+        yield Segment("    ")
+        yield repr_pyfunction(self.implementation)
+
+    def _repr_signature_mismatch(self, args_ok: List[bool]) -> str:
```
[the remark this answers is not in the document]
wesselb: This is now done.

plum/resolver.py
```diff
@@ -20,30 +20,31 @@ class AmbiguousLookupError(LookupError):
...
         self.target = target
         self.methods = methods
 
     def __rich_console__(self, console, options):
-        yield Text(f"{self.fname}{self.target} is ambiguous.")
-        yield Text()
-        yield Text("Valid matches are:")
+        yield Text(
+            f"{self.f_name}{self.target} is ambiguous between the following methods:"
+        )
```
PhilipVinc, line 43: The blank line here was deliberate, to mimic Julia (and IMO to give better readability)

```python
julia> f(1,1)
ERROR: MethodError: f(::Int64, ::Int64) is ambiguous.

Candidates:
  f(x::Int64, y)
    @ Main REPL[1]:1
  f(x, y::Int64)
    @ Main REPL[2]:1

Possible fix, define
  f(::Int64, ::Int64)

Stacktrace:
```

and in the PR this yield 
```python
AmbiguousLookupError: f(1, 1) is ambiguous.

Valid matches are:
f(x: int, y: Any)
    <function f at 0x1064d45e0> @ <ipython-input-2-b5a79a933176>:1
f(x: Any, y: int)
    <function f at 0x2896e5580> @ <ipython-input-3-e0ee52c2c087>:1
```

wesselb: @PhilipVinc Ahhh, that's clear! Let me change that back. Thanks. :)

PhilipVinc: (if however you prefer the more condensed version, do go ahead!)

wesselb: @PhilipVinc Do you know if we can indent the list of shown methods, like Julia does?

PhilipVinc:
```python
from rich.padding import Padding
from rich.columns import Columns

    def __rich_console__(self, console, options):
        yield Text(f"{self.f_name}{self.target} is ambiguous.")
        yield Text()
        yield Text("Candidates:")
        yield Padding(Columns([m.repr_mismatch() for m in self.methods]), (0,1))
```

the 1 is the horizontal padding

PhilipVinc: I think Julia uses 2 as padding, but then the tab looks too ugly inside. I think we could then replace the 4 spaces with just 2 spaces in methods.py

```
        res.append("\n    ")
        res.append_text(repr_pyfunction(self.implementation))
```
with
```
        res.append("\n  ")
        res.append_text(repr_pyfunction(self.implementation))
```

PhilipVinc: UI design is a rabbit hole

wesselb: > UI design is a rabbit hole

It really is...

Thanks for looking this up!! `Padding` works perfectly. I opted for 4 spaces to conform with `repr_pyfunction`.

wesselb:
@PhilipVinc I might add a few more tests, and then I think this is ready to be merged! Are you happy to merge it after a few more tests?

PhilipVinc:
I think that your last change of method.py , in MethodList, should render the method itself, not the mismatch...

plum/method.py
```diff
@@ -77,13 +82,120 @@ def __hash__(self):
...
+
+
+@rich_repr
+class MethodList(list):
+    "A list of :class:`Method`s which is nicely printed by :mod:`rich`."
+
+    def __rich_console__(self, console, options):
+        yield f"List of {len(self)} method(s):"
+        for i, method in enumerate(self):
+            yield Padding(Text(f"[{i}] ") + method.repr_mismatch(), (0, 4))
```
PhilipVinc, line 165: Sorry, that was confused.

I think here it should yield the method itself (which will call `methd.__rich_console__`) instead of `repr_mismatch`.

wesselb: `repr_mismatch` currently gives back a list of `rich` objects. I should look more closely at `rich` how precisely all of this works...

I tried something like

```
            yield Padding(Text(f"[{i}] "), (0, 4))
            yield Padding(method)
```

but that puts the two on different lines.

Just adding them seems to do the right thing

PhilipVinc: I suspect you could safely do
```python
Padding(Text(f"[{i}] ") + method. __rich_console__(console, options):(), (0, 4))
```
because rich console of methods are text objects which can be concatenated.
Alternatively
```
Padding(Text(f"[{i}] ").append(method. __rich_console__(console, options)):(), (0, 4))
```

wesselb: That worked! Though I had to sum the generated elements:

```python
            method_repr = method.__rich_console__(console, options)
            yield Padding(sum(method_repr, Text(f"[{i}] ")), (0, 4))
```

wesselb:
@PhilipVinc How does this look to you? Ready to merge?

wesselb:
@PhilipVinc Actually, you've indicated multiple times that you're ready to have this merged, so I'm gonna go ahead to merge this and tag a new release. :)

PR #111, opened 2023-09-30
[1/3] Very minor NFC (emerged from ruff)

PhilipVinc:
Some very very minor minor changes that I am removing from #110 to make review easier.
Those were detected by running ruff over the codebase.

The main question I have is why you were using `__copy__()`instead of `copy.copy` (which ok, just calls `__copy__`, but seems to me a bit more readable..)


wesselb (approved):
You're totally right that using `copy` is more clear! Thanks for this. :)

PR #112, opened 2023-09-30
[2/3] Split Signature into Signature and Method classes (preparatory work for pretty PR)

PhilipVinc:
This is split from #110 , based on top of #111 .
This PR does not contain any of the 'pretty printing code' (which is messy and long), but contains only the internal changes that make it possible to implement.

Mainly, this PR removes `ìmplementaition` and `return type` from `Signature` and introduces a new class , `Method` that pieces those things together (a method is just a signature, an implementation, and a name).

Conceptually I think this makes sense, because the signatures should only be used to dispatch. 
Also, I think this might make some experimentation easier later on... 

Apart from this two changes, everything is pretty mechanical. @wesselb there is one failing test that I don't understand why it fails. Do you have an idea?


PhilipVinc:
The main reason motivating this split is because, for proper repr support, I would like to have the "Method", aka Signature+implementation, to know the function name of the function they are attached to.

But putting this information inside a signature feels wrong.

Instead, a method object must always be attached to a function and therefore it makes sense for it to have a function name

PhilipVinc:
Tests are done on this one. It should be ready to go.

plum/signature.py
```diff
@@ -2,88 +2,114 @@
...
+    This object differs structurally from `inspect.signature` as
+    it only contains information necessary for performing dispatch.
+
+    For example, for the current implementation of plum, which
+    does not dispatch on keyword arguments, those are left out
+    of this signature object. Similarly, return type information
+    and argument names are not present.
 
     Attributes:
         types (tuple[type, ...]): Types of the arguments.
```
wesselb, line 32:
```suggestion
        types (tuple[:obj:`TypeHint`, ...]): Types of the call signature.
```

plum/signature.py
```diff
@@ -2,88 +2,114 @@
...
+        varargs: OptionalType = _default_varargs,
         precedence: int = _default_precedence,
-        implementation: Optional[Callable] = None,
     ):
+        """
+        Construct a `Signature` object.
+
+        Args:
+            *types (tuple[type, ...]): Types of the arguments.
+            varargs (type, optional): Type of the variable arguments.
```
wesselb, line 56:
```suggestion
            varargs (:obj:`TypeHint`, optional): Type of the variable arguments.
```

wesselb (approved):
I really like this change! This is fantastic, @PhilipVinc! Thank you so much for this. Structurally, I have no comments whatsoever. :)

I've left an array of formatting suggestions to make the style a slightly more consistent with the existing code. I hope you don't mind. If you're alright with these suggestions, then I'm happy for this to be merged, provided that the tests pass.

plum/resolver.py
```diff
@@ -62,12 +96,15 @@ class Resolver:
...
-        is_faithful (bool): Whether all signatures are faithful or not.
+        methods (list[:class:`.method.Method`]): Registered methods.
+        is_faithful (bool): Whether all methods are faithful or not.
     """
 
-    def __init__(self):
-        self.signatures: List[Signature] = []
+    __slots__ = ("methods", "is_faithful", "function_name")
+
+    def __init__(self, function_name="__noname__"):
```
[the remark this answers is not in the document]
PhilipVinc: yep. This is now solved.

PhilipVinc:
@wesselb everything's in shape now. Thanks for the through review! I am merging

PR #113, opened 2023-09-30
Reduce Lookuperror traceback noise

PhilipVinc:
This is an easy one.

Essentially discard all traceback informations from previous raises that just pollute the error

see this PR vs master (you can see that in master there are several errors in the stack, while in this PR only one)

Screenshot 2023-10-01 at 00 55 01


wesselb:
This is great! Thanks!! Wasn’t even aware of this syntax. :)

PR #116, opened 2023-10-05
Try to fix coveralls

PhilipVinc:


PhilipVinc:
@wesselb Coveralls does not show what lines are covered and not on PRs anymore, only the percentage. I think this should fix it... I will merge myself and not bother you with this

wesselb:
Thank you! :)

PR #122, opened 2023-10-16
added typing_extensions to list of hint modules

hodgespodge:


PhilipVinc:
Can you add a quick test as well?

PR #124, opened 2023-10-17
Replace flake8 with ruff

PhilipVinc:
Ruff is a substitute for flake8, which has recently released their first stable release. Nevertheless, they are becoming increasingly popular in the last few years because of their support for auto-fixing flake errors, and much simpler configuration and larger rule set (see https://docs.astral.sh/ruff/rules/)

The syntax is `ruff . --fix` to check and auto fix errors.

I replaced flake8 with ruff in pre commit hook and added it as a dev dependency

### DIFF
--- a/.github/workflows/formatting_check.yml
+++ b/.github/workflows/formatting_check.yml
@@ -1,29 +1,14 @@
 name: Check Formatting
+
 on:
   pull_request:
-
   push:
-    branches:
-      - master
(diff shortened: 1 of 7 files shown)


wesselb:
Yes, this is great! I've been using `ruff` at work and can confirm that it's great. Thanks, @PhilipVinc! :)

Issue #123, opened 2023-10-16
More issues with Annotated


hodgespodge:
Howdy again y'all. 

I'm having issues with Annotated that lead me to believe issue #120 was not actually resolved. 

1. After delving deeper It looks like functions with an Annotated type check just aren't working. Specifically, only the first specified constraint (the type) of the Annotated signature is actually being checked. 

```python
import sys
from beartype.vale import Is
from plum import dispatch
from typing_extensions import Annotated

class MyClass():
    def __init__(self) -> None:
        self.special_attr = 'special_value'
class AnotherClass():
    pass # notable difference: no special_attr

class ClassWithAnnotatedTypeChecker():

    class_type_check = Annotated[object, Is[lambda value: hasattr(value,'special_attr')]]

    @dispatch
    def myFunction(self, value: class_type_check):
        if hasattr(value,'special_attr'):
            print(f'Value has special_attr={value.special_attr}')
        else:
            print(f'ERROR: This code should not be reached- 1')


    string_type_check = Annotated[str, Is[lambda value: len(value) > 3]]

    @dispatch
    def myFunction(self, value: string_type_check):
        if len(value) > 3:
            print(f'Value is a string={value}')
        else:
            print(f'ERROR: This code should not be reached- 2')

    @dispatch
    def myFunction(self, value: int):
        print(f'Value is an int={value}')

my_class_with_annotated_type_checker = ClassWithAnnotatedTypeChecker()

my_class = MyClass()
print(Is[lambda value: hasattr(value,'special_attr')].is_valid(my_class)) # Returns True
my_class_with_annotated_type_checker.myFunction(my_class) # Works

another_class = AnotherClass()
print(Is[lambda value: hasattr(value,'special_attr')].is_valid(another_class)) # Returns False
my_class_with_annotated_type_checker.myFunction(another_class) # Should fail to dispatch... Instead, it dispatches to the str function

print(Is[lambda value: len(value) > 3].is_valid('Hello')) # Returns True
my_class_with_annotated_type_checker.myFunction('Hello') # Works

print(Is[lambda value: len(value) > 3].is_valid('Hi')) # Returns False
my_class_with_annotated_type_checker.myFunction('Hi') # Should fail to dispatch... Instead, it dispatches to the str function

my_class_with_annotated_type_checker.myFunction(1)
```

PhilipVinc:
MWE

```python
In [1]: import sys
   ...: from beartype.vale import Is
   ...: from plum import dispatch, Signature
[...]

PR #126, opened 2023-12-14
Fix annotated #123

hodgespodge:
This should fix issue #123 

This changes get_type_hints(f, include_extras=True) to use the "include_extras" flag which does _not_ remove annotated typing information. Also, cleans up some inconsistency with typing vs typing_extensions to support annotated in py 38

### DIFF
--- a/.github/workflows/formatting_check.yml
+++ b/.github/workflows/formatting_check.yml
@@ -1,29 +1,17 @@
 name: Check Formatting
+
 on:
   pull_request:
-
   push:
-    branches:
-      - master
-
+    branches: [master]
 
 jobs:
-  check:
-    name: Check Formatting
+  pre-commit:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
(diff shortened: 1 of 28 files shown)


wesselb:
Hey @hodgespodge! By closing the PR, do you mean that you would not like to go through with it?

hodgespodge:
Howdy @wesselb,

Sorry about the confusion and thanks for checking. I would like to go through with the PR. I had initially second guessed some of my changes but I think they're probably fine. I'll reopen the PR as a draft since there are a few more things I want to double check. In the mean time, if anything problematic stands out, please let me know!

Also, I wasn't able to run some of the benchmarking tests locally but I assume that probably isn't a big issue.

wesselb:
@hodgespodge Not a problem at all. :) The PR looks pretty good to me, assuming that everything works as expected!

hodgespodge:
@wesselb Alrighty this PR is ready for review

wesselb:
@hodgespodge I think this looks great! Some very minor formatting check is failing, but I'll fix that myself. I'll resolve the conflict with `plum/signature.py`, merge this, and tag a new release right away. :)

PR #128, opened 2024-01-06
Fix "overload" is not exported warning

yukinarit:
Hi I found importing new "overload" function produces the following warning.

image

### DIFF
--- a/plum/__init__.py
+++ b/plum/__init__.py
@@ -14,7 +14,7 @@ from .alias import *  # noqa: F401, F403
 from .autoreload import *  # noqa: F401, F403
 from .dispatcher import *  # noqa: F401, F403
 from .function import *  # noqa: F401, F403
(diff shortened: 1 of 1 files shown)


wesselb:
Hey @yukinarit! Thanks for submitting this PR. :)

I'm slightly confused as to why this would solve the error, because `overload` should now be exported by `__init__.py`. Nevertheless, I think merging your change would be good anyway since `get_overloads` is then also directly accessible from `__init__.py`. I'll therefore merge this right away!

PhilipVinc:
I think the error was caused by a function having the same name as the module/file which confused pylance. 
Be aware that type checkers should ignore all `*` imports, even if pylance for the time being does check up on them.

wesselb:
@PhilipVinc Ah, thanks for the clarification. :) That's good to know.

PR #133, opened 2024-03-16
feat: type_unparametrized

nstarman:
For getting the un-parametrized type of an object. This is useful for doing `type(obj)(*args, **kwargs)`.

### DIFF
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -1,4 +1,4 @@
-from typing import Union, TypeVar
+from typing import Union, TypeVar, Type
 
 import beartype.door
 from beartype.roar import BeartypeDoorNonpepException
@@ -430,7 +430,7 @@ def type_parameter(x):
     )
 
 
-def type_unparametrized(q: T) -> type[T]:
+def type_unparametrized(q: T) -> Type[T]:
     """Return the unparametrized type of an object.
 
     :mod:`plum.parametric` produces parametric subtypes of classes.  This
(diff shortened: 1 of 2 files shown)


wesselb:
Hey @nstarman!

Thanks for submitting two PRs. That’s massively appreciated. :)

I am currently away and unable to review the PR, but should be back soon. I will get back to you some time next week.

wesselb:
Hey @nstarman! I think this looks great and is basically ready to merge. :)

Instead of doing `type(x).mro()[1]`, would it be safer to store the parent class in an attribute of the parametrised class? Then this could instead be implemented as e.g. `type(x)._parent_class`.

In addition, would you be able to fix the test for Python 3.8? I think you should be able to use `typing.Type` instead.

nstarman:
@wesselb, I've made the suggested corrections. Now `@parametric` adds the methods `__class_nonparametric__()` and `__class_unparametrized__()`, which return the original class and the unparameterized class, respectively.

nstarman:
@wesselb, if you want I can also add a `type_nonparametric` function that exposes `__class_nonparametric__`.

wesselb:
@nstarman Apologies for being super slow. Work has been very hectic and eating up all my time.

This is beautiful PR. Very clean! Thanks so much for the contribution. I'm merging this right away. :)

> @wesselb, if you want I can also add a type_nonparametric function that exposes __class_nonparametric__.

That would make for a nice consistent interface. Up to you! :)

wesselb:
I'll release a new version which incorporated this once CI passes!

PR #135, opened 2024-03-18
feat: return conversion function

nstarman:
Allows for the conversion function to be used after being registered.


wesselb:
Hey @nstarman! Very sensible. Thanks for the contribution. Merging this right away. :)

PR #142, opened 2024-04-21
Use Sybil for doctests

nstarman:
This PR add docstring testing and testing of the docs via Sybil (https://sybil.readthedocs.io/en/latest/).
I'm not sure what style of docstrings you want... My last PR was in the NumPy style, but the rest of the docs are in Google's...

If you don't like the `try: ...; ... except Exception as e: print(f"{type(e).__name__}: {e}")`, I think we can subclass `sybil.parsers.myst.PythonCodeBlockParser` and modify it so that the `try: except` operates behind the scenes. IMO I like that it's explicit and the examples are safer to copy-paste.

Found:
- 1 bug (plum/util.py) where classes generated in an atypical fashion, e.g. as in the docs have `__module__ is None`
- Many out-of-date / incorrect docs — including the ones I just contributed!

Squash merge is fine!

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -4,14 +4,49 @@ from __future__ import annotations
 
 from doctest import ELLIPSIS, NORMALIZE_WHITESPACE
 
+import pytest
 from sybil import Sybil
-from sybil.parsers.rest import DocTestParser, PythonCodeBlockParser, SkipParser
+from sybil.parsers.myst import DocTestDirectiveParser as MarkdownDocTestParser
+from sybil.parsers.myst import PythonCodeBlockParser as MarkdownPythonCodeBlockParser
+from sybil.parsers.myst import SkipParser as MarkdownSkipParser
+from sybil.parsers.rest import DocTestParser as ReSTDocTestParser
+from sybil.parsers.rest import PythonCodeBlockParser as ReSTPythonCodeBlockParser
+from sybil.parsers.rest import SkipParser as ReSTSkipParser
 
-pytest_collect_file = Sybil(
+OPTIONS = ELLIPSIS | NORMALIZE_WHITESPACE
+
+
+@pytest.fixture(scope="module")
+def use_clean_dispatcher():
+    import plum
+
+    # Save the original dispatcher
+    dispatcher = plum.dispatch
+    # Swap the dispatcher with a temporary one
+    temp_dispatcher = plum.Dispatcher()
+    plum.dispatch = temp_dispatcher
+    yield
+    # Restore the original dispatcher
+    plum.dispatch = dispatcher
+
+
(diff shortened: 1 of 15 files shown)


nstarman:
Hmm. This passes on py3.12

nstarman:
I think the failure is unrelated

plum/util.py
```diff
@@ -122,7 +122,7 @@ def is_in_class(f: Callable) -> bool:
 
 
 def _split_parts(f: Callable) -> List[str]:
-    qualified_name = f.__module__ + "." + f.__qualname__
+    qualified_name = (f.__module__ or "") + "." + f.__qualname__
```
wesselb, line 125: Shouldn't `f` always have a `__module__`? Or is it during doctests that `__module__` may not be set?

If `__module__` is not set, would it be more correct to just return `f.__qualname__.split(".")`?

conftest.py
wesselb, line 1: Is it possible to move this to `tests/conftest.py`?

plum/parametric.py
```diff
@@ -283,28 +283,48 @@ def __class_nonparametric__(cls):
         :mod:`plum.parametric` produces parametric subtypes of classes. This
         method can be used to get the non-parametric type of an object.
 
+        See Also
```
wesselb, line 286: This docstring is beautiful, but unfortunately I don't think it is rendered in the docs because it is a function inside a function. :(

wesselb (commented):
This looks awesome, @nstarman. Thank you so much for this!

I've got a few very minor comments/questions, which I left as comment in the PR.

plum/util.py
nstarman: Agreed!

conftest.py
nstarman: The same `conftest.py` applies to both `tests`, `plum`, and `docs`

plum/parametric.py
nstarman: They should be exposed to the user on the derived classes, e.g. with `parametric(Class).__class_unparametrized__?` or `help(parametric(Class).__class_unparametrized__)` or `parametric(Class).__class_unparametrized__.__doc__`

plum/util.py
nstarman: I used a ternary

plum/parametric.py
wesselb: Yes, you're totally right! It's just a shame that it doesn't appear in the online docs.

conftest.py
wesselb: Right! Thanks. :)

wesselb:
Again, this is a fantastic PR, @nstarman! Merging this right away. :) Thanks so much!!

PR #145, opened 2024-04-27
feat: type_nonparametric

nstarman:


wesselb (approved):
Thanks again for the super clean PR!! Merging and releasing an update right away.

Apologies for the slow turn around. It's a busy time.

PR #146, opened 2024-04-27
Add type hints to the convert methods

nstarman:


plum/promotion.py
```diff
@@ -57,11 +70,13 @@ def add_conversion_method(type_from, type_to, f):
     """
 
     @_convert.dispatch
-    def perform_conversion(obj: type_from, _: type_to):
+    def perform_conversion(obj, _):
```
wesselb, line 73: The tests might be failing because the type annotations here are not there anymore.

nstarman: Reverted.

wesselb:
This looks great!! Thanks :). Merging right away!

PR #148, opened 2024-04-27
ci: buff out the ruff edges

nstarman:
A sampling of the improvements that leaning into ruff brings.
I used the most common codes, and got ruff to replace `black` and `isort`.
There's more we can configure ruff to do, e.g. blackening code in the docstrings. 

I also added some useful pre-commit hooks.


plum/signature.py
```diff
@@ -4,9 +4,10 @@
 from copy import copy
 from typing import Callable, List, Set, Tuple, Union
 
+from rich.segment import Segment
```
nstarman, line 7: I made beartype a "first-party" package.

wesselb: Makes total sense, thanks! :)

wesselb:
This is great!! Thanks, @nstarman :)

PR #150, opened 2024-05-14
Add test case from #117 checking for varargs ambiguity

PhilipVinc:
I recently noticed that for some reason a test case I had added in #119 that previously did not pass test, now pass.

The test is that the following raises an `AmbiguousLookupError`
```python
def f(x : int, *y: A):
 pass
def f(x: int, *y : B):
 pass

f(1)
```

I would like to add this to the test set of plum just so that it's included.

Note that #117 is still not solved as we still do not pass this other test
```python
assert Sig(int, varargs=int) < Sig(int, Num)
```
but that's a step in the right direction...

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -1,6 +1,7 @@
 import sys
 import textwrap
 import typing
+from numbers import Number as Num
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


wesselb:
Thanks, this looks great! :) Merging right away.

PR #151, opened 2024-05-24
Improve the dispatch algorithm by avoiding unnecessary ambiguities

wesselb:
@PhilipVinc, I've taken some time to go over #119. I've attempted to implement a very similar approach in a way that makes it very clear where precisely we deviate from the subset relationship and why that is justified.

Would you be able to check if this works for your use case? If it does, I'll add some more extensive tests based on the examples from the other PR.


tests/test_signature.py
```diff
@@ -114,6 +114,35 @@ def test_expand_varargs():
     assert s.expand_varargs(4) == (int, int, float, float)
 
 
+def test_varargs_tie_breaking():
```
PhilipVinc, line 117: Let's just add a reminder in here 

```suggestion
def test_varargs_tie_breaking():
    # Those are related to Bug #117
```

tests/test_signature.py
```diff
@@ -114,6 +114,35 @@ def test_expand_varargs():
...
+    assert Sig(int) < Sig(int, varargs=int)
+    assert not Sig(int) >= Sig(int, varargs=int)
+    assert Sig(int, varargs=int) < Sig(int, Num)
+    assert not Sig(int, varargs=int) >= Sig(int, Num)
+    assert Sig(int, int, varargs=int) < Sig(int, Num)
+    assert not Sig(int, int, varargs=int) >= Sig(int, Num)
+
+
```
PhilipVinc, line 125: In #119 I also had added those tests. You might want to include them as well...

```suggestion

    assert not (Sig(int, Num) < Sig(int, varargs=int))
    assert not (Sig(int, varargs=Num) < Sig(int, varargs=int))
    assert not (Sig(int, varargs=Num) <= Sig(int, varargs=int))

    class A:
        pass

    class B:
        pass

    assert not Sig(int, varargs=A) < Sig(int, varargs=B)
    assert not Sig(int, varargs=B) < Sig(int, varargs=A)

    assert not Sig(int, varargs=Num) < Sig(Num, varargs=int)
    assert not Sig(Num, varargs=int) < Sig(int, varargs=Num)


```

PhilipVinc (approved):
Apart from those extra tests, this PR fixes the issues I had with plum and would allow us to stop vendoring it, so I'm supporting it in full.

Moreover, I think this is equivalent to my past PR (albeit much more elegant) so again, I think that's the right approach.

Thanks!

wesselb:
@PhilipVinc Perfect! :) I'll then finalise this PR by adding the missing tests and merge it.

PhilipVinc:
@wesselb can I merge? seems all is good

wesselb:
@PhilipVinc I'm making some final tweaks. If it's ok, I'll merge this in the next ~hour.

wesselb:
@PhilipVinc Actually, that was faster than expected. I think it looks good. I've added a comment what precisely need to be fixed to fully resolve #117. I think it's feasible to fix, but beyond the scope of this PR. Happy to merge this now?

PhilipVinc:
happy to!

wesselb:
Hurray! I'm pretty happy this is finally partially fixed. :)

PR #156, opened 2024-06-07
ci: ensure docstrings are formatted

nstarman:
@wesselb this turns on ruff's check / auto-fix  to ensure code in docstrings is correctly formatted.


wesselb:
Thanks @nstarman :) That’s a very helpful check!

Issue #147, opened 2024-04-27
Configure mypy for CI


nstarman:
It would be great to add type annotations to `plum`. They should be tested by `pre-commit` and configured in `pyproject.toml`.

wesselb:
Hey @nstarman! I fully agree that adding type annotations and eanbling `mypy` would be helpful.

One problem is that multiple dispatch really doesn't play nicely with type checkers like `mypy`. [There are some patterns where you can get things to work together](https://beartype.github.io/plum/integration.html), but getting `mypy` to be happy with code that uses the full feature set of Plum is really challenging. This is the main reason I haven't enabled `mypy` yet.

nstarman:
Agreed! It often doesn't work with mypy. Where it does work, having static checks are the gold standard. Where it doesn't, maybe using `beartype` itself would be sufficient? mypy in pre-commit can be configured to only run on select files. My suggestion is to configure mypy nicely and filter out all files. Then select files can be added as they are typed and if they will support static checks.

wesselb:
@nstarman That sounds very reasonable. I'd be very happy to only enable `mypy` for the files, or parts thereof where it would work. That would be a nice addition! :)

I'm going to be busy for about as week, but should have some capacity after that.

PR #157, opened 2024-06-07
refactor: small cleanup

nstarman:
Slowly adding typing.
This would be good to combine with #147.

### DIFF
--- a/plum/dispatcher.py
+++ b/plum/dispatcher.py
@@ -39,7 +39,7 @@ class Dispatcher:
 
     def __call__(
         self, method: Optional[T] = None, precedence: int = 0
-    ) -> T | Callable[[T], T]:
+    ) -> Union[T, Callable[[T], T]]:
         """Decorator to register for a particular signature.
 
         Args:
--- a/plum/dispatcher.py
+++ b/plum/dispatcher.py
@@ -1,3 +1,5 @@
+import sys
+from dataclasses import dataclass, field
 from functools import partial
 from typing import Any, Dict, Optional, Tuple, TypeVar, Union, overload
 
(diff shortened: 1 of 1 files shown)


wesselb:
Thanks again for the PR, @nstarman! This change looks good to me.

One question about this: What’s the benefit of making Dispatcher a dataclass c.f. a normal class? Is it that you can set `slots`, presumably for a performance boost?

PhilipVinc:
Slots means a tiny bit less overhead memory. But that does not impact anything we do because that’s sort of a singleton.

I like data classes when they can help make the codec more clear , but in general they are a pain to work with because you lose control over __init__ …

‘I’m not sure it’s the best idea to use them here. Though in that particular case it won’t hurt either. But it can add friction for future refactors

nstarman:
> Slots means a tiny bit less overhead memory. But that does not impact anything we do because that’s sort of a singleton.

Only here. `Dispatcher` is part of the public API and other libraries create their own instances.

> I  like data classes when they can help make the codec more clear , but in general they are a pain to work with because you lose control over init …

They reduce boilerplace, add introspection and interop tools (e.g. `replace`), and they don't relinquish control over `__init__` (if an `__init__` is defined on a class, not a superclass, `@dataclass` will not rewrite it).

wesselb:
Thanks for the discussion, both! Since it’s a fairly small change and comes with some benefits (albeit minor ones), I think it’s worthwhile trying to make Dispatcher a dataclass. Should we later decide that we don’t like it after all, because it could make future changes more difficult as @PhilipVinc points out, it would be easy to make Dispatcher a normal class again. Will merge right away! :) Thanks again, @nstarman!

Issue #154, opened 2024-06-02
JIT + plum


nstarman:
Starting from the discussion in #152.

This is my quick thoughts on a suggestion for the docs and potentially an improvement to `dispatch`.

```python
import jax
from jaxtyping import Int, Array, Float, Float64
from plum import dispatch
```

First let's see the speed bump

```
@dispatch
def nojit_func(a: int | Int[Array, ""]) -> int | Int[Array, ""]:
    return 2 * a

# %timeit func(3)
# 8.91 µs ± 33 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

```
@jax.jit
@dispatch
def jit_func(a: int | Int[Array, ""]) -> int | Int[Array, ""]:
    return 2 * a

# func(3)   # trigger jit
# %timeit func(3)
5.03 µs ± 19.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Noticeable even for this trivial operation. 🎉 

Now let's consider the problem of jitting a dispatch

```python
@jax.jit
@dispatch
def func(a: int | Int[Array, ""]) -> int | Int[Array, ""]:
    return 2 * a

func(3) 
# Array(6, dtype=int64, weak_type=True)

try:
    out = func(3.0)
except Exception as e:
    print(e)
else:
    print(repr(out))
# `func(Traced<ShapedArray(float64[], weak_type=True)>with<DynamicJaxprTrace(level=1/0)>)` could not be resolved.

@jax.jit
@dispatch
def func(a: float | Float[Array, ""]) -> float | Float[Array, ""]:
    return 2.1 * a

func(3.0)
# Array(6.3, dtype=float64, weak_type=True)

@jax.jit
@dispatch(precedence=1)
def func(a: Float64[Array, ""]) -> Float64[Array, ""]:
    return 2.2 * a

func(3.0)
# Array(6.3, dtype=float64, weak_type=True)  # THIS IS WRONG!

func.clear_cache()

func(3.0)
# Array(6.6, dtype=float64, weak_type=True)   # THIS IS WRONG!

```

We need to clear the cache for this to work correctly.
When `@dispatch` is called it could check if there are any other dispatches and then it could try clearing the cache if there are.
If there were a jit backend agnostic way of doing that (since JAX is not a dependency) that would be great.

PhilipVinc:
[...]

PR #158, opened 2024-06-07
feat: clear jax cache on register

nstarman:
The idea is to "structurally" identify jax. jax.jit produces as jaxlib object with attribute clear_cache. That's sufficient to identify JAX jitted functions.
I've put notes for longer-term plans.

Needs tests.

### DIFF
--- a/plum/dispatcher.py
+++ b/plum/dispatcher.py
@@ -110,6 +110,14 @@ class Dispatcher:
         f = self._get_function(method)
         for signature in signatures:
             f.register(method, signature, precedence)
+
+        # Hooks to clear the JIT caches of various libraries.
+        # TODO: this needs to be systematized. This should probably work
+        #       by entry points to allow for any JIT library to be added.
+        # JAX:
+        if "jax" in type(method).__module__ and hasattr(method, "clear_cache"):
+            method.clear_cache()
+
         return f
 
     def clear_cache(self):


nstarman:
@wesselb @PhilipVinc what do you think of this simple solution? There's definitely room for improvement, as indicated in the comment, but I think since this is private API it's good enough to get in as is, then iterate. WDYT?

PhilipVinc:
I would be a bit against this.

The thing you proposed works when directly jitting a dispatch function, like
```python
@dispatch
@jit
def myfun(a: jax.Array, b:int):
   return a * b

@dispatch
@jit
def myfun(a:  jax.Array, b:float):
   return a + b
```

however it does not detect usages like
```python
@dispatch
def myfun(a: jax.Array, b:int):
   return a * b

@dispatch
def myfun(a:  jax.Array, b:float):
   return a + b

@jit
def my_algorithm(a, b):
    return 12 * myfun(a, b)

# this causes jax to jit, 
my_algorithm(a, b)

@dispatch
def myfun(a:  jax.Array, b:float):
   return a + b + 10

# jax has no idea that the function changed, so he will return same result as before
my_algorithm(a, b)

```


This also applies to your function if inside a jitted function you call other dispatch functions.

This is the use-case that should be addressed in my opinion, otherwise it's too brittle.

nstarman:
@PhilipVinc some really good points!

Just for completeness, this also helps address the case

```python
@jit
@dispatch
def myfun(a: jax.Array, b:int):
   return a * b

@jit
@dispatch
def myfun(a:  jax.Array, b:float):
   return a + b
```

(The decorator order is important jin-wise)

But your are correct that this doesn't work for "higher-order" functions that call other functions which themselves are multiply-dispactched, as you show later in your comment.

nstarman:
@wesselb, maybe is this related to your suggestion of a `cache` ? Or is this a separate complication?

PhilipVinc:
How does your PR work with `jit(dispatch(...))` ? 
the dispatch logic is executed before the hitting happens so I do not understand how it would work correctly in this case?

nstarman:
See the opening comment of https://github.com/beartype/plum/issues/154#issue-2329848202.
`jit` compiles away the dispatching. Prior to this PR, when a new dispatch was added JAX did not update which function it was going to use. This is exactly what you point out in `my_algorithm`, but applies here to `myfun`. With this PR when a new dispatch is added to `myfun` it wipes the cache and forces the next usage to JIT again.

I haven't tested your "higher-order" case, but I hope (🤞) adding `jit` to `myfun` would actually work with this PR because 
`my_algorithm` wouldn't compile `myfunc` away, but keep it as a distinct function. 

```
@jit
@dispatch
def myfun(a: jax.Array, b:int):
   return a * b

@jit
@dispatch
def myfun(a:  jax.Array, b:float):
   return a + b

@jit
def my_algorithm(a, b):
    return 12 * myfun(a, b)  # myfun is not compiled away
```

If the `jit` on `myfun` had `inline=True`, then IDK what would happen.

But overall I agree this is still brittle and thus not an optimal solution!

PhilipVinc:
But isn't this
```python
        # Hooks to clear the JIT caches of various libraries.
         # TODO: this needs to be systematized. This should probably work
         #       by entry points to allow for any JIT library to be added.
         # JAX:
         if "jax" in type(method).__module__ and hasattr(method, "clear_cache"):
             method.clear_cache()
```

only going to detect if you `@dispatch@jit`? 

```python
import jax

def dispatch(fun):
	print("Fun:", type(fun).__module__)
	return fun


@jax.jit
@dispatch
def test(x, y):
	return x+y

print("fun now:", type(test).__module__)
```
outputs
```
Fun: builtins
fun now: jaxlib.xla_extension
```

so in this case it would not be triggered

nstarman:
Ah. Good catch. I didn't write tests yet for this, which would have caught this mistake. This is the difference between the example I gave in #154 and here.
Yes, I need to refactor this to search through all the registrants, not the current method.
We definitely do want to detect both `jit(dispatch(...` and `dispatch(jit(...`.
But neither approach will work for jitted functions that call a separate dispatched function.
Is there any way around this?!

nstarman:
It's definitely not ideal, but so long as a user imports in all modules (as in registers everything) then it's not so bad. It's only when new dispatches are registered that this becomes annoying.

nstarman:
Maybe @patrick-kidger can advise on how to best make `plum` and `jax` play well together?

patrick-kidger:
AFAIK it's not really possible.
- Adding a new dispatch rule is mutating global state (the methods table);
- I do not know of any way to have JAX automatically re-JIT when changing global state.

Tbh I think supporting this may be an antipattern anyway. Everything prior to `jax.jit` is basically the 'source code', which on your first run you then compile. Looking at other languages: tying to automatically recompile on detecting your source code is nonstandard in C++/Rust-type languages, and in Julia caused a lot of headache: the fact that they *were* handling this meant they were silently footgunning their compiletimes with lots of cache invalidations.

That aside I really don't like the idea that `plum` should try to do something based on specific third-party libraries. If everyone did this it would be incredibly hard to reason about what my code does. I would much rather they each just do their thing without trying to interfere with each other.

nstarman:
@patrick-kidger. Thanks for the info!
Are you suggesting that instead we just add something to the docs showing how `jit` + `dispatch` can work together, but also showing how this can be dangerous when adding new dispatches, and suggesting `jax.clear_caches` for that case?
Instead of clearing the cache, this PR could check if there is a non-empty cache and raise a warning.

wesselb:
> Are you suggesting that instead we just add something to the docs showing how jit + dispatch can work together, but also showing how this can be dangerous when adding new dispatches, and suggesting jax.clear_caches for that case?
Instead of clearing the cache, this PR could check if there is a non-empty cache and raise a warning.

This also sounds like a good approach to me. :)

> @wesselb, maybe is this related to your suggestion of a cache ? Or is this a separate complication?

What I had in mind was slightly different. The cache would remember all dispatch decisions. The next time the function is called, instead of running the resolver on the arguments, the resolver would not look at the arguments, but immediately return the previously resolved method from the cache. This way you could "compile away" the overhead of dispatch.

PR #162, opened 2024-06-13
Test Coveralls status check

wesselb:
This deletes some tests. Coveralls is supposed to fail.


nstarman:
I don't even see coveralls in the Checks!

wesselb:
@nstarman I thought that this is the Coverall status update:

image

The confusing thing is that this does look a bit different from the [screen recording on the Coveralls GitHub Action site](https://github.com/marketplace/actions/coveralls-github-action#demo):

image
 
Maybe I need to add a separate GitHub action?

nstarman:
Yeah, I was expecting it to look like:

notifications-status-updates

PR #165, opened 2024-06-16
Add keyword argument warning

PhilipVinc:


wesselb (approved):
Thanks, @PhilipVinc! Given all the confusion, putting this in the README is imminently sensible.

PhilipVinc:
@wesselb I think it might be a good idea to open a discussion about whether this choice is a good choice.

Unfortunately I realised over the last years that it's extremely unpopular with Python programmers, counterintuitive, and does not fit the standard style paradigm of setting keyword arguments.

I started to ask myself if it might be worth revisiting the choice? Even though It would open a huge can of worms on how to embed argument names into dispatch algorithm.

An alternative thing that I had thought about a few years ago was to use [PEP570 positional only syntax](https://peps.python.org/pep-0570/) which is essentially what we do now.

Back at the day, plum supported python 3.7 as well, so it was not an option, but nowadays we can leverage it because our requirement is python 3.8.

One possibility (though annoying) would be to oblige users to insert the `/` in all signatures of `@dispatch` decorated functions. This way it would be more evident what plum is doing. 

Another thing we can do in the meantime is to actually test for this very common error. If we see that no signature matches but keyword arguments are provided, we can check if some registered function have this argument name as an argument, and in that case just raise an informative error and give a link to the documentation...

wesselb:
@PhilipVinc I agree. It seems to be a thing that people run into very regularly.

This, however, seems like a huge can of worms to me. For example, consider

```python
@dispatch
def f(x: int, y: float):
    ...

@dispatch
def f(y: float, x: int):
    ...
```

Under the current designate-by-position system, these are two different methods. If we identify arguments by name, then possibly these two methods would be considered the same and hence the second definition would overwrite the first? If you don't go that route, but consider the methods different, then that might be unexpected too.

Perhaps one middle ground is this pattern that's come up before a few times:

```python
def f(x, y):
    return _f(x, y)

@dispatch
def f(x: int, y: float):
    ...
```

It would perhaps be possible to automatically generate such an "interface" method, perhaps with syntax like this:


```python
@dispatch.interface
def f(x, y):
    ...

@dispatch
def f(x: int, y: float):
    ...
```

PR #168, opened 2024-06-23
Fix default arguments bug and improve `MethodRedefinitionWarning`

wesselb:
What the title says. WIP.

### DIFF
--- a/plum/function.py
+++ b/plum/function.py
@@ -494,6 +494,9 @@ class _BoundFunction:
             method = self._f.invoke(type(self._instance), *types)
             return method(self._instance, *args, **kw_args)
 
-        wrapped_method.__wrapped_by_plum__ = self._f
+        # We set `f.__wrapped_by_plum__` for :func:`Function.invoke`, but here we do
+        # not: this method has `self._instance` prepended to its arguments, so there
+        # is no "wrapped method". In addition, bound functions cannot be directly
+        # extended, so unwrapping is likely never desired.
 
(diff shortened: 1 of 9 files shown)


PhilipVinc:
@wesselb The warning is very noisy and several users are complaining. Can we revert the previous PR until we have something that works fine?

wesselb:
@PhilipVinc I agree that it's noisy, but it's also already uncovered some issues, so it does have significant utility.

How about an opt-out mechanism, where the warning is supressed if the module of the implementation that overwrites the current method is part of a list?

You'd then do something like

```python
from plum import opt_out_method_redefinition_warning_for_package

opt_out_method_redefinition_warning_for_package("netket")
```

PhilipVinc:
What issues?

plum/function.py
```diff
@@ -417,6 +417,8 @@ def invoke(self, *types: TypeHint) -> Callable:
         def wrapped_method(*args, **kw_args):
             return _convert(method(*args, **kw_args), return_type)
 
+        wrapped_method.__wrapped_by_plum__ = method
```
PhilipVinc, line 420: Can you add a docstring to explain what this does and why?

wesselb: Definitely! Will also add a test which illustrates the edge case. Sorry, haven’t had a moment yet to do this.

wesselb: Added in the tests!

PR #170, opened 2024-06-25
fix: warning on self promotion

### LINKED ISSUE #169: Problems with parametric types and promotion
Consider 

```python
from dataclasses import dataclass
from plum import promote, parametric, add_promotion_rule
from typing import Any

@parametric
@dataclass
class Class:
    a: int | float
    b: int | float

    @classmethod
    def __init_type_parameter__(self, a: type, b: type) -> tuple[type, type]:
        """Check whether the type parameters are valid."""
        return a, b

    @classmethod
    def __infer_type_parameter__(self, a: Any, b: Any) -> tuple[type, type]:
        """Inter the type parameter from the arguments."""
        print(a, b)
        return type(a), type(b)

x = Class(1, 2)  # Class[int, int]
y = Class(1.0, 2.0)  # Class[float, float]
```

`plum` appears to auto-register a self-promotion pathway, but parametric types produce subclasses, so the promotion doesn't cover the types.

```python
promote(x, y)
# TypeError: No promotion rule for `__main__.Class[int, int]` and `__main__.Class[float, float]`.
```

If I register a generic promotion for `Class` to cover its subclasses then I hit the the `MethodRedefinitionWarning`

```python
add_promotion_rule(Class, Class, Class)
promote(x, y)
# MethodRedefinitionWarning: `Method(function_name='_promotion_rule', ...
# (Class[int, int](a=1, b=2), Class[float, float](a=1.0, b=2.0))
```

There does not appear to be a way to register a generic dispatch for parametric types without incurring `MethodRedefinitionWarning`!

nstarman:
Fixes #169

### DIFF
--- a/tests/test_promotion.py
+++ b/tests/test_promotion.py
@@ -158,3 +158,8 @@ def test_inheritance(convert, promote):
     assert promote(n, Rat()) == (n, "Num from Rat")
     assert promote(Re(), n) == ("Num from Re", n)
     assert promote(Re(), Rat()) == ("Num from Re", "Num from Rat")
+
+    # Test that explicit self-promotion works.
+    # This should also trigger the "escape hatch" in `add_promotion_rule`.
+    add_promotion_rule(Num, Num, Num)
+    assert promote(n, n) == (n, n)
--- a/plum/promotion.py
+++ b/plum/promotion.py
@@ -139,7 +139,9 @@ def add_promotion_rule(type1, type2, type_to):
     def rule(t1: type1, t2: type2):
         return type_to
 
-    # If the types are the same, the method will get overwritten.
+    # If the types are the same, we don't need to add the reverse rule.
+    if type1 is type2:
+        return  # Escape early.
 
     @_promotion_rule.dispatch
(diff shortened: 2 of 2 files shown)


nstarman:
I think I've fixed all other bugs, at least in `quaxed`. This is for `unxt`. Next we'll see what `coordinax` uncovers!

wesselb:
@nstarman The warning seems to uncover a lot! 😅

Thanks for this PR. :) Merging right way. I’ll add one more test myself and then release a new version!

tests/test_promotion.py
```diff
@@ -158,3 +158,8 @@ def test_inheritance(convert, promote):
     assert promote(n, Rat()) == (n, "Num from Rat")
     assert promote(Re(), n) == ("Num from Re", n)
     assert promote(Re(), Rat()) == ("Num from Re", "Num from Rat")
+
+    # Test that explicit self-promotion works.
+    # This should also trigger the "escape hatch" in `add_promotion_rule`.
+    add_promotion_rule(Num, Num, Num)
```
nstarman, line 164: The cookie monster of promotion rules.

PR #172, opened 2024-06-28
feat: type annotate signature.py

nstarman:
Misc additional type hints to signature.py


nstarman:
I don't think the failure is related. ?

wesselb:
@nstarman, no, the flaky test is becoming really annoying! I should investigate if something is off or whether the test is just too strict.

Thanks again for yet another PR!! Merging right away.

PR #173, opened 2024-06-28
feat: deprecate Val

nstarman:
I noticed in the code that the requirements to begin the deprecation process were fulfilled.

### DIFF
--- a/docs/parametric.md
+++ b/docs/parametric.md
@@ -242,39 +242,36 @@ def i_expect_that(that: Kind["that"]):
     ...
 ```
 
-## `Val`
+## `typing.Literal` (and `Val`)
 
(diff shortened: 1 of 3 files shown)


wesselb:
@nstarman You're totally right! :) This looks great!! Let me know when the PR is ready for review.

@PhilipVinc, I believe you used `Val` in some places in the past. Do you also think `Val` can be deprecated, or is there still an edge case where `Literal` cannot replace `Val`?

PhilipVinc:
We stopped using them long ago in favour of Literal.

wesselb:
@PhilipVinc Thanks! Lets merge this then whenever @nstarman thinks its ready. :)

PR #175, opened 2024-06-28
feat: deprecate multihash

### LINKED ISSUE #174: deprecate `multihash`?
It's only used once in the code and it's just `hash(tuple(args))`

nstarman:
Fixes #174

### DIFF
--- a/plum/__init__.py
+++ b/plum/__init__.py
@@ -23,6 +23,9 @@ from .type import *  # noqa: F401, F403
 from .type import resolve_type_hint
 from .util import *  # noqa: F401, F403
 
+# Deprecated
+from .util import multihash  # noqa: F401, F403
+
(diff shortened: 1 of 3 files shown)


PhilipVinc (approved):
I would be ok with removing this, but Wessel used it so he might disagree.
I don't think deprecation is needed because this is internal stuff.

nstarman:
It's imported into the main `__init__`. Per [PEP8](https://pep8.org/#public-and-internal-interfaces) it should be public, no?

wesselb:
Technically it's part of the public API, so deprecating is the most correct. I'm not sure if anyone actually uses this function, but, to be sure and since @nstarman has already done the hard work, let's deprecate. :)

Thanks, @nstarman! PR looks super great. :)

Issue #144, opened 2024-04-21
Use `codspeed` for benchmarks


nstarman:
https://codspeed.io.

I'd be interested to see bottlenecks and would probably contribute speed-optimization PRs.

wesselb:
@nstarman Will do! Will report back here when that's enabled.

wesselb:
@leycec, would you be OK with using https://codspeed.io for Plum? I've attempted to add `beartype/plum` to https://codspeed.io, and there should be a request for you to accept or deny. :)

leycec:
**Superbness.** I'll immediately make it so. Tangentially, I've also been meaning to:

* Escalate your **@beartype org** privileges to co-**Admin.** Every repo benefits from multiple Admins, right? What do you say? Wanna be a @beartype org Admin? Then you won't need to ping me anymore. You'll be free to go ham and wreak havoc as if GitHub was your own private demolition derby. This also helps the @beartype community to avoid *force majeure*-style cataclysmic events like Ontario becoming Province Zero for the inevitable [Sophon alien invasion](https://www.polygon.com/24105807/3-body-problem-sophons-protons-explainer). <sup>*...wait. the Sophons are just proteins folded into hyper-dimensional supercomputers? laaame.*</sup>
* Announce a new `beartype/bearden` repo to discuss @beartype org-wide issues like this.

None of that's happened yet. Thrash me in a month if I still haven't done any of that. :face_holding_back_tears:

wesselb:
> What do you say?

Whoa, yes, this would be amazing. :) Of course I will promise to use my very best judgement when exercising co-admin powers! Coincidentally it seems that I can't quite yet configure Codspeed because I'm missing some permissions, so that would solve that too.

> Announce a new beartype/bearden repo to discuss https://github.com/beartype org-wide issues like this.

Aha! Great idea. :) I like it!

leycec:
**Welcome!** Welcome to your new-found superpowers. I am your Bear Father, @leycec. <sup>*NOOOOOOOOOOOO—*</sup>

![](https://media1.tenor.com/m/loBzwA4zEgUAAAAC/black-hell-no.gif)

There are only Two Rules to Bear Club. It's a small club and we're it:

1. **Please avoid promoting anybody else to the status of "Owner" in the @beartype organization.** I trust you implicitly. You are @wesselb, a well-known force of goodness. I don't necessarily trust anyone *else* implicitly, however. Rule One of Bear Club has a little-known Sub-Rule...
[...]

PR #176, opened 2024-06-28
feat: type annotations for dispatcher.multi

nstarman:


nstarman:
I think the cache speed test is flaky. #144 might help.

PhilipVinc (approved):

wesselb:
Thanks again, @nstarman! The cache test definitely is flaky. It's a little annoying.

#144 had falling off my radar! I'll look into that very shortly. :)

PR #177, opened 2024-06-28
feat: update deprecations

nstarman:
After #175 and #173 were / will be merged I took a look and noticed some small improvements.
This PR moves `Val` out of `__all__` and applies the https://peps.python.org/pep-0702/ `deprecated` decorator to make the deprecations visible to static analyzers.

Hm. Maybe I should move the stuff from here to #173 ?
Also, this PR is a good example of why the `edit` button on files in GH is not a great way to make a PR.

### DIFF
--- a/plum/__init__.py
+++ b/plum/__init__.py
@@ -24,6 +24,8 @@ from .type import resolve_type_hint
 from .util import *  # noqa: F401, F403
 
 # Deprecated
+# isort: split
+from .parametric import Val  # noqa: F401, F403
 from .util import multihash  # noqa: F401, F403
 
 # Ensure that type checking is always entirely correct! The default O(1) strategy
--- a/plum/parametric.py
+++ b/plum/parametric.py
@@ -1,6 +1,7 @@
 import contextlib
-import warnings
 from typing import Type, TypeVar, Union
+
+from typing_extensions import deprecated
 
 import beartype.door
 from beartype.roar import BeartypeDoorNonpepException
@@ -19,7 +20,6 @@ __all__ = [
     "type_unparametrized",
     "kind",
     "Kind",
-    "Val",
 ]
 
 T = TypeVar("T")
@@ -632,6 +632,7 @@ def kind(SuperClass=object):
 Kind = kind()  #: A default kind provided for convenience.
 
 
+@deprecated("Use `typing.Literal[val]` instead.")
 @parametric
 class Val:
     """A parametric type used to move information from the value domain to the type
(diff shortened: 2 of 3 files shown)


plum/util.py
```diff
@@ -47,6 +49,7 @@ def __init__(self):
         raise TypeError("`Missing` cannot be instantiated.")
 
 
+@deprecated("Use hash(tuple_of_args) instead.")
```
wesselb, line 52:
```suggestion
@deprecated("Use `hash(tuple_of_args)` instead.")
```

plum/parametric.py
```diff
@@ -631,6 +632,7 @@ def get(self):
 Kind = kind()  #: A default kind provided for convenience.
 
 
+@deprecated("Use typing.Literal[val] instead.")
```
wesselb, line 635:
```suggestion
@deprecated("Use `typing.Literal[val]` instead.")
```

wesselb (commented):
Ah, I like this. Should this replace the `warning.warn` calls?

plum/__init__.py
```diff
@@ -16,14 +16,15 @@
 from .function import *  # noqa: F401, F403
 from .overload import *  # noqa: F401, F403
 from .parametric import *  # noqa: F401, F403
+
+# Deprecated
```
nstarman, line 20: Isort moved that. I'll move it back and add a isort split.

nstarman: Done!

PhilipVinc (approved):
LGTM. I'm happy to discover about `typing_extensions.deprecated`.

By the way @wesselb , a thing that always bothered me about plum is that we have a severely polluted namespace and several modules exposed which are unclear how public they are:

```python
In [1]: import plum

In [2]: plum.[/tab/tab]
  deactivate_union_aliases() dispatcher                 get_context()              isinstance()               List                       multihash()
  Dict                       Function                   get_overloads()            issubclass()               method                     NotFoundLookupError
< dispatch                   function                   is_faithful()              Kind                       Missing                    overload()                 >
  Dispatcher                 get_class()                is_in_class()              kind()                     ModuleType                 parametric()
```

I think one thing that might be helpful down the line would be to move everything to a `_src` module, and expose just few things in the main namespace...

wesselb:
@PhilipVinc Hmmm, you’re totally right. It would be considerable improvement to better structure what’s exported where. I agree that the namespace currently is polluted. I’ve noted this down! Thanks! :)

@nstarman, thanks for yet another super clean PR!! I really do appreciate the many small, very well structured PRs. You’ve contributed a lot of improvements lately. :) This also looks great, so merging right away.

PR #183, opened 2024-07-17
build: update typing-extensions dependency

nstarman:
typing_extensions is used for `deprecation`. This is a runtime requirement for all python versions.


wesselb:
Good spot! Thanks :).

PR #186, opened 2024-07-20
fix: different dispatch instances on same function

nstarman:


wesselb:
Hey @nstarman!

Thanks for creating a PR. :) I'm not entirely sure about this change. For a function, `function._f` corresponds to the method that was given whenever the function was created. Here's an example:

```python
from number import Number


@dispatch.abstract
def add(x: Number, y: Number):
    """Add two numbers.

    Generic docstring.
    """


@dispatch
def add(x: int, y: int):
    return x + y


# At this point, `add._f` is the function given in the abstract implementation.
```

I'm therefore worried that this might result in unexpected behaviour where people expect the attribute `_f` to be something that it isn't.

In general, I think passing a _function_ to `@dispatch` is not correct. Instead, one should use `@dispatch` to define _methods_. Hopefully this makes sense, but what I'm trying to say is that a function is a generic function name, like `add`, and a method is a specific implementation of a function with a type signature, like `add(x: int, y: int)`. In particular, a function does not have a type signature, because it chooses which signature to use whenever the function is called (the process of dispatching). Therefore, giving a function to `@dispatch` wouldn't be right, because there is no well-defined type signature.

Instead, I'm thinking that this instead we should give an error. What would you think about this?

nstarman:
I'm not sure I see what you mean. This PR is about making it possible to do

```python
@dispatch2
@dispatch1
def f(x: int, y: float):
   return x + y
```

where `dispatch1` and `dispatch2` are different instances of `Dispatcher`.

I don't think this approach suffers from the problem you outline. The following should (correct me if I'm wrong) find the correct *method*.

```python
if isinstance(method, Function):
    signature = Signature.from_callable(method._f)
    method, _ = method.resolve_method(signature)
```

wesselb:
@nstarman What would you think of changing the pattern to something like this?

```python
@dispatch1.chain(dispatch2)
def f(x: int, y: float):
    return x + y
```

Then, internally, the `chain` method would simply call `dispatch1(f)` and `dispatch2(f)`, so there wouldn't be any unwrapping required. It's slightly less nice syntax wise, but perhaps acceptable?

nstarman:
Interesting! Or what about having a DispatcherBundle class such that it's


```python

@(dispatch1 | dispatch2)
def func(...): ...

```

Where 

```python 
class AbstractDispatcher: ...

class Dispatcher(AbstractDispatcher):
    def __or__(self, other: Dispatcher) -> DispatcherBundle: ...

class DispatcherBundle(AbstractDispatcher):
    def __call__(...):
        for dispatcher in self:
            ...
    
```

wesselb:
@nstarman Aha, I like that even better, since now the syntax is symmetric and indeed more Pythonic! Would be very happy to go for that. :)

nstarman:
Just need to add a guard for python 3.8 since @(...) syntax is not supported until py3.9+.
As a side note, when were you planning to drop 3.8 support?

PhilipVinc:
If I can share a thought about `DispatcherBundle` and the double registration of functions...

I would caution against including such features into plum itself, and would rather refactor and open up the internals in a reasonable way that makes it easy to build such things on top organically and without fragmenting the ecosystem.

One of the reasons is that multiple dispatch in python is already somewhat 'weird' and not native. Plum is trying to make it possible and reasonable, but there's already a long way to go to get there. While other languages offer great inspiration and carefully crafted ideas or frameworks, we lack a coherent view on how to get a 'reasonable MD that works and is intuitive' in Python.

However, I do not think that adding many features to plum itself is the best way to get there.  What is proposed in this PR is an exotic use case that is intricated.
Already MD is 'hard to understand' to the average python user, and dispatcher bundles seem to me an even more complicated concept **never** seen in another language and that hides intent even more! 
While I understand that it might make sense in your particular project* , I do not think that those untested ideas belong to plum itself, where they might confuse users even more and push them in directions I _personally_ do not think we should endorse.

At the same time, I would like plum to foster experimentation of such 'crazy ideas', because maybe at some point we might land on the 'right approach' to MD in Python. To do that, I think that a reasonable part of plum's internal should be expandable and reusable such that you'd be able to build this feature in your own library, on top of plum.

In short, I think this is a strong idea about how MD should work, going in a step that does not belong to MD (multiple dispatchers for the same functions). 


* I really have never seen such an exotic idea and I think that if it was never attempted, there might be a reason. I would strongly suggest to think twice about such a complicated approach and try to take a few steps back and solve it differently.

wesselb:
@PhilipVinc, you make some solid arguments! @nstarman, what do you think?

In this case, I think that the functionality could be fairly easily implemented with a third-party function resulting in something like this:

```python
@chain_dispatchers(dispatch1, dispatch2)
def f(…):
    …
```

@nstarman, would you also be happy with a solution like this, or do you think this functionality should belong in Plum? If the latter, possibly we could organise more experimental functions that are not to be used by the average user in e.g. `plum.experimental`, unless @PhilipVinc thinks that we should consider doing not even that.

nstarman:
I think @PhilipVinc raises good points!
Without language changes, MD in Python will be implemented by on and by objects of which that can be multiple instances. I agree this is not ideal (and hopefully not too common) and that we don't necessarily want to add functionality to make this more visible.
I'm happy for `chain_dispatchers` to be in my 3rd party code, but IMO a `plum.experimental` would be a better place.

wesselb:
Happy to go with `plum.experimental` if @PhilipVinc is also happy with that!

nstarman:
@PhilipVinc. For me this is to support the multiple different versions of the same function that exist in NumPy and other array libraries during the Array API transition. For example `jax.numpy.arange` and `jax.experimental.array_api` have different signatures and requirements.
See [unxt](https://github.com/GalacticDynamics/unxt/blob/93a833f27d86ab301db1cffefba5d5617f3647f4/src/unxt/_quantity/register_dispatches.py#L35-L55) and [quaxed/array_api](https://github.com/GalacticDynamics/quaxed/blob/main/src/quaxed/array_api/_creation_functions.py) and [quaxed/numpy](https://github.com/GalacticDynamics/quaxed/blob/6de6552250d2ac3b4dca9d68b732825112576916/src/quaxed/numpy/_creation_functions.py#L36)
This specific situation will eventually be resolved once array-api is fully adopted.

wesselb:
@PhilipVinc Multiple dispatchers are required to emulate something like how scoping works in Julia. For example, in Julia, if package A defines a function f but does not export it, then you can define your own function f without the methods defined in A. Only once A exports f become these methods visible in your local scope. With a singleton dispatcher, all methods for function f would be visible to everyone and you could accidentally override methods or extend functions that weren’t supposed to be public.

@nstarman If `chain_dispatchers` is only required temporarily, maybe its better to define it in your package. If you think it’s useful long term, then `plum.experimental` sounds great. :)

PR #192, opened 2024-11-13
Update org name from `wesselb` to `beartype` in docs & errors

gy-mate:
Update GitHub organization name from `wesselb` to `beartype` in docs config file and error messages.


wesselb:
Thank you, @gy-mate! Good catch. :)

PR #193, opened 2024-11-14
feat: see methods from the bound function

nstarman:


wesselb:
@nstarman This looks great! Would you be able to add a quick test that checks the presence of methods in bound functions? Just in case someone accidentally removes or refactors it in the future. :)

nstarman:
Looking at the test suite I'm not sure the best place to put the tests, feel free to move!

wesselb:
@nstarman I think this is perfect! Thank you :)

PR #195, opened 2024-11-14
refactor: use dependency groups

nstarman:


wesselb:
@nstarman That is unfortunate! I wasn’t aware that dependency groups are possible. I love it!

nstarman:
Would you be interested to use `uv` in the CI? https://docs.astral.sh/uv/guides/integration/github/#syncing-and-running
uv is not a build backends, so there's no change re hatch, its just the frontend commands. See https://github.com/GalacticDynamics/unxt for an example where hatch is the backend but uv is used for installing / running stuff.

wesselb:
@nstarman I think so! I haven’t used uv, but it looks great. Would you recommend uv?

nstarman:
Very much so!

wesselb:
Let’s do it then!!

wesselb:
@nstarman Looks great!! Merging right away. :)

PR #196, opened 2024-11-14
refactor: consolidate hatch config

nstarman:


nstarman:
@wesselb, test failure appears unrelated.

wesselb:
@nstarman I agree! Also thanks for this!! :)

PR #202, opened 2025-01-06
types(promotion): add typing to convert

nstarman:

PR #203, opened 2025-01-06
typing: add more type annotations

nstarman:
Requires #202


wesselb:
@nstarman Thanks so much for all your work and effort! This is really great. :) Also merging this right away!

PR #217, opened 2025-09-30
deps: drop py3.8, support py3.9+

nstarman:
It's very old and even new versions of pip don't support it.


nstarman:
2nd commit simplifies a lot of imports!

tests/test_type.py
```diff
@@ -29,6 +29,11 @@ class Literal(metaclass=LiteralMeta):
 )
 from plum.util import Callable
 
+skipifLT310 = pytest.mark.skipif(
```
wesselb:
```suggestion
skip_if_less_than_py310 = pytest.mark.skipif(
```

Slight improvement for readability. :)

tests/advanced/test_annotated.py
wesselb, line 1: Why did you remove this test? I believe `Annotated` still exists.

wesselb (commented):
Yes, this is great! Thanks, @nstarman. :) It's about time Plum drops support for 3.8.  I've left two comments.

Would it perhaps also be possible to replace e.g. `typing.List` by `list` in the code blocks in the docs? This is a bit annoying, so I can also do this myself.

tests/advanced/test_annotated.py
nstarman: Shoot. yes. I saw that it was testing `plum.Anntotated`, but forgot that was a re-export of `typing.Annotated.`

tests/test_type.py
nstarman: Done!

tests/advanced/test_annotated.py
nstarman: Fixed!

tests/test_type.py
wesselb: Amazing! Ty :)

tests/advanced/test_annotated.py
wesselb: Thanks :)

wesselb (approved):
Thanks again, @nstarman! This is so much cleaner. Merging right away. :)

PR #225, opened 2025-11-23
Enable py3.14 and deprecate some alias functionality

### LINKED ISSUE #224: Problems with Union Aliases in Python 3.14
In Python 3.14 `typing.Union` is immutable and the following error is raised

```
.../plum/alias.py:131: in activate_union_aliases
    _union_type.__repr__ = _new_repr
    ^^^^^^^^^^^^^^^^^^^^
E   TypeError: cannot set '__repr__' attribute of immutable type 'typing.Union'
```

### LINKED ISSUE #227: Alternative for Union Aliases in Py3.14+
Right now, plum has the concept of “union aliases” managed via `set_union_alias`, `activate_union_aliases`, and `deactivate_union_aliases`. These are essentially a runtime mechanism to treat a particular `Union[...]` as if it had a stable, named alias for dispatch and introspection, without that alias necessarily appearing in the function’s source-level annotations.

In Python 3.14+ this is broken (#224).

There might be a workable alternative. With Python 3.13’s `typing.TypeAliasType`, we can define custom aliases and swap them for `Union` objects. This will work, however we might lose the ability to have activation/deactivation toggles.

When `plum.dispatch` is called, plum already walks the annotations to build its internal signature representation. At that ingestion point, plum can:
1. Look up each annotated type in the existing union alias registry (populated by `set_union_alias`).
2. If an annotation is a `Union[...]` (or equivalent `X | Y`) that is registered under some alias name `AliasName`, dynamically create a `TypeAliasType` object:
```python
from typing import TypeAliasType

AliasName = TypeAliasType(
    name="AliasName",
    value=original_union_type,  # e.g. int | str
    type_params=(),
)
```
3. Replace the `Union` annotation in the ingested signature with this `TypeAliasType` instance.

From plum’s perspective, the function now genuinely appears to be annotated with a named alias type, even though the original source used `Union[...]` directly. Dispatch and introspection can key off that alias type instead of pattern-matching on the underlying union structure.

This transformation is essentially one-way:
- Once plum has rewritten a `Union[...]` annotation into a `TypeAliasType("AliasName", union_value, ())`, the stored signature no longer contains the original raw Union anywhere; it only knows about the alias.
- The alias is still semantically tied to the union via `AliasName.__value__`, but “going back” to a state where the signature refers to `Union[...]` instead of the alias would require explicitly undoing the rewrite and re-injecting the bare union. That might be possible, but it's not guaranteed to be a safe reverse mapping.

nstarman:
Aliases are now always turned on. The alias machinery dynamically creates `TypeAliasType` objects that are ONLY used in string representations, never in the actual type. 

Fixes #224
Fixes #227


wesselb:
> Meanwhile can I make this a no-op for Python 3.14+?

Absolutely. Let's do this for now until we figure out whether this is still possible for 3.14+.

nstarman:
Fully functional until py3.14 then they turn into deprecated no-ops.

nstarman:
> Thus, I ask you: "In 2026, can't Plum users just use type aliases directly? Are Plum-specific union aliases even desired or required anymore?" That is to say:

This PR deprecates `activate_union_aliases`, but keeps `set_union_alias`, using dynamically generated PEP 695 type aliases to enable (in repr only, not during type-checking) the alias name to be displayed instead of the union. If one has control over the creation of a function, yes, direct use of type aliases are to be preferred:

```python
scalar_types = tuple(np.sctypeDict.values())  # All NumPy scalar types

type Scalar = Union[scalar_types]

@dispatch
def add(x: Scalar, y: Scalar):
    return x + y
```

But these aliases are "local" to the context they are used. With `set_union_alias` we can make it so that **any** type annotations that `plum` displays will use the alias, which is especially useful for functions we do *not* control:

```python
# imaginary_library (I don't control)

two_elt_dtype = np.dtype(...)

def add(x: Fraction | tuple[float | np.floating, float | np.floating] | two_elt_dtype, y: Fraction | tuple[float | np.floating, float | np.floating] | two_elt_dtype) -> Fraction:
    return x + y


# my_library (I do control)

plum.dispatch(imaginary_library.add)  # register in their method

type Scalar = Union[scalar_types]  # love it

@plum.dispatch
def add(x: Scalar, y: Scalar) -> Scalar:  # register in our method
    return x + y

# IDE
>>> add.methods
List of X method(s):
    [0] add(x: Scalar, y: Scalar] -> Scalar: ...
    [1] add(x: Fraction | tuple[float | np.floating, float | np.floating] | two_elt_dtype, y: Fraction | tuple[float | np.floating, float | np.floating] | two_elt_dtype) -> Fraction: ...
    
>>> plum.set_union_alias(Fraction | tuple[float | np.floating, float | np.floating] | two_elt_dtype, "FractionLike")
>>> add.methods
List of X method(s):
    [0] add(x: Scalar, y: Scalar] -> Scalar: ...
    [1] add(x: FractionLike, y: FractionLike) -> Fraction: ...
```

`plum.set_union_alias` makes Boim's life easier when examining plum dispatches for methods he didn't define (without actually interfering with those methods). In `beartype`, I'm suggesting that the diagnostic printouts might similarly benefit — grossly long type annotations can be condensed to a convenient alias without actually mutating the code.

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+    if isinstance(x, TypeAliasType):
+        return x
+
+    # Get the union args to check if it's registered
+    args = get_args(x) if isinstance(x, _union_type) else None
+    if args:
+        args_set = set(args)
+        # Look for a matching alias in the registry
+        for union_args, type_alias in _ALIASED_UNIONS.items():
+            if set(union_args) == args_set:
```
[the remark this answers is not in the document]
nstarman: I'm not sure this is correct. The tests all pass. 
This is a subtle one.
Ping @wesselb

docs/union_aliases.md
```diff
@@ -41,6 +41,22 @@ by aliasing a union, you change the way it is displayed.
...
 
+% invisible-code-block: python
+%
+% import sys
+
+% skip: start if(sys.version_info < (3, 14), reason="Union repr changed in Python 3.14+")
+
+```python
+>>> from plum import set_union_alias
+>>> set_union_alias(Scalar, alias="Scalar")
```
wesselb:
```suggestion
>>> from plum import set_union_alias

>>> set_union_alias(Scalar, alias="Scalar")
```

docs/union_aliases.md
```diff
@@ -41,6 +41,22 @@ by aliasing a union, you change the way it is displayed.
 Union aliases must be activated explicitly, because the feature
 monkeypatches `Union.__str__` and `Union.__repr__`.
```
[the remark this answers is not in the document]
wesselb: This suggestion is pretty good, @nstarman. Shall we write something along these lines?

docs/union_aliases.md
```diff
@@ -98,9 +142,12 @@ typing.Union[numpy.int8, numpy.int16, numpy.int32, numpy.longlong, numpy.int64,
 You can deactivate union aliases with `deactivate_union_aliases`:
 
 ```python
+>>> import warnings
 >>> from plum import deactivate_union_aliases
```
wesselb:
```suggestion
>>> import warnings

>>> from plum import deactivate_union_aliases
```

docs/union_aliases.md
```diff
@@ -98,9 +142,12 @@ typing.Union[numpy.int8, numpy.int16, numpy.int32, numpy.longlong, numpy.int64,
...
 
->>> deactivate_union_aliases()
+>>> with warnings.catch_warnings():
+...     warnings.simplefilter("ignore")
+...     deactivate_union_aliases()
```
wesselb:
```suggestion
>>> deactivate_union_aliases()
```

Which warning are you catching with this? I'm locally not seeing any warnings. Would it be possible to omit this to keep it simpler in the docs?

docs/union_aliases.md
```diff
@@ -41,6 +41,22 @@ by aliasing a union, you change the way it is displayed.
...
+
+```python
+>>> from plum import set_union_alias
+>>> set_union_alias(Scalar, alias="Scalar")
+numpy.bool | numpy.float16 | ...
+```
+
+% skip: end
+
+% skip: start if(sys.version_info >= (3, 14), reason="Union repr changed in Python 3.14+")
```
wesselb:
```suggestion
% skip: start if(sys.version_info >= (3, 14), reason="Representation of unions changed in Python 3.14.")
```

docs/union_aliases.md
```diff
@@ -68,6 +86,30 @@ For example, printing just `Scalar` would omit the type parameter(s).
 
 Let's see with a few more examples how this works:
 
+% invisible-code-block: python
+%
+% import sys
+
+% skip: start if(sys.version_info < (3, 14), reason="Union repr changed in Python 3.14+")
```
wesselb:
```suggestion
% skip: start if(sys.version_info < (3, 14), reason="Representation of unions changed in Python 3.14.")
```

docs/union_aliases.md
```diff
@@ -68,6 +86,30 @@ For example, printing just `Scalar` would omit the type parameter(s).
...
+>>> Union[tuple(scalar_types) + (tuple,)]       # Scalar or tuple
+numpy.bool | numpy.float16 | ... | tuple
+
+>>> Union[tuple(scalar_types) + (tuple, list)]  # Scalar or tuple or list
+numpy.bool | numpy.float16 | ... | tuple | list
+```
+
+% skip: end
+
+% skip: start if(sys.version_info >= (3, 14), reason="Union repr changed in Python 3.14+")
```
wesselb:
```suggestion
% skip: start if(sys.version_info >= (3, 14), reason="Representation of unions changed in Python 3.14.")
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+
+        When used with plum's dispatch system, the union will be automatically
+        transformed into a `TypeAliasType` during signature extraction, allowing
+        dispatch to key off the alias name instead of the union structure.
+
+        Args:
+            union (type or type hint): A union type or a single type.
+            alias (str): Alias name for the union.
+
+        """
```
wesselb:
```suggestion
        """Register a union alias for use in Plum's dispatch system.

        When used with Plum's dispatch system, the union will be automatically
        transformed into a `TypeAliasType` during signature extraction, allowing
        dispatch to key off the alias name instead of the union structure.

        Args:
            union (type or type hint): A union type or a single type.
            alias (str): Alias name for the union.

        Returns:
            type or type hint: The given union.
        """
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+        # Handle both union types and single types, matching < 3.14 behaviour.
```
wesselb:
```suggestion
        # Handle both union types and single types, matching pre-3.14 behaviour.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+        args = get_args(union) if isinstance(union, _union_type) else (union,)
+
+        # Check for conflicting aliases
```
wesselb:
```suggestion
        # Check for conflicting aliases.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
 
-_ALIASED_UNIONS: list[tuple[tuple[Any, ...], str]] = []
+        return union
 
 
-def set_union_alias(union: UnionT, alias: str) -> UnionT:
-    """Change how a `typing.Union` is printed. This does not modify `union`.
+def _transform_union_alias(x: object, /) -> object:
+    """Transform a Union type hint to a TypeAliasType if it's registered in the alias
+    registry. This is used by plum's dispatch machinery to use aliased names for unions.
```
wesselb:
```suggestion
    """Transform a union type hint to a `TypeAliasType` if it's registered in the alias
    registry. This is used by Plum's dispatch machinery to use aliased names for unions.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
 
     Args:
-        union (type or type hint): A union.
-        alias (str): How to print `union`.
+        x (type or type hint): Type hint, potentially a Union.
```
wesselb:
```suggestion
        x (type or type hint): Type hint, potentially a union.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
 
     Returns:
-        type or type hint: `union`.
+        type or type hint: If `x` is a Union registered in `_ALIASED_UNIONS`, returns
+            the TypeAliasType. Otherwise returns `x` unchanged.
```
wesselb:
```suggestion
        type or type hint: If `x` is a union registered in `_ALIASED_UNIONS`, returns
            the `TypeAliasType`. Otherwise returns `x` unchanged.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
-            else:
-                union_str = repr(union)
-            raise RuntimeError(f"`{union_str}` already has alias `{existing_alias}`.")
-    _ALIASED_UNIONS.append((args, alias))
-    return union
+    # Fast path: if aliases are not active, return `x` immediately.
+    if not _ALIASES_ARE_ACTIVE:
+        return x
+
+    # TypeAliasType instances are already transformed, return as-is
```
wesselb:
```suggestion
    # `TypeAliasType` instances are already transformed, so return as-is.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+    if isinstance(x, TypeAliasType):
+        return x
+
+    # Get the union args to check if it's registered
```
wesselb:
```suggestion
    # Get the union args to check if it's registered.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+    args = get_args(x) if isinstance(x, _union_type) else None
+    if args:
+        args_set = set(args)
+        # Look for a matching alias in the registry
```
wesselb:
```suggestion
        # Look for a matching alias in the registry.
```

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+        for union_args, type_alias in _ALIASED_UNIONS.items():
+            if set(union_args) == args_set:
+                return type_alias
+
+    # Not a union or not aliased, return as-is
```
wesselb:
```suggestion
    # Not a union or not aliased, so return as-is.
```

src/plum/repr.py
```diff
@@ -60,14 +65,20 @@ def repr_short(x: object, /) -> str:
     """Representation as a string, but in shorter form. This just calls
     :func:`typing._type_repr`.
 
+    If the type is a union registered in plum's alias registry, the alias name
+    is used instead.
```
wesselb:
```suggestion
    If the type is a union registered in Plum's alias registry, the alias name
    is used instead.
```

src/plum/repr.py
```diff
@@ -60,14 +65,20 @@ def repr_short(x: object, /) -> str:
...
     Args:
         x (object): Object.
 
     Returns:
         str: Shorter representation of `x`.
     """
-    # :func:`typing._type_repr` is an internal function, but it should be available in
-    # Python versions 3.9 through 3.13.
+    if isinstance(transformed := _transform_union_alias(x), TypeAliasType):
+        # It's an aliased union — use the alias name
```
wesselb:
```suggestion
        # It's an aliased union — use the alias name.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+import beartype
+import beartype.door
+
+import plum
+
+# These tests are for Python >= 3.14 only.
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (3, 14),
+    reason="Union aliasing tests for Python >= 3.14",
+)
```
wesselb:
```suggestion
# These tests are for Python 3.14 and later only.
pytestmark = pytest.mark.skipif(
    sys.version_info < (3, 14),
    reason="Union aliasing tests for Python 3.14 and later.",
)
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def test_activate_union_aliases() -> None:
+    """Test that activate_union_aliases sets _ALIASES_ARE_ACTIVE to True."""
```
wesselb:
```suggestion
    """Test that `activate_union_aliases` sets `_ALIASES_ARE_ACTIVE` to `True`."""
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum._alias._ALIASES_ARE_ACTIVE = False
+    plum.activate_union_aliases()
+    assert plum._alias._ALIASES_ARE_ACTIVE is True
+
+
+def test_deactivate_union_aliases() -> None:
+    """Test that deactivate_union_aliases sets _ALIASES_ARE_ACTIVE to False."""
```
wesselb:
```suggestion
    """Test that `deactivate_union_aliases` sets `_ALIASES_ARE_ACTIVE` to `False`."""
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum._alias._ALIASES_ARE_ACTIVE = True
+    plum.deactivate_union_aliases()
+    assert plum._alias._ALIASES_ARE_ACTIVE is False
+
+
+@pytest.mark.parametrize("union", [int | str, Union[int, str]])  # noqa: UP007
+def test_repr_short_uses_alias(union) -> None:
+    """Test that repr_short substitutes registered union aliases."""
```
wesselb:
```suggestion
    """Test that `repr_short` substitutes registered union aliases."""
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum.set_union_alias(union, alias="IntStr")
+
+    # Aliased union should use the alias name
```
wesselb:
```suggestion
    # Aliased union should use the alias name.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    assert plum.repr.repr_short(int | str) == "IntStr"
+    assert plum.repr.repr_short(Union[int, str]) == "IntStr"  # noqa: UP007
+
+    # Non-aliased unions should be unchanged
```
wesselb:
```suggestion
    # Non-aliased unions should be unchanged.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    assert "IntStr" not in plum.repr.repr_short(int | float)
+    assert "IntStr" not in plum.repr.repr_short(Union[int, float])  # noqa: UP007
+
+    # Plain types should be unchanged
```
wesselb:
```suggestion
    # Plain types should be unchanged.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    assert plum.repr.repr_short(int) == "int"
+    assert plum.repr.repr_short(float) == "float"
+
+    # Signature printing should use the alias
```
wesselb:
```suggestion
    # Signature printing should use the alias.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    sig = plum.Signature(int | str, float)
+    assert "IntStr" in repr(sig)
+    assert repr(sig) == "Signature(IntStr, float)"
+
+    # Signature printing should use the alias
```
wesselb:
```suggestion
    # Signature printing should use the alias.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum.set_union_alias(int | str | float, alias="IntStrFloat")  # noqa: UP007
+    assert display(int | str | float) == "IntStrFloat"  # noqa: UP007
+
+
+def test_repr_short_with_type_alias_type_passthrough():
+    """Test that repr_short handles a TypeAliasType passed directly (not as a union).
+
+    This exercises the early-return path in _transform_union_alias where the
+    input is already a TypeAliasType instance.
+    """
```
wesselb:
```suggestion
    """Test that `repr_short` handles a `TypeAliasType` passed directly (not as a union).

    This exercises the early-return path in `_transform_union_alias` where the
    input is already a `TypeAliasType` instance.
    """
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+def test_double_registration():
+    # We can register with the same alias, but not with a different alias.
+    plum.set_union_alias(Union[int, str], alias="IntStr")  # noqa: UP007
+    with pytest.raises(RuntimeError, match=r"already has alias"):
+        plum.set_union_alias(Union[int, str], alias="OtherIntStr")  # noqa: UP007
+    with pytest.raises(RuntimeError, match=r"already has alias"):
+        plum.set_union_alias(int | str, alias="OtherIntStr")
+
+    # The same applies for Union types.
```
wesselb:
```suggestion
    # The same applies for union types.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+    # We can also register plain types, but the same rules apply.
+    plum.set_union_alias(int, alias="MyInt")
+    plum.set_union_alias(int, alias="MyInt")  # This is OK.
+    with pytest.raises(RuntimeError, match=r"already has alias"):
+        plum.set_union_alias(int, alias="MyOtherInt")
+
+
+def test_set_union_alias_generated_type_alias() -> None:
+    """Test that set_union_alias generates a TypeAliasType for unions."""
```
wesselb:
```suggestion
    """Test that `set_union_alias` generates a `TypeAliasType` for unions."""
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum.set_union_alias(Union[int, str], alias="IntStr")  # noqa: UP007
+
+    IntStr = plum._alias._ALIASED_UNIONS[(int, str)]
+
+    # The returned value should be a TypeAliasType
```
wesselb:
```suggestion
    # The returned value should be a `TypeAliasType`.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    assert hasattr(IntStr, "__name__")
+    assert IntStr.__name__ == "IntStr"
+    assert hasattr(IntStr, "__value__")
+    # The underlying value should be the union
```
wesselb:
```suggestion
    # The underlying value should be the union.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    from typing import get_args
+
+    assert set(get_args(IntStr.__value__)) == {int, str}
+
+
+def test_dispatch_with_union_alias(dispatch: plum.Dispatcher) -> None:
+    """Test that dispatch works correctly with union aliases."""
+    # Register an alias for Union[int, str]
```
wesselb:
```suggestion
    # Register an alias for `Union[int, str]`.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum.set_union_alias(Union[int, str], alias="IntStr")  # noqa: UP007
+
+    # Define a function using the alias
```
wesselb:
```suggestion
    # Define a function using the alias.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    @dispatch
+    def process(x: int | str) -> str:
+        return "int or str"
+
+    @dispatch
+    def process(x: float) -> str:
+        return "float"
+
+    # Test dispatch
```
wesselb:
```suggestion
    # Test dispatch.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    assert process(42) == "int or str"
+    assert process("hello") == "int or str"
+    assert process(3.14) == "float"
```
wesselb:
```suggestion
    """Test that dispatch works when using union directly if registered."""
    # Register the alias.
    plum.set_union_alias(Union[int, str], alias="IntStr")  # noqa: UP007

    # Define a function using union directly (should still work).
    @dispatch
    def process(x: int | str) -> str:
        return "int or str"

    @dispatch
    def process(x: float) -> str:
        return "float"

    # Test dispatch.
    assert process(42) == "int or str"
    assert process("hello") == "int or str"
    assert process(3.14) == "float"
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    plum.set_union_alias(int | str, alias="IntStr")
+
+    @dispatch
+    def example(x: int | str, y: float) -> str:
+        return "test"
+
+    # Check that the signature contains the alias name
+    # The signature should show "IntStr" rather than "Union[int, str]"
+    sig_str = str(example.methods[0].signature)
+    assert "IntStr" in sig_str
```
wesselb:
```suggestion
    """Test that function signatures are nicely printed with `TypeAliasType` names."""
    # Register an alias.
    plum.set_union_alias(int | str, alias="IntStr")

    @dispatch
    def example(x: int | str, y: float) -> str:
        return "test"

    # Check that the signature contains the alias name. The signature should show
    # "IntStr" rather than "Union[int, str]".
    sig_str = str(example.methods[0].signature)
    assert "IntStr" in sig_str
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def test_beartype_strict_mode_compatibility(dispatch: plum.Dispatcher) -> None:
+    """Test that strict beartype works with plum dispatch on aliased unions."""
```
wesselb:
```suggestion
    """Test that strict `beartype` works with Plum dispatch on aliased unions."""
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    original_is_bearable = plum._is_bearable
+
+    # Temporarily set strict mode for this test
```
wesselb:
```suggestion
    # Temporarily set strict mode for this test.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+        # These should work
+        assert strict_process(42) == "processed: 42"
+        assert strict_process("hello") == "processed: hello"
+
+        # This should not match the signature (float is not in IntStr)
+        with pytest.raises(plum.NotFoundLookupError):
+            strict_process(3.14)
+    finally:
+        # Restore original
+        plum._is_bearable = original_is_bearable
```
wesselb:
```suggestion
        # Register an alias.
        plum.set_union_alias(Union[int, str], alias="IntStr")  # noqa: UP007

        # Define a function using the alias with Plum dispatch.
        @dispatch
        def strict_process(x: int | str) -> str:
            return f"processed: {x}"

        # These should work.
        assert strict_process(42) == "processed: 42"
        assert strict_process("hello") == "processed: hello"

        # This should not match the signature (`float` is not in `IntStr`).
        with pytest.raises(plum.NotFoundLookupError):
            strict_process(3.14)
    finally:
        # Restore original.
        plum._is_bearable = original_is_bearable
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def test_multiple_aliases_in_signature(dispatch: plum.Dispatcher) -> None:
+    """Test that multiple aliased unions in the same signature work correctly."""
+    # Register multiple aliases
```
wesselb:
```suggestion
    # Register multiple aliases.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+    @dispatch
+    def method(x: int | str) -> str:
+        return f"Integer: {x}"
+
+    assert "IntOrStr" in repr(method.methods)
+
+
+def test_alias_priority_in_dispatch(dispatch: plum.Dispatcher) -> None:
+    """Test that aliased unions are treated like a union in dispatch."""
+    # Register an alias
```
wesselb:
```suggestion
    # Register an alias.
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+    @dispatch
+    def handle(x: int | str) -> str:
+        return "alias"
+
+    @dispatch
+    def handle(x: int) -> str:
+        return "int"
+
+    # The more specific 'int' should match first
```
wesselb:
```suggestion
    # The more specific `int` should match first.
```

tests/test_alias_upto313.py
```diff
@@ -1,18 +1,28 @@
...
 from plum._alias import _ALIASED_UNIONS
 
+# These tests are for Python <= 3.13 only.
+pytestmark = [
+    pytest.mark.skipif(
+        sys.version_info >= (3, 14),
+        reason="Union aliasing tests for Python <= 3.13",
+    ),
+]
+
```
wesselb:
```suggestion
# These tests are for Python 3.13 and earlier only.
pytestmark = pytest.mark.skipif(
    sys.version_info >= (3, 14),
    reason="Union aliasing tests for Python 3.13 and earlier.",
)
```

wesselb (commented):
This is great, @nstarman! Thanks so much for putting this together. I've left an array of minor suggestions to keep the style in sync with the rest of the code. I hope you don't mind.

If I understand correctly, the algorithm for `TypeAliasType` checks if a union occurs in `_ALIASED_UNIONS` and replaces it with the alias if there is a match. This works if the union is exactly contained in `_ALIASED_UNIONS`, but doesn't if it is a union of two alias unions or if there are other types in the union too.

For example, `(str, Scalar)` is expanded to `(str, float, int, Number, ...)`, which would ideally be aliased back to `(str, Scalar)`. The case of the if-statement for before Python 3.14 has some logic to handle the general case. Do you think it would be possible to do this for the Python 3.14 and later case too?

I'm happy to jump in and hack around myself to see if I can make this work. It seems like the logic to determine which aliases are contained in a union could possibly be shared.

src/plum/_alias.py
```diff
@@ -32,135 +32,209 @@
...
+        """Register a union alias for use in plum's dispatch system.
+
+        When used with plum's dispatch system, the union will be automatically
+        transformed into a `TypeAliasType` during signature extraction, allowing
+        dispatch to key off the alias name instead of the union structure.
+
+        Args:
+            union (type or type hint): A union type or a single type.
+            alias (str): Alias name for the union.
+
```
nstarman:
```suggestion
        """Register a union alias for use in plum's signature representations.

        This does NOT alter the actual annotations, only their string representations.

        Args:
            union (type or type hint): A union type or a single type.
            alias (str): Alias name for the union.

```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
+import functools as ft
+import sys
+from typing import Union
```
nstarman:
```suggestion
from typing import TypeAliasType, Union
```

tests/test_alias_314plus.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def test_repr_short_with_type_alias_type_passthrough():
+    """Test that repr_short handles a TypeAliasType passed directly (not as a union).
+
+    This exercises the early-return path in _transform_union_alias where the
+    input is already a TypeAliasType instance.
+    """
+    from typing_extensions import TypeAliasType
+
```
nstarman:
```suggestion
```

docs/union_aliases.md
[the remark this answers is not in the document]
nstarman: Done!

src/plum/_alias.py
nstarman: Done!

docs/union_aliases.md
nstarman: Done!

docs/union_aliases.md
nstarman: Done!

docs/union_aliases.md
nstarman: Done!

docs/union_aliases.md
nstarman: Done!

nstarman:
Ping @wesselb  :)

nstarman:
SGTM. It should only make repr changes, not actual dispatch changes, so most people won't notice.

nstarman:
@wesselb I can't merge this in.

nstarman:
Ping @wesselb to merge and release :)

wesselb (approved):
@nstarman Thanks for the reminder once more! Merging in now. :)

wesselb:
Released as `v2.9.0`! Thanks again @nstarman for enabling union aliases for Python 3.14 and later. This is really great. :) (And so sorry for having been so slow throughout all this. It's been unusually busy.)

PR #268, opened 2026-05-08
More typing

nstarman:
Improves the typing.
One API change is that `parametric` can't take None anymore. It couldn't before, but now it really can't.

### DIFF
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -505,6 +505,30 @@ class _DispatchFunction(Protocol):
     ) -> Self | Callable[[Callable[..., Any]], Self]: ...
 
 
+class _BoundFunctionProto(Protocol):
+    """Subset of :class:`Function`'s interface required by :class:`_BoundFunction`.
+
+    Declaring ``_BoundFunction._f`` with this Protocol rather than :class:`Function`
+    directly prevents mypy from applying ``Function.__get__``'s descriptor protocol
+    when resolving instance-attribute accesses of ``_f``.
+    """
+
+    _f: Callable[..., Any]
+
+    def __call__(self, *args: object, **kw: object) -> object: ...
+
+    def invoke(self, *types: TypeHint) -> Callable[..., Any]: ...
+
+    @property
+    def methods(self) -> MethodList: ...
+
+    def dispatch(
+        self,
+        method: Callable[..., Any] | None = None,
+        precedence: int = 0,
+    ) -> Any: ...
+
+
 class _BoundFunction:
     """A bound instance of `.function.Function`.
 
@@ -513,7 +537,7 @@ class _BoundFunction:
         instance (object): Instance to which the function is bound.
     """
 
-    _f: "Function"
+    _f: "_BoundFunctionProto"
     _instance: object
 
     def __init__(self, f: "Function", instance: object) -> None:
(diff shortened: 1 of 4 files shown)


src/plum/_function.py
```diff
@@ -499,6 +505,30 @@ def __call__(
     ) -> Self | Callable[[Callable[..., Any]], Self]: ...
 
 
+class _BoundFunctionProto(Protocol):
+    """Subset of :class:`Function`'s interface required by :class:`_BoundFunction`.
+
+    Declaring ``_BoundFunction._f`` with this Protocol rather than :class:`Function`
+    directly prevents mypy from applying ``Function.__get__``'s descriptor protocol
+    when resolving instance-attribute accesses of ``_f``.
```
wesselb:
```suggestion
    Declaring `_BoundFunction._f` with this Protocol rather than :class:`Function`
    directly prevents `mypy` from applying `Function.__get__`'s descriptor protocol
    when resolving instance-attribute accesses of `_f`.
```

src/plum/_parametric.py
```diff
@@ -200,37 +203,38 @@ def __le_type_parameter__(cls, left, right) -> bool:
...
         # `isinstance(A[SubType](), A[Type])`. `CovariantMeta` comes first in the MRO,
         # but the implementation of `__instancecheck__` should be taken from
         # `original_meta` if it exists. The implementation of `CovariantMeta` should be
         # used as a fallback. Note that `original_meta.__instancecheck__` always exists.
         # We check that it is not equal to the default `type.__instancecheck__`.
         if original_meta.__instancecheck__ != type.__instancecheck__:
-            return original_meta.__instancecheck__(cls, instance)
+            return cast(
+                Callable[[type, object], bool], original_meta.__instancecheck__
+            )(cls, instance)
```
wesselb:
```suggestion
            target_type = Callable[[type, object], bool]
            return cast(target_type, original_meta.__instancecheck__)(cls, instance)
```

Saves a line!

src/plum/_function.py
```diff
@@ -507,7 +537,7 @@ class _BoundFunction:
         instance (object): Instance to which the function is bound.
     """
 
-    _f: "Function"
+    _f: "_BoundFunctionProto"
```
wesselb, line 540: Ah, nice fix! Would another idea to make this work be to remove the class-level annotations? `mypy` might not use the descriptor protocol then. I mean this:

```python
class _BoundFunction:
    ...
    
    # Nothing here.

    def __init__(self, f: Function, instance: object) -> None:
        self._f: Function = f
        wraps(f._f)(self)  # This will call the setter for `__doc__`.
        self._instance: object = instance
```

src/plum/_parametric.py
```diff
@@ -603,25 +609,25 @@ def type_unparametrized(q: T, /) -> type[T]:
     parameter(s).
     """
     typ = type(q)
-    return q.__class_unparametrized__() if isinstance(typ, ParametricTypeMeta) else typ
+    return q.__class_unparametrized__() if isinstance(typ, ParametricTypeMeta) else typ  # type: ignore[redundant-expr]
```
wesselb, line 610: Is the ignore here necessary?

wesselb (approved):
Looks great!! Left a few questions. Besides the questions, happy to merge whenever you're happy too :)

src/plum/_parametric.py
nstarman: Unfortunately Mypy seems to think so. I'd love to try to refactor at some later PR, but I think getting mypy and therefore mypyc happy will take some larger refactoring than I'm aiming for in this PR.

src/plum/_function.py
nstarman: It's probably better to keep these. checks key off these more reliably than when they try to introspect `__init__`.

src/plum/_parametric.py
wesselb: Make sense!

src/plum/_function.py
wesselb: Sounds very reasonable!

nstarman:
rebasing to fix merge conflicts

PR #269, opened 2026-05-08
Remove deprecated `Val` class

nstarman:
Removes `plum.Val` and its tests, which was deprecated in v0.5.0 in favor of `typing.Literal`. Also updates docs to drop the `Val` reference.


wesselb (approved):
Thanks, @nstarman! Very sensible. Merging right away :)

PR #281, opened 2026-07-13
Make concurrent dispatch resolution thread-safe (#274)

### LINKED ISSUE #274: `@dispatch` first-call resolution causes intermittent runtime errors in multi-threaded code
## Summary

When two threads simultaneously make the **first** dispatch call on a `@dispatch` Function, `_resolve_pending_registrations` races on the same function's `__annotations__` dict and produces intermittent runtime errors. In production this surfaces as flaky CI failures: the errors are rare (~0.2% per `@dispatch` Function on first use under CPython's default GIL switch interval), but they are real and reproducible.

> **Note on provenance:** The root-cause analysis below was produced with AI assistance. I, @mbbush (a bona-fide human developer) read and ran the reproducer, and  confirmed that the bug is real and reproducible on the versions listed.

## Reproducer

A standalone reproducer is available at: **https://github.com/mbbush/plum-dispatch-race-repro**

## Versions

- `plum-dispatch 2.9.0`
- `beartype 0.22.9`
- Reproduced on CPython 3.13.7, 3.13.13, 3.14.0, and 3.14.5

## What Claude thinks is happening

When a `@dispatch` Function is first called, `__call__` invokes `_resolve_pending_registrations`, which iterates `self._pending` and calls `Signature.from_callable(f, ...)` for each registered implementation. That call chain reaches `beartype.peps.resolve_pep563(f)`, which mutates `f.__annotations__` in place — reading the original string annotations, resolving each one, and writing the resulting type objects back.

There is **no lock** around any of this. When two threads race the first call:

```
Thread A                                   Thread B
─────────────────────────────────────────────────────────
_resolve_pending_registrations()           _resolve_pending_registrations()
  for f in self._pending: ...                for f in self._pending: ...   ← same list
    Signature.from_callable(f)               Signature.from_callable(f)   ← same f
      resolve_pep563(f)                        resolve_pep563(f)
        # iterates f.__annotations__             # mutates f.__annotations__
        #          ← RACE →
```

Depending on the exact interleave, four distinct errors can result:

| Symptom | Cause |
|---|---|
| `RuntimeError: dictionary changed size during iteration` | Thread A iterates `f.__annotations__` while Thread B adds/removes keys |
| `AssertionError: <hint> not stringified type hint.` | Thread A resolves an annotation to a type object; Thread B's copy of the same annotation is now a type rather than a string when `resolve_hint` asserts it must be a string |
| `NameError: name '__beartype_object_<id>' is not defined` | Beartype's `is_bearable` checker function body and parameter list are generated from a `func_scope` dict that was concurrently mutated, leaving them referencing different type-object IDs |
| `plum.NotFoundLookupError` | A partially-completed resolve leaves the dispatch table in a corrupt state; a subsequent call with valid args finds no match |

All four share a single root cause: `_resolve_pending_registrations`
performs a multi-step, stateful mutation of each registered function's
annotations with no concurrency protection.

## Potential fix
Claude seems to think you can "just use a lock" to fix this, but I don't know if it's actually that simple. I imagine you'd at least want to look at the performance tradeoffs. I removed most of the AI's "I'm going to mansplain to the maintainer of the library how their own code works" suggested fixes. 

## Consumer-side workaround for users who don't understand any of this but just want their tests to pass consistently

Force resolution at import time (under CPython's single-threaded import lock) immediately after declaring all overloads:

```python
@dispatch
def my_fn(x: int) -> int: ...

@dispatch
def my_fn(x: str) -> str: ...

my_fn._resolve_pending_registrations()  # pre-warms; safe at import time
```

Worker threads then hit a fully-resolved dispatch table and never trigger the lazy resolution path.

nstarman:
## Summary

Refs #274. This closes the **resolution** half of that issue; the remaining half lives upstream in beartype (see Scope below), so #274 should stay open after this merges.

When two threads race the **first** dispatch/resolution of a `@dispatch` `Function`, `_resolve_pending_registrations` concurrently drives beartype's `resolve_pep563`, which mutates the same function's `__annotations__` in place with no synchronization — producing intermittent `AssertionError: ... not stringified type hint`, `RuntimeError: dictionary changed size during iteration`, or `NameError`.

## Changes

- Add a per-`Function` `threading.RLock` (created in `__init__`).
- Guard `_resolve_pending_registrations` with **double-checked locking**: the outer `if not self._pending` check stays unlocked so the resolved hot dispatch path pays no lock cost; a second check inside the lock ensures resolution runs exactly once.
- Guard `clear_cache` with the same lock (its `reregister` branch swaps `_pending`/`_resolved`/`_resolver` in multiple steps). `RLock` is used because `_resolve_pending_registrations` calls `clear_cache`, which reacquires it.
- Add `test_concurrent_resolution_is_thread_safe`, built on a factory fixture that hands out a fresh, unresolved `Function` per iteration. It has to be a factory: the race only occurs on a `Function`'s *first* resolution, so reusing one `Dispatcher` would exercise re-registration instead.

## Verification

| | result |
|---|---|
| New test with `_function.py` reverted to master | **20/20 runs fail** |
| New test with the lock | **20/20 runs pass** |
| Cached hot-path dispatch (`f(1, 2)`, resolved) | **418 ns vs 418 ns on master** — unchanged |

The failure without the lock is the error from #274:

```
AssertionError: [AssertionError("<class 'int'> not stringified type hint."),
                 AssertionError("<class 'str'> not stringified type hint."), ...]
```

The unlocked fast path means the hot dispatch path is genuinely untouched: `_resolve_method_with_cache` already guarded its call on `if self._pending`, so a resolved function never enters the locked region at all.

## Scope: what this does *not* fix

A second, independent race lives in beartype's lazy type-checker **compilation**, reached via `is_bearable` during dispatch *matching*. It corrupts beartype's **process-global** state, so no per-`Function` lock can address it. This is tracked upstream as [beartype/beartype#670](https://github.com/beartype/beartype/issues/670), whose reported symptoms match exactly.

Concretely, with this PR applied, the full concurrent-first-*dispatch* reproducer from #274 still fails **40/40 trials**, bottoming out in beartype rather than plum:

```
File "beartype/_check/signature/sigmake.py", line 98, in make_func_signature
    for arg_name, arg_value in func_scope.items():
RuntimeError: dictionary changed size during iteration
```

Hence `Refs #274` rather than `Fixes #274` — merging this should not auto-close the issue.

## Known limits of the guarantee

"Thread-safe" here means specifically: *first resolution is serialised, under the GIL.* It is not a general thread-safety guarantee for `Function`.

- `register()` still appends to `_pending` outside the lock, and resolution ends with `self._pending = []` (a rebind, not a `.clear()`). In principle a concurrent `register()` could load the old list reference and append to an orphaned list. I stress-tested this — 2400 concurrent registrations against 4 resolver threads at `switchinterval=1e-7` — and lost **zero** methods; the window is a couple of bytecodes and the GIL closes it. So this is safe in practice, but by accident of the GIL rather than by design.
- `_cache` reads and writes remain unlocked, relying on CPython dict operations being effectively atomic.
- None of the above holds on free-threaded builds. CI runs no `3.13t`/`3.14t` job and the package makes no free-threading declaration, so that is out of scope here.

### DIFF
--- a/tests/test_function.py
+++ b/tests/test_function.py
@@ -626,29 +626,41 @@ def test_name_after_clearing_cache(dispatch: plum.Dispatcher):
     assert some_function_name._resolver.function_name == "some_function_name"
 
 
-def _make_stringly_annotated_function() -> plum.Function:
-    """Create a fresh dispatch function with several overloads using *string*
-    annotations. String annotations force beartype's PEP 563 resolution
-    (`resolve_pep563`) to mutate `__annotations__` in place, which is the unsynchronised
-    operation that races across threads. See GitHub issue #274."""
-    dispatch = plum.Dispatcher()
+@pytest.fixture
+def make_stringly_annotated_function():
+    """Factory for a fresh, unresolved dispatch function with *string* annotations.
 
-    @dispatch
-    def f(x: "int") -> "str":
-        return "int"
+    String annotations force beartype's PEP 563 resolution (`resolve_pep563`) to mutate
+    `__annotations__` in place, which is the unsynchronised operation that races across
+    threads. See GitHub issue #274.
 
-    @dispatch
-    def f(x: "str") -> "str":
-        return "str"
+    This is a factory rather than the function itself because the race only occurs on a
+    `Function`'s first resolution, so each iteration needs its own `Dispatcher`. Reusing
+    one `Dispatcher` would redefine the same `Function` object instead, which exercises
+    re-registration rather than first resolution.
+    """
 
-    @dispatch
-    def f(x: "float") -> "str":
-        return "float"
+    def make() -> plum.Function:
+        dispatch = plum.Dispatcher()
 
-    return f
+        @dispatch
+        def f(x: "int") -> "str":
+            return "int"
+
+        @dispatch
+        def f(x: "str") -> "str":
+            return "str"
+
+        @dispatch
+        def f(x: "float") -> "str":
+            return "float"
+
+        return f
+
+    return make
 
 
-def test_concurrent_resolution_is_thread_safe():
+def test_concurrent_resolution_is_thread_safe(make_stringly_annotated_function):
     """Multiple threads racing `_resolve_pending_registrations` on the same `Function`
     must not corrupt its registrations. Without the lock this raises `AssertionError:
     ... not stringified type hint` (among other errors) because beartype's
@@ -670,7 +682,7 @@ def test_concurrent_resolution_is_thread_safe():
         # unresolved function each iteration and loop enough to trip it reliably.
         # Without the lock this fails on essentially every iteration.
         for _ in range(100):
-            f = _make_stringly_annotated_function()
+            f = make_stringly_annotated_function()
             barrier = threading.Barrier(n_threads)
             errors: list[BaseException] = []
 
--- a/tests/test_function.py
+++ b/tests/test_function.py
@@ -688,7 +688,7 @@ def test_concurrent_resolution_is_thread_safe(make_stringly_annotated_function):
 
             def worker(f=f, barrier=barrier, errors=errors):
                 try:
-                    barrier.wait()  # release all threads onto resolution together
+                    barrier.wait()  # Release all threads onto resolution together.
                     f._resolve_pending_registrations()
                 except BaseException as e:  # noqa: BLE001
                     errors.append(e)
--- a/tests/test_function.py
+++ b/tests/test_function.py
@@ -1,6 +1,8 @@
 import abc
 import os
+import sys
 import textwrap
+import threading
 import typing
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_function.py
```diff
@@ -622,3 +624,74 @@ def some_function_name(x: int):
...
 
     assert some_function_name._resolver.function_name == "some_function_name"
+
+
+def _make_stringly_annotated_function() -> plum.Function:
+    """Create a fresh dispatch function with several overloads using *string*
+    annotations. String annotations force beartype's PEP 563 resolution
+    (`resolve_pep563`) to mutate `__annotations__` in place, which is the
+    unsynchronised operation that races across threads (see issue #274)."""
+    dispatch = plum.Dispatcher()
```
wesselb:
```suggestion
def _make_function_with_string_annotations() -> plum.Function:
    """Create a fresh dispatch function with several overloads using string
    annotations."""
    dispatch = plum.Dispatcher()
```

A little verbose perhaps (I'm assuming this is from an LLM). Changed the name to make it a little easier to read.

Also, don't we have a fixture for this?

tests/test_function.py
```diff
@@ -622,3 +624,74 @@ def some_function_name(x: int):
...
+        # The race only occurs on a `Function`'s first resolution, so use a fresh,
+        # unresolved function each iteration and loop enough to trip it reliably.
+        for _ in range(100):
+            f = _make_stringly_annotated_function()
+            barrier = threading.Barrier(n_threads)
+            errors: list[BaseException] = []
+
+            def worker(f=f, barrier=barrier, errors=errors):
+                try:
+                    barrier.wait()  # release all threads onto resolution together
```
wesselb:
```suggestion
                    barrier.wait()  # Release all threads onto resolution together.
```

tests/test_function.py
```diff
@@ -622,3 +624,74 @@ def some_function_name(x: int):
...
+    `Function` must not corrupt its registrations. Without the lock this
+    intermittently raises `AssertionError: ... not stringified type hint` (and
+    other errors) because beartype's `resolve_pep563` mutates the shared
+    `__annotations__` dict concurrently. See issue #274.
+
+    Note: this exercises the resolution race that the per-`Function` lock fixes.
+    A separate, unrelated race in beartype's lazy type-checker *compilation*
+    (triggered later via `is_bearable` during matching) is tracked upstream in
+    beartype and is out of scope here.
+    """
```
wesselb:
```suggestion
    """Test that `_resolve_pending_registrations` is thread safe.
    
    `_resolve_pending_registrations` calls `resolve_pep563`, which is not thread safe,
    so `function._lock` is used, which is tested here.
    """
```

wesselb: Perhaps a little simpler and more to the point.

wesselb (approved):
Looks great, @nstarman! Thanks!!

Did you test that the test fails without the lock?

Could you also make sure that docstrings and comments use up the full 88 characters? I'm assuming most comments and docstrings are LLM generated. They seem to be truncated at 79 characters.

tests/test_function.py
nstarman: Yes - "A little verbose perhaps (I'm assuming this is from an LLM)" 
SGTM - "Changed the name to make it a little easier to read." 
Not that works for this test - "Also, don't we have a fixture for this?"  since the tested race condition is specific to a Function's first resolution. 
But I have updated to make a relevant pytest fixture here.

PR #282, opened 2026-07-13
Make `type[X]` dispatch faithful and cacheable

nstarman:
## Problem

A method signature containing `x: type[X]` (or `typing.Type[X]`) makes plum's dispatch **non-faithful**, which disables the dispatch cache for the **entire** function — every call re-runs full resolution and is slow. This is contagious: a single `type[X]` overload disables caching for all overloads of that function.

The argument to a `type[X]` parameter is a *class*, and `type(a_class)` is the metaclass (usually just `type`), which says nothing about whether the class is a subclass of `X`. `int` and `str` both have `type(...) == type`, yet `type[int]` accepts one and rejects the other. So the type-keyed cache (`tuple(map(type, args))`) cannot distinguish them, and `_is_faithful` correctly reported `type[X]` as unfaithful.

## Fix

**1. `_function._cache_key`** — a class argument is keyed on `type[arg]` (its class identity) rather than on `type(arg)`.

**2. `_type._is_faithful`** — `type[X]` is now reported faithful, since membership (`issubclass(x, X)`) depends only on the class identity of the argument, which the cache now keys on.

Both are required together: making `type[X]` faithful *without* refining the key would produce an **unsound** cache (`type[int]` and `type[str]` calls would both key to `(type,)` and return the wrong method).

**3. `Signature.dispatches_on_classes`** — the refined key is only used by functions that actually dispatch on a subscripted `type[X]`, computed once at registration time and aggregated onto the resolver next to `is_faithful`. This is not an optimization detail but a correction to the first two commits — see "Why the refined key is gated" below.

## Why it's sound

Matching itself never relies on `type(x)`: `Signature.compute_mismatches` always uses beartype's `is_bearable(v, t)` regardless of faithfulness. Faithfulness gates **only** whether the already-correct resolution result is memoized.

The key's soundness does **not** rest on the annotations being of the form `type[X]`. The key never takes part in matching — on a cache miss, `_resolve_method_with_cache` resolves against `args`, not against the key. So the key's only job is to define an equivalence relation on argument tuples, and the requirement is just *equal keys ⟹ equal resolution*.

Since `type[arg]` is an injective encoding of the identity of `arg`, the key is **strictly finer** than `type(arg)` for class arguments — it separates classes that the metaclass key would have merged — and refining an equivalence relation cannot introduce a wrong hit. No premise about the shape of the annotation is needed, which matters because plenty of non-`type[X]` annotations accept class arguments (`object`, bare `type`, `Hashable`, `Callable`, any custom metaclass).

<sub>Footnote: injectivity rests on `GenericAlias.__eq__`, which delegates to `==` on origin and args, so a metaclass with a pathological `__eq__` can make `type[A] == type[B]` for distinct classes. The previous key had the same exposure (both keyed to the metaclass), so this is not a regression — but the guarantee is precisely "as injective as class equality is".</sub>

## Why the refined key is gated

The first two commits applied the refined key unconditionally, and that was a bad trade. `_cache_key` replaced the C-level `tuple(map(type, args))` with a Python generator expression behind a function call, on the hottest path in the library — **143ns → 331ns, paid by every dispatched call** whether or not `type[X]` is involved.

Gating on `dispatches_on_classes` restores the fast path for everyone who doesn't use `type[X]`. Detection recurses through type arguments, so `type[int] | str` and `list[type[int]]` opt in too: over-approximating is safe (merely slower), under-approximating would make the cache unsound.

Gating also removes two side effects of the unconditional version:

- **Cache growth was global.** Every function receiving class arguments got one entry per distinct class, including functions with no `type[X]` anywhere (`def f(x: object)` called with 8 classes stored 8 entries instead of 1).
- **`invoke` shared the key namespace, order-dependently.** For `class Meta(type)` / `class C(metaclass=Meta)` and `def f(x: Meta)`, `f.invoke(type[C])` raised `NotFoundLookupError` on a cold cache but returned the method if `f(C)` had run first — `invoke` resolves by signature-encompassment (`type[C] <= Meta` is false) while `__call__` resolves by value (`is_bearable(C, Meta)` is true). Such functions no longer use identity keys, so this is now consistent either way.

## Results

Micro-benchmark, `timeit`, CPython 3.14, single machine — the two large effects are mechanistically explained, the ~3% residual on the common path is within measurement noise:

| | master | unconditional key | gated (HEAD) |
|---|---|---|---|
| `f(1, 2)` for `def f(x: int, y: int)` — common path | 420 ns | 647 ns | **435 ns** |
| `g(int)` for `def g(x: type[int])` | 2615 ns | 657 ns | **664 ns** |
| cache entries, `def h(x: object)` × 8 class args | 1 | 8 | **1** |

So: ~4× faster on `type[X]` dispatch, which now caches at all, with the common path left effectively where it was.

## Tests

- `test_type.py` — `is_faithful(type[int])`, `is_faithful(Type[int])`, `is_faithful(type[SomeClass])` are `True`; bare `type` still faithful.
- `test_signature.py` / `test_resolver.py` — `type[X]` signatures keep `is_faithful` `True`; mixing an unfaithful type still flips it to `False`.
- `test_signature.py::test_dispatches_on_classes_detects_nested_occurrences` — `type[X]` nested inside `|`, `list[...]`, `dict[...]` still opts in, in argument and varargs position.
- `test_signature.py::test_dispatches_on_classes_ignores_hints_without_type_subscript` — hints with no subscripted `type[X]`, including bare `type`, keep the fast key.
- `test_cache.py::test_cache_type_dispatch` — caching now happens for `type[X]` dispatch, with a cache-hit-vs-miss timing assertion.
- `test_cache.py::test_cache_class_key_only_when_dispatching_on_classes` — non-`type[X]` functions keep the single coarse key.
- `test_cache.py::test_cache_class_key_is_contagious_within_a_function` — a `type[X]` method registered *after* the cache is warm clears the stale coarse-keyed entries (registration is lazy, so the flag is read only after pending registrations resolve).
- `test_function.py::test_type_dispatch_correctness` — `f(5)` → instance method, `f(int)` → `type[int]` method (no collision), `type[Sub]` beats `type[Base]`; `invoke(int)` vs `invoke(type[int])` resolve independently.

`docs/types.md` updated to note that dispatching on classes via `type[X]` is faithful and cached.

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -106,10 +106,17 @@ isinstance(x, t) == issubclass(type(x), t)
 For example, `int` is faithful, since `type(1) == int`;
 but `Literal[1]` is not faithful, since `issubclass(int, Literal[1])` is false.
 
+Dispatching on a class via `type[X]` (or the equivalent `typing.Type[X]`) is also
+faithful, and therefore cached: membership in `type[X]` depends only on the class
+identity of the argument (whether `issubclass(x, X)` holds), so Plum keys such
+arguments on the class itself rather than on the metaclass `type(x)`.
+
 Methods which have signatures that depend only on faithful types will
 be performant.
 On the other hand, methods which have one or more signatures with one or more
 unfaithful types cannot use caching and will therefore be less performant.
+Note that this is per function: a single method with an unfaithful signature disables
+caching for _all_ methods of that function.
 
 Example:
 
@@ -148,6 +155,9 @@ True
 
 >>> is_faithful(Literal[1])
 False
+
+>>> is_faithful(type[int])
+True
 ```
 
 If you implement, e.g., a type with a custom `__instancecheck__`, then `is_faithful`
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -47,13 +47,21 @@ def _cache_key(args: tuple[object, ...], /) -> tuple[object, ...]:
     """Compute the dispatch cache key for a tuple of arguments.
 
     For a normal argument, the key element is its runtime type `type(arg)`. For an
-    argument that is itself a class, the key element is `type[arg]` instead: a class
-    argument *is* its own dispatch-relevant value (dispatch on `type[X]` depends on
-    `issubclass(arg, X)`, a function of the class identity alone, not of the metaclass
-    `type(arg)`). Keying such arguments on `type[arg]` makes dispatch on `type[X]`
-    cacheable. `type[arg]` is hashable, never collides with a runtime `type(x)` key,
-    and coincides with `invoke(type[arg])`, since resolving the concrete class `arg`
-    selects the same method as resolving the hint `type[arg]`.
+    argument that is itself a class, the key element is `type[arg]` instead, which makes
+    dispatch on `type[X]` cacheable: membership is `issubclass(arg, X)`, a function of
+    the class identity alone, not of the metaclass `type(arg)`.
+
+    Soundness does not depend on the *annotations* being of the form `type[X]`. The key
+    never takes part in matching: on a cache miss, resolution runs against the arguments
+    themselves, so the key only defines an equivalence relation on argument tuples, and
+    the requirement is merely that equal keys imply equal resolution. Since `type[arg]`
+    is an injective encoding of the identity of `arg`, this key is *strictly finer* than
+    `type(arg)` for class arguments -- it separates classes that the metaclass key would
+    have merged -- and refining an equivalence relation cannot introduce a wrong hit.
+
+    This key is only used for functions that actually dispatch on `type[X]`, since it is
+    slower to compute than `map(type, args)` and produces one cache entry per distinct
+    class argument. See `Signature.dispatches_on_classes`.
 
     Args:
         args (tuple[object, ...]): Arguments.
@@ -439,7 +447,12 @@ class Function(metaclass=_FunctionMeta):
             # Attempt to use the cache based on the types of the arguments.
             # At this point, `args` must be a tuple (not `Signature` or `None`).
             assert isinstance(args, tuple)
-            types = _cache_key(args)
+            # Only pay for the class-aware key on functions that actually dispatch
+            # on `type[X]`. Everyone else keeps the fast C-level `map(type, ...)`.
+            if self._resolver.dispatches_on_classes:
+                types = _cache_key(args)
+            else:
+                types = tuple(map(type, args))
         try:
             return self._cache[types]
         except KeyError:
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -43,6 +43,27 @@ _owner_transfer: dict[type, type] = {}
 a function (see :meth:`Function.owner`), make the corresponding value the owner."""
 
 
+def _cache_key(args: tuple[object, ...], /) -> tuple[object, ...]:
+    """Compute the dispatch cache key for a tuple of arguments.
+
+    For a normal argument, the key element is its runtime type `type(arg)`. For an
+    argument that is itself a class, the key element is `type[arg]` instead: a class
+    argument *is* its own dispatch-relevant value (dispatch on `type[X]` depends on
+    `issubclass(arg, X)`, a function of the class identity alone, not of the metaclass
+    `type(arg)`). Keying such arguments on `type[arg]` makes dispatch on `type[X]`
+    cacheable. `type[arg]` is hashable, never collides with a runtime `type(x)` key,
+    and coincides with `invoke(type[arg])`, since resolving the concrete class `arg`
+    selects the same method as resolving the hint `type[arg]`.
+
+    Args:
+        args (tuple[object, ...]): Arguments.
+
+    Returns:
+        tuple[object, ...]: Cache key.
+    """
+    return tuple(type[arg] if isinstance(arg, type) else type(arg) for arg in args)
+
+
 class _FunctionMeta(type):
     """:class:`Function` implements `__doc__`, which overrides the docstring of the
     class. This simple metaclass ensures that `Function.__doc__` still prints as the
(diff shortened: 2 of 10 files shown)


wesselb (commented):
This is very clever, @nstarman! I like it!

This basically looks good to go to me. I have only one question about the logic here:

```
    return tuple(type[arg] if isinstance(arg, type) else type(arg) for arg in args)
```

The simplification `isinstance(x, t) == issubclass(x, T)` with `t = type[T]` is valid whenever `t` is of the form `type[T]`, not whenever `x` is a type. Currently, keying by `type[x]` happens whenever `x` is a `type`, regardless of whether `t` is of the form `type[T]`. How do you justify this?

Is the reasoning that `isinstance(x, t)` with `x` a `type` can only be true if `t` is of the form `type[T]`? Is that true in generality?

nstarman:
@wesselb This was a good question, and one that actually led to preventing a performance regression for standard dispatch! 
Re your question no, `isinstance(x, t)` with `x` a type is not only true for `t = type[T]`. `object`, bare `type`, `Hashable`, `Callable`, and any custom-metaclass annotation are all counterexamples. For example, `def g(x: Callable)` accepts `g(int)`.  The key doesn't need that argument since it never takes part in matching: on a cache miss, `_resolve_method_with_cache` resolves against `args`, not against the key. So the key's job is to define an equivalence relation with the method resolution. 

test | master | first commit | now
-- | -- | -- | --
f(1, 2) common path | 420 ns | 647 ns | 435 ns
g(int) for def g(x: type[int]) | 2615 ns | 657 ns | 664 ns
cache entries, def h(x: object) × 8 classes | 1 | 8 | 1

wesselb:
Thanks, @nstarman!

I'm still a little worried about correctness in the general case. Sorry for the wall of text. Please bear with me!

The starting point is this: If the types in all methods are all faithful, then caching by _type_ is valid. Unfortunately, caching in this way is limited, because, e.g., it doesn't work for types of the form `type[T]`.

To expand the scope of types that we can cache for, let's define _cacheability_ more precisely. Call a type `t` _cacheable_ if, for all `x`, `isinstance(x, t)` only depends on `cache_key(x)`. Consequently, for a type `t`, if `cache_key(x)` is the same, then `isinstance(x, t)` must be the same. If `cache_key = type`, then a type is cacheable if it is faithful.

Following this construction, the goal is to make caching work for types of the form `type[t]`. We use `cache_key` from the PR. This `cache_key` does the following: if `isinstance(x, type)`, then `cache_key` encodes `x` itself (cache by _identity_); else, `cache_key(x)` returns `type(x)`. It is then always true that `isinstance(x, t)` is the same whenever `cache_key(x)` is the same?

Theoretically, I don't think that's the case! Consider

```python
@dispatch
def f(x: type[list]): ...

@dispatch
def f(x: type[tuple]): ...
```

and the calls `f(list[int])` and `f(tuple[int])`.

Now, `isinstance(list[int], type) == False`, so

```
cache_key(list[int]) = type(list[int]) = types.GenericAlias
cache_key(tuple[int]) = type(tuple[int]) = types.GenericAlias
```

which are the same. But clearly these arguments should match different methods! The point is that `isinstance(x, type[t])` could be true for `x` that is _not_ a `type`, which by the current `cache_key` could possibly get aliased to the same value `type(x)`.

Unfortunately, this particular example doesn't work, but that's for another reason: `list[int]` does not match `type[list]`:

```python
>>> is_bearable(list[int], type[list])
False
```

`TypeHint`, on the other hand, does do the right thing:

```python
>>> TypeHint(type[list[int]]) <= TypeHint(type[list])
True
```

Although the counterexample isn't functional, it would ease my mind if there were a simple and clear proof that the proposed caching strategy is valid in general.


---

Orthogonally, since this is getting pretty complicated conceptually, it would be good to be very clear on the definitions of everything. Whereas your modification of `_is_faithful` correctly implements the desired caching, `_is_faithful` now does not check whether something is faithful anymore! I was thinking of proposing that we should separate the definitions of cacheability and faithfulness and consequently separate these in the code, so have both `is_faithful(x)` and `is_cacheable(x)`, but we only really need `is_cacheable`.

Therefore, since you've identified a way to do caching with types that are not faithful (which is great!), perhaps the clearest approach would be to rework `is_faithful` into `is_cacheable` and get rid of the concept of faithfulness in the code:

```python
def cache_key(x: object) -> tuple[type, type | None]:
    """Cache key for `x`.

    If `x` is a type, then the key identifies `x`. Otherwise, the key identifies the type of `x`.
    """
    # Cache `type`s by identity!
    if isinstance(x, type):
        identity_key = x
    else:
        identity_key = None
    return (type(x), identity_key)


def is_cacheable(t: object):
    """Determine whether `t` is cacheable.

    If `t` is cacheable, then `is_bearable(x, t)` only depends on `cache_key(x)`.
    """
    ...
```

---

EDIT:

I was digging into it more, and it seems that `is_bearable(x, type[t])` iff `isinstance(x, type)` and `issubtype(x, t)`. I don’t really understand why `x` necessarily needs to be a `type` here. Semantically, `type[t]` should be the type of `t` and any subtype of `t`. Since, e.g., `t[int]` is a subtype of `t`, it should therefore also be a `type[t]`.

Additionally, perhaps ripping out `is_faithful` is a bit too radical, and it would be better to have `is_cacheable` as a new function, so alongside `is_faithful`.

nstarman:
These are excellent points. And very thorough!! I've spent the past few hours prodding and iterating with Claude + superpowers:writing-plans to look though this PR, your points, and the other open PRs — particularly https://github.com/beartype/plum/pull/272 — to see how we can make a unified framework that supports both `type[X]` and `Generic[type]` while being both performant and well-written, and also moves from `is_faithful` to an `is_cacheable` model. 
I'm opening an Issue that describes a PR sequence that will supersede this PR and will rebase and simplify #272.

nstarman:
TL;DR: we can get speedups of 7.9× on `type[X]`, 66× on `Literal`, 2.5× on builtins, 3.0× on uncacheable, and the common faithful path is 7% faster by adopting the changes proposed in #290 instead of this PR.

Add in the mypyc and we get 20% speedups on top of the previous ones almost across the board.

PR #285, opened 2026-07-27
build(mypyc): compile `_signature` for faster parametric dispatch

nstarman:
## Summary

Expands the `mypyc` compilation allowlist from 5 to 6 modules by adding `src/plum/_signature.py`. This compiles the `Signature` type and its comparison machinery to native code.

The allowlist lives under `[tool.hatch.build.targets.wheel.hooks.mypyc]` in `pyproject.toml` and still carries a `# TODO: include more files` marker — this is one step against that TODO.

## Performance

Measured with `tests/benchmark.py` (5-run medians, pure-Python editable install vs. `HATCH_BUILD_HOOKS_ENABLE=1` compiled wheel, same machine):

| Dispatch path | Pure Python | Compiled | Change |
|---|---|---|---|
| **Parametric calls** | **3.82 µs** | **3.12 µs** | **~18% faster** |
| Function calls | 0.42 µs | 0.44 µs | flat (noise) |
| Class calls | 2.00 µs | 2.00 µs | flat |
| Class-attribute calls | 2.01 µs | 2.06 µs | flat |

The parametric path improves because it exercises `Signature` type-hint comparison, now native. The flat paths are dominated by `_function.Function.__call__`, which is still interpreted (see note below).

## Required compatibility fix

Compiling `_signature` makes `Signature` a native extension type. That in turn caused the already-compiled `_dispatcher.multi` to enforce its `Signature | tuple[...]` vararg union **at the C boundary**, raising `TypeError` on bad input (e.g. `dispatch.multi(1)`) *before* the method's own `ValueError` — breaking `test_dispatcher.py::test_dispatch_multi` on the compiled build only.

Fix: annotate `Dispatcher.multi`'s `*signatures` parameter as `object`, so the existing runtime `isinstance` checks remain the sole validation. Behaviour is now identical between the pure-Python and compiled builds. (Ruff removed the now-unused `TypeHint` import as a side effect.)

## Testing

- Compiled wheel builds and `from plum import COMPILED; assert COMPILED` passes.
- `pytest tests/ -k "not incompatible_with_mypyc and not test_cache" --ignore=tests/advanced` passes on the compiled wheel.
- The pure-Python suite is unchanged.

> Note: `_function.py` — the per-call hot path and the biggest remaining win — cannot be compiled yet due to an upstream `mypyc` internal codegen error. Tracked separately in a follow-up issue. A further PR will add `_type`/`_util`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ include = [ # TODO: include more files
     "src/plum/_method.py",
     "src/plum/_overload.py",
     "src/plum/_resolver.py",
+    "src/plum/_signature.py",
 ]
 mypy-args = [
     "--ignore-missing-imports",
--- a/src/plum/_dispatcher.py
+++ b/src/plum/_dispatcher.py
@@ -85,7 +85,9 @@ class Dispatcher:
             elif isinstance(signature, tuple):
                 resolved_signatures.append(Signature(*signature))
             else:
-                raise ValueError(
+                # `TypeError` (not `ValueError`) to match the compiled build, where the
+                # typed vararg rejects bad input at the C boundary before this runs.
+                raise TypeError(
                     f"Signature `{signature}` must be a tuple or of type "
                     f"`plum.signature.Signature`."
                 )
--- a/tests/test_dispatcher.py
+++ b/tests/test_dispatcher.py
@@ -68,8 +68,10 @@ def test_dispatch_multi(dispatch: plum.Dispatcher):
     assert f("1") == "float or str"
     assert dispatch.functions["f"].methods[2].signature.precedence == 1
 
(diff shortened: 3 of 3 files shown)


nstarman:
@wesselb free 18% speedup 😎

wesselb (approved):
@nstarman Love it!! This is great. :)

nstarman:
Going to rebase to fix the merge conflict in pyproject.toml, then merge :shipit:

nstarman:
Oops. I merged PRs in the wrong order.
https://github.com/beartype/plum/pull/287 was stacked on this PR. 
I should have merged this first and then that one. 
Same code diff landed though 😆

Issue #286, opened 2026-07-27
mypyc: cannot compile `_function.py` yet — internal codegen error on `Function.dispatch`


nstarman:
## Context

The `mypyc` compilation allowlist in `pyproject.toml` (`[tool.hatch.build.targets.wheel.hooks.mypyc]`) carries a `# TODO: include more files`. `_function.py` is the highest-value remaining target: `Function.__call__` → `_resolve_method_with_cache` is the per-call dispatch hot path, so compiling it is where the biggest speedup lives. #285 added `_signature` and confirmed the simple-call benchmark paths are gated by `_function` remaining interpreted.

This issue tracks **why `_function.py` cannot be compiled yet**, so the TODO can note it explicitly.

## Blocker 1 (worked around): metaclass subclasses `type`

`_FunctionMeta(type)` (and, separately, the `type`-subclasses in `_type.py`/`_util.py`) hit:

```
error: Inheriting from most builtin types is unimplemented
note: Potential workaround: @mypy_extensions.mypyc_attr(native_class=False)
```

This is expected — mypyc cannot make a `type` subclass a native class. The documented workaround (`@mypyc_attr(native_class=False)`) works and keeps the metaclass interpreted while the rest of the module compiles. Guarding the `from mypy_extensions import mypyc_attr` import behind `try/except` avoids adding a runtime dependency.

## Blocker 2 (hard stop): internal mypyc codegen error on `Function.dispatch`

With Blocker 1 worked around, the build fails with an **internal mypyc error** during code generation:

```
src/plum/_function.py:<dispatch line>: AssertionError:
src/plum/_function.py:<dispatch line>: note: this is an internal mypyc error; please report a bug at https://github.com/mypyc/mypyc/issues
```

The `AssertionError` has an empty message and the traceback frames are swallowed even with `--show-traceback`.

### Reproduces on
- Python **3.13** and **3.14** (so it is not a 3.14-only issue; note release wheels target cp310–cp313).
- `mypy>=1.18.2`, `hatch-mypyc>=0.16.0`.
- The full compiled module set (`_bear`, `_dispatcher`, `_method`, `_overload`, `_resolver`, `_signature`, `_function`) with the hook's `--disable-error-code=*` args.

### Localization

The error is anchored to `Function.dispatch`. Stubbing it to an **untyped** signature compiles cleanly; **any typed signature** re-triggers the `AssertionError`. It survived every mitigation attempted:

[...]

PR #288, opened 2026-07-27
build(mypyc): compile `_function` natively for faster dispatch

nstarman:
## Summary

Adds `src/plum/_function.py` to the `mypyc` allowlist and makes **`Function` a native extension class**, cutting per-call dispatch overhead.

`tests/benchmark.py` (pure vs. compiled medians, same machine):

| Path | Pure | Compiled | Gain |
|---|---|---|---|
| Function calls | 0.42 µs | 0.32 µs | **~24%** |
| Parametric calls | 3.85 µs | 3.38 µs | **~12%** |

> A follow-up makes `_BoundFunction` native too, extending the win to class-method calls (~28%).

## Why "native", not just "compiled"

A non-native compiled class is actually **slower** than pure Python — every attribute access is a dict lookup and every method call goes through Python dispatch. The speedup only appears once `Function` is a native extension class. The cold features that forced it non-native are replaced:

- **Metaclass dropped.** A `_DocDescriptor` serves `__doc__` for both class access (`_class_doc`) and instance access (the computed merged docstring); a `_ModuleDescriptor` serves `__module__` as the wrapped function's module. Both attached with `setattr` (mypyc replaces a class-body `__doc__` with a filler; `__module__` is read-only on a native instance).
- **`functools.wraps` dropped.** `__name__`/`__qualname__`/`__wrapped__` are set explicitly in `__init__`, and every instance attribute is declared (including the `threading.RLock` from #274).

## Working around mypyc/mypyc#1205

mypyc cannot compile a closure that captures an enclosing variable on a class ([mypyc/mypyc#1205](https://github.com/mypyc/mypyc/issues/1205), plum tracking #286). `Function.dispatch`/`dispatch_multi` build their deferred decorators with `functools.partial` over a shared `_dispatch_one` helper; the `invoke` closures become the callable classes `_InvokedMethod` / `_BoundInvokedMethod`.

## Other compat fixes

`_promised_convert` is annotated `Callable[..., Any] | None`; the two `resolve_method` `except` clauses use distinct variable names (mypyc gives a reused exception variable a single type); `dispatch_multi` raises `TypeError` (not `ValueError`) to match the compiled C-boundary check (test updated).

## Testing

- Compiled wheel builds; `assert COMPILED` passes; full suite green (`-k "not incompatible_with_mypyc and not test_cache" --ignore=tests/advanced`).
- Pure-Python suite unchanged; lint, mypy, and pyright clean.

Known minor limitation: `Function.__module__` accessed on the *class* (not an instance) returns the descriptor object rather than a string (CPython does not invoke `__get__` for class-level `__module__`). Instance access is correct and `Function` instances are not picklable either way, so this is inert.

### DIFF
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -54,11 +54,10 @@ class _Wrappable(Protocol):
 
 
 def _wraps(wrapper: _Wrappable, wrapped: Callable[..., Any], /) -> None:
-    """:func:`functools.wraps` for a `mypyc` native wrapper.
+    """Copy `wrapped`'s metadata onto `wrapper`, like :func:`functools.wraps`.
 
-    Copies the metadata a native instance can hold. `functools.wraps` cannot be used: it
-    also writes the read-only native `__module__` and updates a `__dict__` that native
-    classes lack. `__module__`/`__doc__` are served by the descriptors below instead.
+    `functools.wraps` cannot be used: it writes the read-only native `__module__` and
+    updates a `__dict__` that native instances lack.
     """
     wrapper.__name__ = wrapped.__name__
     wrapper.__qualname__ = _generate_qualname(wrapped)
@@ -67,10 +66,11 @@ def _wraps(wrapper: _Wrappable, wrapped: Callable[..., Any], /) -> None:
 
 @mypyc_attr(native_class=False)
 class _InvokedMethod:
-    """Callable returned by :meth:`Function.invoke`: it runs the resolved `method` and
-    converts the result. A class (holding `method`/`return_type` as attributes) rather
-    than a `self`-capturing closure, which `mypyc` cannot compile (mypyc/mypyc#1205).
-    Non-native so :func:`functools.wraps` can copy `__name__`/`__doc__` onto instances.
+    """Run the resolved `method` and convert the result.
+
+    Callable returned by :meth:`Function.invoke`. A class rather than a closure,
+    which `mypyc` cannot compile (mypyc/mypyc#1205); non-native so
+    :func:`functools.wraps` can copy `__name__`/`__doc__` onto instances.
     """
 
     def __init__(
@@ -86,30 +86,28 @@ class _InvokedMethod:
 
 
 class _DocDescriptor:
-    """Serves `__doc__` for both class access (`-> _class_doc`) and instance access
-    (`-> _compute_doc()`, the computed/merged docstring), replacing the old metaclass.
-    Shared by `Function` and `_BoundFunction`. Attached with `setattr` below because
-    `mypyc` replaces a class-body `__doc__` with a filler. A descriptor (not a
-    metaclass) lets the classes be native, which makes the hot `__call__` path faster.
+    """Serve `__doc__`: `_class_doc` on class access, `_compute_doc()` on instances.
+
+    Shared by `Function` and `_BoundFunction`.
     """
 
     def __get__(self, instance: Any, owner: type) -> str | None:
         if instance is None:
-            return Function._class_doc
+            return getattr(owner, "_class_doc", None)
         doc: str | None = instance._compute_doc()
         return doc
 
 
 @mypyc_attr(native_class=False)
 class _ModuleDescriptor(str):
-    """Serves `__module__` as the wrapped function's module on instance access (as
-    `functools.wraps` did; native instances cannot hold `__module__` themselves), while
-    reading as a plain module string on class access. Shared by `Function`/`_BoundFn`.
+    """Serve `__module__` as the wrapped function's module on instance access.
 
-    A `str` subclass because CPython returns a class-level `__module__` verbatim without
-    calling `__get__`, so the value itself must be a valid module string for tools like
-    Sphinx; instance access does call `__get__`. (`str` subclass -> non-native, but
-    `__module__` access is not on the hot path.)
+    Shared by `Function` and `_BoundFunction`.
+
+    A `str` subclass because CPython returns a class-level `__module__` verbatim
+    without calling `__get__`, so the value itself must be a valid module string for
+    tools like Sphinx; instance access does call `__get__`. (A `str` subclass is
+    non-native, but `__module__` access is not on the hot path.)
     """
 
     __slots__ = ()
@@ -139,10 +137,10 @@ class Function:
     _owner_name: str | None
     _owner: type | None
     _warn_redefinition: bool
-    _pending: list[tuple[Callable[..., Any], "Signature | None", int | None]]
-    _resolved: list[tuple[Callable[..., Any], "Signature | None", int | None]]
+    _pending: list[tuple[Callable[..., Any], Signature | None, int | None]]
+    _resolved: list[tuple[Callable[..., Any], Signature | None, int | None]]
     _resolver: Resolver
-    _lock: "threading.RLock"
+    _lock: threading.RLock
     __name__: str
     __qualname__: str
     __wrapped__: Callable[..., Any]
@@ -200,11 +198,10 @@ class Function:
         return self._owner
 
     def _compute_doc(self) -> str | None:
-        """Compute the function's documentation: the documentation given at
-        initialisation with the documentation of all other registered methods appended.
+        """Compute the function's documentation.
 
-        Served as the instance `__doc__` by :class:`_DocDescriptor` (mypyc clobbers a
-        class-body property named `__doc__`).
+        This is the documentation given at initialisation, with the documentation of
+        all other registered methods appended.
         """
         try:
             self._resolve_pending_registrations()
@@ -272,8 +269,8 @@ class Function:
             function: Decorator.
         """
         if method is None:
-            return partial(self._dispatch_one, precedence=precedence)
-        return self._dispatch_one(method, precedence=precedence)
+            return partial(self._register_one, precedence=precedence)
+        return self._register_one(method, precedence=precedence)
 
     def dispatch_multi(
         self: Self, *signatures: Signature | tuple[TypeHint, ...]
@@ -300,29 +297,31 @@ class Function:
                     f"Signature `{signature}` must be a tuple or of type "
                     f"`plum.signature.Signature`."
                 )
-        return partial(self._dispatch_one, signatures=resolved_signatures)
+        return partial(self._register_multiple, signatures=resolved_signatures)
 
-    def _dispatch_one(
-        self: Self,
-        method: Callable[..., Any],
-        /,
-        *,
-        signatures: list[Signature] | None = None,
-        precedence: int = 0,
+    def _register_one(
+        self: Self, method: Callable[..., Any], /, *, precedence: int = 0
     ) -> Self:
-        """Register `method` and return the function itself.
+        """Register `method` by `precedence` and return the function itself.
 
-        Shared registration path for :meth:`dispatch` (single method, by `precedence`)
-        and :meth:`dispatch_multi` (explicit `signatures`). Kept a bound method so the
-        decorator forms can be built with :func:`functools.partial` rather than a
-        `self`-capturing closure, which `mypyc` cannot compile (mypyc/mypyc#1205).
+        Registration path for :meth:`dispatch`. A bound method so the decorator form
+        can be built with :func:`functools.partial` rather than a `self`-capturing
+        closure, which `mypyc` cannot compile (mypyc/mypyc#1205).
         """
-        if signatures is None:
-            self.register(method, precedence=precedence)
-        else:
+        self.register(method, precedence=precedence)
+        return self
+
+    def _register_multiple(
+        self: Self, method: Callable[..., Any], /, *, signatures: list[Signature]
+    ) -> Self:
+        """Register `method` for every signature and return the function itself.
+
+        Registration path for :meth:`dispatch_multi`. See :meth:`_register_one` for
+        why this is a bound method.
+        """
+        for signature in signatures:
             # `precedence` is derived from each signature, so it is left as `None`.
-            for signature in signatures:
-                self.register(method, signature=signature, precedence=None)
+            self.register(method, signature=signature, precedence=None)
         return self
 
     def clear_cache(self, reregister: bool = True) -> None:
@@ -640,21 +639,25 @@ class _BoundFunctionProto(Protocol):
 
 @mypyc_attr(native_class=False)
 class _BoundFunction:
-    """A bound instance of `.function.Function`.
+    #: The class-level docstring, served as `_BoundFunction.__doc__` by
+    #: `_DocDescriptor`.
+    _class_doc: ClassVar[str] = """A bound instance of `.function.Function`.
 
     Args:
         f (:class:`.function.Function`): Bound function.
         instance (object): Instance to which the function is bound.
     """
 
+    # Declared so `_wraps` can write them (see the `_Wrappable` protocol).
+    __name__: str
+    __qualname__: str
+    __wrapped__: Callable[..., Any]
+
     def __init__(self, f: "Function", instance: object) -> None:
         self._f: _BoundFunctionProto = f
         self._instance = instance
-        # Set explicitly instead of `functools.wraps`; `__doc__`/`__module__` are the
-        # descriptors attached below.
-        self.__name__ = f.__name__
-        self.__qualname__ = f.__qualname__
-        self.__wrapped__ = f._f
+        # `__doc__`/`__module__` are served by the descriptors attached below.
+        _wraps(self, f._f)
 
     def _compute_doc(self) -> str | None:
         return self._f.__doc__
@@ -681,7 +684,7 @@ class _BoundFunction:
 
 # See `Function` above: the descriptors serve `__doc__`/`__module__` on instances.
 setattr(_BoundFunction, "__doc__", _DocDescriptor())  # noqa: B010
-setattr(_BoundFunction, "__module__", _ModuleDescriptor())  # noqa: B010
+setattr(_BoundFunction, "__module__", _ModuleDescriptor(__name__))  # noqa: B010
 
 
 @mypyc_attr(native_class=False)
@@ -696,5 +699,6 @@ class _BoundInvokedMethod:
         wraps(bound.__wrapped__)(self)
 
     def __call__(self, *args: Any, **kw: Any) -> Any:
+        # TODO: Can we do this without `type` here?
         method = self._bound._f.invoke(type(self._bound._instance), *self._types)
         return method(self._bound._instance, *args, **kw)
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -5,18 +5,21 @@ import textwrap
 import threading
 from collections.abc import Callable
 from copy import copy
-from functools import wraps
+from functools import partial, wraps
 from types import MethodType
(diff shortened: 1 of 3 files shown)


nstarman:
Follow-up that makes `_BoundFunction` native too (extends the speedup to class-method dispatch): #289.

nstarman:
Ok. I need to fix the coveralls.
But this PR is a bit of a doozy. I iterated with Claude on the concept for a while, but less so on the code cleanup. Happy to try to do more. The point is that we need to turn `Function` into a native class to get 12-25% faster code. This unfortunately means hacking around metaclasses and the nice `functools.wraps`.

nstarman:
@leycec Based on work in other PRs and for Astropy, I've been working on these `mypyc` PRs here in `plum` for a bit, consistently getting >25% speedups in many functions when using compiled wheels. This is independent of performance refactors. I'd be happy to point Claude at `beartype` to analyze where, if anywhere, `beartype` could be sped up by compilations. (given the  dynamic code in beartype I suspect this is only in a limited number of places).

leycec:
**Ho, ho... ho.** Microoptimizing pure-Python is an autistic pet passion of mine. Anything that requires arcane black magic, obscure incantatory idioms, and copious hand-waving is something I'm probably already obsessed with. I must confess, therefore, that @beartype is unlikely to ever adopt or embrace non-pure-Python... well, *anything*.

Pure-Python is what @beartype does. Always has been. Always will be. It's why I'm here.

@beartype isn't like, say, Pydantic – whose entire core was refactored to Rust a few years back. From the Pydantic perspective, Python was just a utilitarian means to an ends. Abandoning Python for a lower-level systems language was an obvious non-decision for them.

@beartype takes a different perspective. @beartype is Team Python. I personally adore Python. I consider Python the language of both the present and the future. This language has intrinsic value to me. I also consider attempts to abandon Python to be shortsighted at best and counterproductive at worst. Technically, that's merely a subjective opinion. Pragmatically, though...

CPython now has an official JIT: `PYTHON_JIT=1`. CPython's JIT yields zero-cost performance gains on a yearly cadence with each new CPython release each October. Unsurprisingly, Python 3.15's main deliverable is just that: JIT gains of 10% geometric mean on average workloads. Python 3.16 is likely to be no different. Those gains come at no cost. Everyone gets them for free – everyone still in Python Land, that is.

Innate Python speedups inevitably take the wind out of non-Pythonic sails. Pivoting away from Python has always felt like the wrong design move to me. In 2026, as `PYTHON_JIT=1` continues marching on Mordor, that feeling has only intensified.

nstarman:
I largely agree with this. The one exception I’d point to is mypyc, since I don’t really see it as leaving Python. The source stays Python, you still ship pure-Python wheels, and mypyc just lets you also ship optimized wheels compiled from Python's type annotations. As Python improves (e.g. its JIT), I’d expect the gains to compound with mypyc rather than compete since you’re still running Python, just with strictly enforced static typing.

One thing I hope is that as mypyc matures, some of the more awkward pure-Python workarounds (like some in this PR) can hopefully disappear. The optimizations move into the compiler rather than the source code. This is a good long-term direction while staying firmly in Python-land.

nstarman:
@wesselb I think this PR is largely ready. If you agree, I think it would be a good idea to have a detailed comment at the top of the file explaining which parts of the change were to work around the `mypyc` bug and how the code can be refactored when/if that's fixed. It's mostly about the existence of the private `_InvokedMethod`, etc.

nstarman:
@wesselb this PR, in conjunction with #290 makes using dispatches with `type[X]` and `Literal[X]` about as fast as faithful instance annotations!

nstarman:
Ping @wesselb :)

src/plum/_function.py
```diff
@@ -44,20 +47,81 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
...
-    @property
-    def __doc__(self) -> str | None:  # type: ignore[override]
-        return self._class_doc
+def _wraps(wrapper: _Wrappable, wrapped: Callable[..., Any], /) -> None:
+    """:func:`functools.wraps` for a `mypyc` native wrapper.
+
+    Copies the metadata a native instance can hold. `functools.wraps` cannot be used: it
+    also writes the read-only native `__module__` and updates a `__dict__` that native
+    classes lack. `__module__`/`__doc__` are served by the descriptors below instead.
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention and length.

Imperative one-line summary, and the last sentence repeats what the descriptors and the `setattr` comment already document:

```suggestion
    """Copy `wrapped`'s metadata onto `wrapper`, like :func:`functools.wraps`.

    `functools.wraps` cannot be used: it writes the read-only native `__module__` and
    updates a `__dict__` that native instances lack.
    """
```

src/plum/_function.py
```diff
@@ -44,20 +47,81 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
...
+    wrapper.__wrapped__ = wrapped
+
+
+@mypyc_attr(native_class=False)
+class _InvokedMethod:
+    """Callable returned by :meth:`Function.invoke`: it runs the resolved `method` and
+    converts the result. A class (holding `method`/`return_type` as attributes) rather
+    than a `self`-capturing closure, which `mypyc` cannot compile (mypyc/mypyc#1205).
+    Non-native so :func:`functools.wraps` can copy `__name__`/`__doc__` onto instances.
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention; wording.

One-line imperative summary, elaboration below. Also, the old `invoke` closure captured the locals `method`/`return_type`, not `self` (the `self`-capturing lambda was in `dispatch`), so "a `self`-capturing closure" misdescribes what this class replaces:

```suggestion
    """Run the resolved `method` and convert the result.

    Callable returned by :meth:`Function.invoke`. A class rather than a closure,
    which `mypyc` cannot compile (mypyc/mypyc#1205); non-native so
    :func:`functools.wraps` can copy `__name__`/`__doc__` onto instances.
    """
```

src/plum/_function.py
```diff
@@ -44,20 +47,81 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
...
+        return _convert(self._method(*args, **kw), self._return_type)
+
+
+class _DocDescriptor:
+    """Serves `__doc__` for both class access (`-> _class_doc`) and instance access
+    (`-> _compute_doc()`, the computed/merged docstring), replacing the old metaclass.
+    Shared by `Function` and `_BoundFunction`. Attached with `setattr` below because
+    `mypyc` replaces a class-body `__doc__` with a filler. A descriptor (not a
+    metaclass) lets the classes be native, which makes the hot `__call__` path faster.
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention and length.

One-line imperative summary. The metaclass references are history rather than current design, and the `setattr` rationale is already documented at the `setattr` call sites — documenting each fact once keeps them from drifting:

```suggestion
    """Serve `__doc__`: `_class_doc` on class access, `_compute_doc()` on instances.

    Shared by `Function` and `_BoundFunction`.
    """
```

src/plum/_function.py
```diff
@@ -44,20 +47,81 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
...
+
+    def __get__(self, instance: Any, owner: type) -> str | None:
+        if instance is None:
+            return Function._class_doc
+        doc: str | None = instance._compute_doc()
+        return doc
```
wesselb, line 98: *Comment generated by Claude.*

Wrong class docstring for `_BoundFunction`.

The class-access branch returns `Function._class_doc` regardless of which class is being accessed, so `_BoundFunction.__doc__` (and `help(_BoundFunction)`) now shows "A function. ..." instead of "A bound instance of ...". Looking up the doc on `owner` fixes this:

```suggestion
    def __get__(self, instance: Any, owner: type) -> str | None:
        if instance is None:
            return getattr(owner, "_class_doc", None)
        doc: str | None = instance._compute_doc()
        return doc
```

This needs `_BoundFunction` to carry its own `_class_doc: ClassVar[str]` holding its class docstring, mirroring `Function` (reading `_BoundFunction.__doc__` at `setattr` time is not reliable under `mypyc`'s docstring filler). Please also add a regression test pinning class-level `__doc__` and `__module__` for both classes — that covers this and the empty-`__module__` bug in my other comment.

src/plum/_function.py
```diff
@@ -44,20 +47,81 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
...
+class _ModuleDescriptor(str):
+    """Serves `__module__` as the wrapped function's module on instance access (as
+    `functools.wraps` did; native instances cannot hold `__module__` themselves), while
+    reading as a plain module string on class access. Shared by `Function`/`_BoundFn`.
+
+    A `str` subclass because CPython returns a class-level `__module__` verbatim without
+    calling `__get__`, so the value itself must be a valid module string for tools like
+    Sphinx; instance access does call `__get__`. (`str` subclass -> non-native, but
+    `__module__` access is not on the hot path.)
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention; invented shorthand.

One-line imperative summary; "`_BoundFn`" is an abbreviation that exists nowhere else — write `_BoundFunction`. The wraps history and the read-only-`__module__` fact (documented at the `setattr` sites) can also go:

```suggestion
    """Serve `__module__` as the wrapped function's module on instance access.

    Shared by `Function` and `_BoundFunction`.

    A `str` subclass because CPython returns a class-level `__module__` verbatim
    without calling `__get__`, so the value itself must be a valid module string for
    tools like Sphinx; instance access does call `__get__`. (A `str` subclass is
    non-native, but `__module__` access is not on the hot path.)
    """
```

wesselb (commented):
*Review generated by Claude.*

Nice work getting `Function` native. I reviewed the pure-Python build — instance-level metadata, descriptor behaviour, and the test suite all check out there; the compiled-build claims were not independently verified. Two small bugs in the `_BoundFunction` descriptor wiring (see inline comments), plus minor docstring/style points.

One remark on the PR description: the "Known minor limitation" paragraph says class-level `Function.__module__` "returns the descriptor object rather than a string" — but the final design already fixes this: `_ModuleDescriptor` is a `str` subclass constructed with the module name, so class access returns a valid module string (verified: `'plum._function'`). The paragraph could be dropped or reworded.

src/plum/_function.py
```diff
@@ -66,11 +130,22 @@ class Function(metaclass=_FunctionMeta):
...
+    _f: Callable[..., Any]
+    _cache: dict[tuple[TypeHint, ...], tuple[Callable[..., Any], TypeHint]]
+    _doc: str
+    _owner_name: str | None
+    _owner: type | None
+    _warn_redefinition: bool
+    _pending: list[tuple[Callable[..., Any], "Signature | None", int | None]]
+    _resolved: list[tuple[Callable[..., Any], "Signature | None", int | None]]
+    _resolver: Resolver
+    _lock: "threading.RLock"
```
wesselb: *Comment generated by Claude.*

Unnecessary quoting.

`Signature` and `threading` are imported at module level, and the neighbouring declarations (`_resolver: Resolver`) are unquoted — the string forms can go:

```suggestion
    _pending: list[tuple[Callable[..., Any], Signature | None, int | None]]
    _resolved: list[tuple[Callable[..., Any], Signature | None, int | None]]
    _resolver: Resolver
    _lock: threading.RLock
```

src/plum/_function.py
```diff
@@ -130,13 +199,12 @@ def owner(self) -> type | None:
...
-        documentation of the function given at initialisation with the documentation
-        of all other registered methods appended.
+    def _compute_doc(self) -> str | None:
+        """Compute the function's documentation: the documentation given at
+        initialisation with the documentation of all other registered methods appended.
 
-        Upon instantiation, this property is available through `obj.__doc__`.
+        Served as the instance `__doc__` by :class:`_DocDescriptor` (mypyc clobbers a
+        class-body property named `__doc__`).
         """
```
wesselb: *Comment generated by Claude.*

Docstring convention and length.

One-line summary; the served-by paragraph belongs to `_DocDescriptor`, which already documents it:

```suggestion
        """Compute the function's documentation.

        This is the documentation given at initialisation, with the documentation of
        all other registered methods appended.
        """
```

src/plum/_function.py
```diff
@@ -610,3 +677,24 @@ def methods(self) -> MethodList:
     def dispatch(self) -> _DispatchFunction:
         """See :meth:`.Function.dispatch`."""
         return self._f.dispatch
+
+
+# See `Function` above: the descriptors serve `__doc__`/`__module__` on instances.
+setattr(_BoundFunction, "__doc__", _DocDescriptor())  # noqa: B010
+setattr(_BoundFunction, "__module__", _ModuleDescriptor())  # noqa: B010
```
wesselb: *Comment generated by Claude.*

Empty class-level `__module__`.

`_ModuleDescriptor()` without an argument is the empty string, so `_BoundFunction.__module__` (class access) is now `""` — unlike `Function`, which gets `_ModuleDescriptor(__name__)` and reads as `'plum._function'`. That breaks exactly the "value itself must be a valid module string for tools like Sphinx" property the descriptor's docstring promises:

```suggestion
setattr(_BoundFunction, "__module__", _ModuleDescriptor(__name__))  # noqa: B010
```

src/plum/_function.py
```diff
@@ -562,44 +647,26 @@ class _BoundFunction:
...
     def __init__(self, f: "Function", instance: object) -> None:
-        self._f = f
-        wraps(f._f)(self)  # This will call the setter for `__doc__`.
+        self._f: _BoundFunctionProto = f
         self._instance = instance
+        # Set explicitly instead of `functools.wraps`; `__doc__`/`__module__` are the
+        # descriptors attached below.
+        self.__name__ = f.__name__
+        self.__qualname__ = f.__qualname__
+        self.__wrapped__ = f._f
```
wesselb: *Comment generated by Claude.*

Reuse `_wraps`.

These three assignments are exactly `_wraps(self, f._f)`: `f.__name__` is `f._f.__name__` and `f.__qualname__` is `_generate_qualname(f._f)`, both set by `_wraps` in `Function.__init__`, and `__wrapped__` is `f._f` in both spellings. Reusing the helper gives `_wraps` a second call site (it is currently single-use) and keeps the wraps-replacement logic in one place:

```suggestion
        # `__doc__`/`__module__` are served by the descriptors attached below.
        _wraps(self, f._f)
```

src/plum/_function.py
```diff
@@ -233,18 +294,36 @@ def dispatch_multi(
...
                     f"`plum.signature.Signature`."
                 )
+        return partial(self._dispatch_one, signatures=resolved_signatures)
 
-        def decorator(method: Callable[..., Any]) -> "Function":
-            # The precedence will not be used, so we can safely set it to `None`.
-            for signature in resolved_signatures:
-                self.register(method, signature=signature, precedence=None)
-            return self
+    def _dispatch_one(
```
wesselb: Is this perhaps better called `_register_multiple`?

Can we require `signatures` to be `list[Signature]`, so disallow `None`?

src/plum/_function.py
```diff
@@ -610,3 +677,24 @@ def methods(self) -> MethodList:
...
+
+    def __init__(self, bound: "_BoundFunction", types: tuple[TypeHint, ...]) -> None:
+        self._bound = bound
+        self._types = types
+        # `bound.__wrapped__` is the underlying function (`f._f`), set in
+        # `_BoundFunction.__init__`.
+        wraps(bound.__wrapped__)(self)
+
+    def __call__(self, *args: Any, **kw: Any) -> Any:
+        method = self._bound._f.invoke(type(self._bound._instance), *self._types)
```
wesselb, line 703:
```suggestion
        # TODO: Can we do this without `type` here?
        method = self._bound._f.invoke(type(self._bound._instance), *self._types)
```

wesselb (approved):
This is pretty complicated, but I think it looks good! Thanks a lot, @nstarman!! These `mypyc` speed-ups are incredibly neat.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied. The correction is right: on `master`, `Function.invoke`'s closure captured `method`/`return_type`, and the `self`-capturing closure was the `dispatch` lambda — that one is now handled by `functools.partial` over `_register_one`, not by this class.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Confirmed and fixed. `_BoundFunction.__doc__` was showing `Function`'s docstring; `__get__` now reads `_class_doc` off `owner`, and `_BoundFunction` carries its own `_class_doc: ClassVar[str]` mirroring `Function`.

Regression test added (`test_class_doc_and_module`), parametrised over both classes and pinning class-level `__doc__` and `__module__` — it covers the empty-`__module__` bug too. Verified against a compiled wheel: `_BoundFunction.__doc__` is `"A bound instance of ..."` and `__module__` is `"plum._function"`.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied. Both are module-level imports, so the quotes were unnecessary; mypy and pyright are clean without them.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Confirmed and fixed. Covered by the new `test_class_doc_and_module`.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied. Verified the two spellings coincide: `_wraps(self, f)` in `Function.__init__` sets `f.__name__ = f._f.__name__` and `f.__qualname__ = _generate_qualname(f._f)`, so `_wraps(self, f._f)` here reproduces all three assignments exactly. `_BoundFunction` now declares `__name__`/`__qualname__`/`__wrapped__` so it satisfies the `_Wrappable` protocol.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Both done, by splitting rather than renaming: `dispatch` and `dispatch_multi` had genuinely different registration paths sharing one method only via the `signatures: ... | None` branch. It is now `_register_one(method, *, precedence)` and `_register_multiple(method, *, signatures: list[Signature])` — `signatures` is required and non-`None`, and each `functools.partial` targets its own method.

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied — the TODO was on `master` and got dropped when the closure became `_BoundInvokedMethod`; restored.

PR #289, opened 2026-07-27
build(mypyc): make `_BoundFunction` native too

nstarman:
Follow-up to #288 (**stacked on it** — until #288 merges, this PR's diff shows both commits; review the second, `build(mypyc): make _BoundFunction native too`).

## Summary

`Function.__get__` allocates a fresh `_BoundFunction` on **every** bound-method access, so making `_BoundFunction` a native extension class (rather than a non-native compiled one) speeds up class-method dispatch substantially.

`tests/benchmark.py` (pure vs. compiled medians):

| Path | Pure | Compiled | Gain (before → after) |
|---|---|---|---|
| Class calls | 0.88 µs | 0.63 µs | ~4% → **~28%** |
| Class-attr calls | 0.90 µs | 0.64 µs | ~4% → **~29%** |

Now the whole hot path is ~24-29% faster than pure Python.

## How

Same recipe as `Function` in #288: drop `@mypyc_attr(native_class=False)`, declare every instance attribute, and set `__name__`/`__qualname__`/`__wrapped__` explicitly in `__init__`. The `_DocDescriptor`/`_ModuleDescriptor` are reused unchanged — they already delegate to `instance._compute_doc()` / `instance._f.__module__`, both of which `_BoundFunction` provides.

## Testing

- Compiled wheel builds; full suite green.
- Pure-Python suite unchanged; lint, mypy, and pyright clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,6 +94,7 @@ include = [ # TODO: include more files
     "src/plum/_resolver.py",
     "src/plum/_signature.py",
     "src/plum/_type.py",
+    "src/plum/_function.py",
 ]
 mypy-args = [
     "--ignore-missing-imports",
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -5,18 +5,21 @@ import textwrap
 import threading
 from collections.abc import Callable
 from copy import copy
-from functools import wraps
+from functools import partial, wraps
 from types import MethodType
-from typing import Any, Protocol, TypeVar, overload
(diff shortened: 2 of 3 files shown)


nstarman:
@wesselb this PR is ready! Another ~25% speedup on some hot-path functionality.

nstarman:
Ping @wesselb :)

wesselb (approved):
Looks great, @nstarman! More speed-ups!! :)

PR #292, opened 2026-07-31
perf(promotion): skip conversion when a return annotation is already satisfied

nstarman:
## Summary

A method with a return annotation pays `convert` on **every call**. `convert` walks the annotation with `resolve_type_hint` *twice*, dispatches to find a conversion method, builds an `_InvokedMethod` wrapper via `invoke`, and runs a `beartype` check — almost always to hand back the object it was given, because a return annotation is overwhelmingly a type the method already returns.

`tests/benchmark.py`, medians (every pre-existing benchmark unchanged):

| Path | Before | After | Gain |
|---|---|---|---|
| Annotated return | 3.68 µs | 0.51 µs | **7.2×** |
| Annotated return, union | 7.03 µs | 0.50 µs | **14.1×** |

For scale: an *unannotated* return costs ~0.43 µs (7× native). Annotating the return made the same call 57×–109× native. After this change it is 8×.

Found while profiling [unxt](https://github.com/GalacticDynamics/unxt), where `u.unit(...)` — a three-method dispatch whose bodies are a dict lookup or an identity — cost ~9 µs a call. ~95% of that was this path; `resolve_type_hint` alone was 64% of cumulative time under cProfile, called 10× per `unit()` call.

## How

`plum._function.identity_conversions` records, per `(type(obj), target_type)`, whether conversion is the identity; `_convert` returns the object directly on a hit.

Two conditions make the answer a property of the **key** rather than of the object, which is what makes the memo unable to change a result:

* **the conversion resolved to the fallback method** — no conversion method applied, so the value came back unchanged after a type check;
* **`is_faithful(resolved)`** — defined in plum as `isinstance(x, t) == issubclass(type(x), t)`, so the fallback's check depends only on `type(obj)`. Without it the match can depend on the *value* (a `Literal`, a custom `__instancecheck__`) and must be re-run.

This is the same invariant `Function._cache` already relies on to key methods on argument types, applied to the return side.

Negative answers are recorded too. That matters: `is_faithful` is uncached and costs ~6 µs on a 4-way union, so without recording negatives `convert` would get *slower* for every pair the memo cannot help. With them, the analysis runs once per pair rather than once per call — the paths the memo can't short-circuit still come out ahead:

| Path | Before | After |
|---|---|---|
| `-> Literal[1]` (unfaithful, never memoised) | 6.40 µs | 5.81 µs |
| `-> Base` with a conversion method registered | 2.60 µs | 1.46 µs |
| `-> Any` (control, already short-circuited) | 0.357 µs | 0.357 µs |

`convert` also now calls the resolved method directly instead of through `invoke`, which built a wrapper object per call — that is where the non-memoised gains come from.

### Invalidation

Mirrors the existing method cache: `clear_all_cache()` clears it, and so does `add_conversion_method`, since a newly registered method can claim a pair that was previously an identity (`type_from` may be a subclass of `type_to`). The same in-place type-mutation caveat as `Function._cache` applies — `type_mapping`, a later-delivered `ModuleType` — and is already documented as requiring `clear_all_cache`.

## Testing

- Full suite green; **10 new tests** in `tests/test_promotion.py` pinning the soundness conditions, not the speed. They go through a dispatched function, because that — not `plum.convert` — is what reads the memo.
- Each guard was **mutation-tested**: removing the `is_faithful` gate, either invalidation hook, the fallback-method check, or the size bound each makes a specific test fail.
- Notable cases covered: a conversion method registered *after* a pair is memoised; two objects of the same type that a custom `__instancecheck__` judges differently; `Literal` targets; a raising conversion (never recorded); union targets; the size bound.
- `mypy --enable-error-code=no-redef`, `pyright`, and `ruff` clean.

## Compatibility

- **#288 (mypyc-native `Function`)**: verified — built the compiled wheel with these changes on top of a tree containing #288, `assert COMPILED` passes, `_function` is a native `.so`, and the mypyc CI test selection is green. The new module state is a plain annotated `dict` plus a `def`, mirroring the existing `_owner_transfer`. The size-limit constant deliberately lives in the *uncompiled* `_promotion` so it stays monkeypatchable under a compiled build.
- **`feat/cacheability-aspects`**: orthogonal — that work caches the *argument* side, this one the *return* side, which it leaves untouched (measured at 8.8 µs for a union return both before and after it). Applies cleanly on top; full suite and the new tests pass there too.
- Also validated end-to-end against unxt's 3982-test suite with this plum: all pass, and `u.unit(apyu.km)` goes 9.6 µs → 0.55 µs.

### DIFF
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -62,11 +62,11 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
     # walks the annotation with `resolve_type_hint` twice, resolves a conversion
     # method, and runs a `beartype` check -- all to hand back `obj`. Once a pair is
     # known to be an identity, none of that is needed.
-    try:
-        if identity_conversions.get((type(obj), target_type)):
-            return obj
-    except TypeError:  # `target_type` is unhashable, so it is never recorded.
-        pass
+    #
+    # An unhashable `target_type` raises `TypeError` here rather than below, which is
+    # what it has always done: the conversion path hashes it too, to look up a method.
+    if identity_conversions.get((type(obj), target_type)):
+        return obj
     assert _promised_convert is not None
     return _promised_convert(obj, target_type)
 
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -40,11 +40,6 @@ mutated in place afterwards (`plum.type_mapping`, an undelivered `ModuleType` th
 later delivered) requires `clear_all_cache`, as its docstring already states."""
 
 
-def _clear_identity_conversions() -> None:
-    """Clear :obj:`identity_conversions`. Called by :func:`plum.clear_all_cache`."""
-    identity_conversions.clear()
-
-
 def _convert(obj: Any, target_type: TypeHint, /) -> Any:
     """Convert an object to a particular type. Only converts if `target_type` is set.
 
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -22,6 +22,29 @@ _promised_convert = None
 SomeExceptionType = TypeVar("SomeExceptionType", bound=Exception)
 
 
+identity_conversions: dict[tuple[Any, Any], bool] = {}
+"""What conversion does for a pair `(type(obj), target_type)`, once it is known.
+
+`True` means conversion is the identity and can be skipped. `False` means it cannot be,
+and -- just as importantly -- that the pair has already been analysed, so the analysis
+(which includes an uncached `is_faithful` walk over the annotation) is not repeated on
+every call. A missing key means "not yet analysed", and the full path runs.
+
+Populated by :func:`plum.convert`, which owns the analysis because it is the module that
+can see the registered conversion methods; read here because this is the hot path. See
+that function for why a recorded answer is a property of the key alone.
+
+Cleared by :func:`plum.clear_all_cache` and whenever a conversion method is added. It
+carries the same staleness contract as :attr:`Function._cache`: a type whose meaning is
+mutated in place afterwards (`plum.type_mapping`, an undelivered `ModuleType` that is
+later delivered) requires `clear_all_cache`, as its docstring already states."""
+
+
+def _clear_identity_conversions() -> None:
+    """Clear :obj:`identity_conversions`. Called by :func:`plum.clear_all_cache`."""
+    identity_conversions.clear()
+
+
 def _convert(obj: Any, target_type: TypeHint, /) -> Any:
     """Convert an object to a particular type. Only converts if `target_type` is set.
 
@@ -34,9 +57,18 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
     """
     if target_type is Any:
         return obj
-    else:
-        assert _promised_convert is not None
-        return _promised_convert(obj, target_type)
+    # A return annotation is overwhelmingly a type the method already returns, making
+    # the conversion the identity. `convert` is nonetheless a dispatched call that
+    # walks the annotation with `resolve_type_hint` twice, resolves a conversion
+    # method, and runs a `beartype` check -- all to hand back `obj`. Once a pair is
+    # known to be an identity, none of that is needed.
+    try:
+        if identity_conversions.get((type(obj), target_type)):
+            return obj
+    except TypeError:  # `target_type` is unhashable, so it is never recorded.
+        pass
+    assert _promised_convert is not None
+    return _promised_convert(obj, target_type)
 
 
 _owner_transfer: dict[type, type] = {}
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -23,21 +23,11 @@ SomeExceptionType = TypeVar("SomeExceptionType", bound=Exception)
 
 
 identity_conversions: dict[tuple[Any, Any], bool] = {}
-"""What conversion does for a pair `(type(obj), target_type)`, once it is known.
+"""Whether conversion is the identity for a `(type(obj), target_type)` pair.
 
-`True` means conversion is the identity and can be skipped. `False` means it cannot be,
-and -- just as importantly -- that the pair has already been analysed, so the analysis
-(which includes an uncached `is_faithful` walk over the annotation) is not repeated on
-every call. A missing key means "not yet analysed", and the full path runs.
-
-Populated by :func:`plum.convert`, which owns the analysis because it is the module that
-can see the registered conversion methods; read here because this is the hot path. See
-that function for why a recorded answer is a property of the key alone.
-
-Cleared by :func:`plum.clear_all_cache` and whenever a conversion method is added. It
-carries the same staleness contract as :attr:`Function._cache`: a type whose meaning is
-mutated in place afterwards (`plum.type_mapping`, an undelivered `ModuleType` that is
-later delivered) requires `clear_all_cache`, as its docstring already states."""
+Written by :func:`plum.convert`, which decides what is recordable; read here because
+this is the hot path. Missing means "not yet analysed". Same staleness contract as
+:attr:`Function._cache`: mutating a type's meaning in place needs `clear_all_cache`."""
 
 
 def _convert(obj: Any, target_type: TypeHint, /) -> Any:
@@ -52,14 +42,8 @@ def _convert(obj: Any, target_type: TypeHint, /) -> Any:
     """
     if target_type is Any:
         return obj
-    # A return annotation is overwhelmingly a type the method already returns, making
-    # the conversion the identity. `convert` is nonetheless a dispatched call that
-    # walks the annotation with `resolve_type_hint` twice, resolves a conversion
-    # method, and runs a `beartype` check -- all to hand back `obj`. Once a pair is
-    # known to be an identity, none of that is needed.
-    #
-    # An unhashable `target_type` raises `TypeError` here rather than below, which is
-    # what it has always done: the conversion path hashes it too, to look up a method.
+    # Skip `convert` entirely when it is known to be the identity for this pair. An
+    # unhashable `target_type` raises here instead of below, as it always has.
     if identity_conversions.get((type(obj), target_type)):
         return obj
     assert _promised_convert is not None
--- a/src/plum/_promotion.py
+++ b/src/plum/_promotion.py
@@ -9,6 +9,7 @@ __all__ = [
 ]
 
 from collections.abc import Callable
+from contextlib import suppress
 from typing import TYPE_CHECKING, Any, Protocol, TypeAlias, TypeVar
 
 from beartype.door import TypeHint
@@ -16,7 +17,7 @@ from beartype.door import TypeHint
 import plum._function
 from ._bear import is_bearable
 from ._dispatcher import Dispatcher
-from ._type import resolve_type_hint
+from ._type import is_faithful, resolve_type_hint
 from .repr import repr_short
 
 T = TypeVar("T")
@@ -31,6 +32,18 @@ else:
 
 _dispatch = Dispatcher()
 
+_IDENTITY_CONVERSION_LIMIT = 4096
+"""Maximum number of entries in :obj:`plum._function.identity_conversions`.
+
+An entry is built from a return annotation and an argument's type, both of which come
+from the program's source rather than its data, so the natural size is the number of
+`(argument type, return annotation)` pairs ever returned through. The limit is a
+ceiling for the pathological case only: once it is reached, further pairs simply take
+the full conversion path, exactly as they did before this cache existed.
+
+The policy lives here rather than beside the set because this module decides what is
+memoisable; `_function` only stores the answer and reads it on the hot path."""
+
 
 @_dispatch
 def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
@@ -43,20 +56,67 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
     Returns:
         object: `obj` converted to type `type_to`.
     """
-    type_to = resolve_type_hint(type_to)
     # TODO: Can we implement this without using `type`?!
-    return _convert.invoke(type(obj), type_to)(obj, type_to)
+    type_from = type(obj)
+    cache = plum._function.identity_conversions
+    # `type_to` is keyed unresolved because that is what the hot path holds, and it is
+    # looked up before resolving so that a known pair costs a single lookup.
+    try:
+        known: bool | None = cache.get((type_from, type_to))
+    except TypeError:  # `type_to` is unhashable, so it can never be recorded.
+        known = False
+    if known:
+        return obj
+
+    resolved = resolve_type_hint(type_to)
+    # Resolve and call the conversion method directly rather than through
+    # `_convert.invoke`, which builds a wrapper object on every call.
+    method, return_type = _convert._resolve_method_with_cache(
+        types=(type_from, resolved)
+    )
+    result = plum._function._convert(method(obj, resolved), return_type)
+
+    # Record what conversion did for this pair, so that `plum._function._convert` can
+    # skip straight to the result next time -- and so that the analysis below, which
+    # walks the annotation, runs once per pair rather than once per call. Two
+    # conditions make the outcome a property of `(type_from, type_to)` alone rather
+    # than of `obj`:
+    #
+    #   * `method is _identity_conversion`, i.e. no conversion method applied and the
+    #     fallback returned `obj` unchanged. A conversion method may be registered
+    #     later, which is why `add_conversion_method` clears the cache.
+    #   * `is_faithful(resolved)`, which is *defined* as
+    #     `isinstance(x, t) == issubclass(type(x), t)`, so the fallback's check depends
+    #     only on `type_from`. Without it the check can depend on the value (a
+    #     `Literal`, a type with a custom `__instancecheck__`) and must be re-run.
+    #
+    # A raising call never reaches here, so a recorded `True` is always a conversion
+    # that succeeded and returned `obj` itself.
+    if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
+        with suppress(TypeError):  # `type_to` may be unhashable.
+            cache[type_from, type_to] = method is _identity_conversion and is_faithful(
+                resolved
+            )
+
+    return result
 
 
 # Deliver `convert`.
 plum._function._promised_convert = convert
 
 
-@_dispatch
-def _convert(obj, type_to):  # type: ignore[no-untyped-def]
+def _identity_conversion(obj, type_to):  # type: ignore[no-untyped-def]
+    """Fallback conversion: check the type and return `obj` unchanged.
+
+    Bound to a module-level name so that :func:`convert` can recognise it by identity
+    and conclude that no conversion method applied.
+    """
     if not is_bearable(obj, resolve_type_hint(type_to)):
         raise TypeError(f"Cannot convert `{obj}` to `{repr_short(type_to)}`.")
     return obj
+
+
+_convert = _dispatch(_identity_conversion)
 
 
 class _ConversionCallable(Protocol[T, R]):
@@ -80,6 +140,12 @@ def add_conversion_method(
     @_convert.dispatch
     def perform_conversion(obj: type_from, _: type_to):
         return f(obj)
+
+    # A pair previously recorded as an identity conversion may now be handled by this
+    # method: `type_from` can be a subclass of a `type_to` that an earlier call proved
+    # convertible only by the fallback. The cache is a pure optimisation, so dropping
+    # all of it is both correct and cheap -- conversion methods are registered rarely.
+    plum._function._clear_identity_conversions()
 
 
 def conversion_method(
--- a/src/plum/_promotion.py
+++ b/src/plum/_promotion.py
@@ -142,7 +142,7 @@ def add_conversion_method(
     # method: `type_from` can be a subclass of a `type_to` that an earlier call proved
     # convertible only by the fallback. The cache is a pure optimisation, so dropping
     # all of it is both correct and cheap -- conversion methods are registered rarely.
-    plum._function._clear_identity_conversions()
+    plum._function.identity_conversions.clear()
 
 
 def conversion_method(
--- a/src/plum/_promotion.py
+++ b/src/plum/_promotion.py
@@ -34,14 +34,8 @@ _dispatch = Dispatcher()
 _IDENTITY_CONVERSION_LIMIT = 4096
 """Maximum number of entries in :obj:`plum._function.identity_conversions`.
 
-An entry is built from a return annotation and an argument's type, both of which come
-from the program's source rather than its data, so the natural size is the number of
-`(argument type, return annotation)` pairs ever returned through. The limit is a
-ceiling for the pathological case only: once it is reached, further pairs simply take
-the full conversion path, exactly as they did before this cache existed.
-
-The policy lives here rather than beside the set because this module decides what is
-memoisable; `_function` only stores the answer and reads it on the hot path."""
+Entries come from annotations and argument types, not from data, so this is a ceiling
+for the pathological case only: past it, pairs just take the full conversion path."""
 
 
 @_dispatch
@@ -58,38 +52,24 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
     # TODO: Can we implement this without using `type`?!
     type_from = type(obj)
     cache = plum._function.identity_conversions
-    # `type_to` is keyed unresolved because that is what the hot path holds, and it is
-    # looked up before resolving so that a known pair costs a single lookup. An
-    # unhashable `type_to` raises `TypeError` here, as it already did further down when
-    # the conversion method was looked up.
+    # Keyed on the *unresolved* `type_to`, which is what the hot path holds.
     known: bool | None = cache.get((type_from, type_to))
     if known:
         return obj
 
     resolved = resolve_type_hint(type_to)
-    # Resolve and call the conversion method directly rather than through
-    # `_convert.invoke`, which builds a wrapper object on every call.
+    # Call the method directly; `_convert.invoke` builds a wrapper object per call.
     method, return_type = _convert._resolve_method_with_cache(
         types=(type_from, resolved)
     )
     result = plum._function._convert(method(obj, resolved), return_type)
 
-    # Record what conversion did for this pair, so that `plum._function._convert` can
-    # skip straight to the result next time -- and so that the analysis below, which
-    # walks the annotation, runs once per pair rather than once per call. Two
-    # conditions make the outcome a property of `(type_from, type_to)` alone rather
-    # than of `obj`:
-    #
-    #   * `method is _identity_conversion`, i.e. no conversion method applied and the
-    #     fallback returned `obj` unchanged. A conversion method may be registered
-    #     later, which is why `add_conversion_method` clears the cache.
-    #   * `is_faithful(resolved)`, which is *defined* as
-    #     `isinstance(x, t) == issubclass(type(x), t)`, so the fallback's check depends
-    #     only on `type_from`. Without it the check can depend on the value (a
-    #     `Literal`, a type with a custom `__instancecheck__`) and must be re-run.
-    #
-    # A raising call never reaches here, so a recorded `True` is always a conversion
-    # that succeeded and returned `obj` itself.
+    # Two conditions make the outcome a property of the key rather than of `obj`: no
+    # conversion method applied, so the fallback returned `obj` unchanged; and
+    # `is_faithful`, i.e. `isinstance(x, t) == issubclass(type(x), t)`, so that check
+    # depends only on `type_from`. A `Literal` or a custom `__instancecheck__` fails
+    # the second. Negatives are recorded too, or `is_faithful` -- which is not cached
+    # -- would rerun on every call it cannot help.
     if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
         cache[type_from, type_to] = method is _identity_conversion and is_faithful(
             resolved
@@ -138,10 +118,8 @@ def add_conversion_method(
     def perform_conversion(obj: type_from, _: type_to):
         return f(obj)
 
-    # A pair previously recorded as an identity conversion may now be handled by this
-    # method: `type_from` can be a subclass of a `type_to` that an earlier call proved
-    # convertible only by the fallback. The cache is a pure optimisation, so dropping
-    # all of it is both correct and cheap -- conversion methods are registered rarely.
+    # This method may now claim a pair recorded as an identity: `type_from` can be a
+    # subclass of `type_to`. Dropping the whole cache is cheap; registration is rare.
     plum._function.identity_conversions.clear()
 
 
--- a/src/plum/_promotion.py
+++ b/src/plum/_promotion.py
@@ -9,7 +9,6 @@ __all__ = [
 ]
 
 from collections.abc import Callable
-from contextlib import suppress
 from typing import TYPE_CHECKING, Any, Protocol, TypeAlias, TypeVar
 
 from beartype.door import TypeHint
@@ -60,11 +59,10 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
     type_from = type(obj)
     cache = plum._function.identity_conversions
     # `type_to` is keyed unresolved because that is what the hot path holds, and it is
-    # looked up before resolving so that a known pair costs a single lookup.
-    try:
-        known: bool | None = cache.get((type_from, type_to))
-    except TypeError:  # `type_to` is unhashable, so it can never be recorded.
-        known = False
+    # looked up before resolving so that a known pair costs a single lookup. An
+    # unhashable `type_to` raises `TypeError` here, as it already did further down when
+    # the conversion method was looked up.
+    known: bool | None = cache.get((type_from, type_to))
     if known:
         return obj
 
@@ -93,10 +91,9 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
     # A raising call never reaches here, so a recorded `True` is always a conversion
     # that succeeded and returned `obj` itself.
     if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
-        with suppress(TypeError):  # `type_to` may be unhashable.
-            cache[type_from, type_to] = method is _identity_conversion and is_faithful(
-                resolved
-            )
+        cache[type_from, type_to] = method is _identity_conversion and is_faithful(
+            resolved
+        )
 
     return result
 
--- a/tests/benchmark.py
+++ b/tests/benchmark.py
@@ -99,3 +99,46 @@ factor = int(np.round(dur_plum / dur_native))
 print("# Class Attribute Calls")
 print(f"Native call: {dur_native:6.2f} us ({1:.1f} x)")
 print(f"Plum call:   {dur_plum:6.2f} us ({factor:.1f} x)")
+print()
+
+
+# Every benchmark above returns through an unannotated method, which plum
+# short-circuits. An annotated return goes through `convert` instead, so it is measured
+# separately here -- and a union is the case that costs the most to check.
+
+
+class R1:
+    pass
+
+
+class R2:
+    pass
+
+
+def h_native(x):
+    return x
+
+
+@plum.dispatch
+def h(x: R1) -> R1:
+    return x
+
+
+@plum.dispatch
+def h_union(x: R1) -> R1 | R2:
+    return x
+
+
+r = R1()
+
+dur_native = benchmark(h_native, (r,), n=1000, burn=10)
+dur_plum = benchmark(h, (r,), n=1000, burn=10)
+dur_union = benchmark(h_union, (r,), n=1000, burn=10)
+
+factor = int(np.round(dur_plum / dur_native))
+factor_union = int(np.round(dur_union / dur_native))
+
+print("# Annotated Return Calls")
+print(f"Native call:      {dur_native:6.2f} us ({1:.1f} x)")
+print(f"Plum call:        {dur_plum:6.2f} us ({factor:.1f} x)")
+print(f"Plum call, union: {dur_union:6.2f} us ({factor_union:.1f} x)")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,7 +42,7 @@ def convert():
     _convert.clear_cache(reregister=True)
     # Restoring the methods changes which conversions are the identity, so the
     # memo of identity conversions has to go with them.
-    plum._function._clear_identity_conversions()
+    plum._function.identity_conversions.clear()
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -40,6 +40,9 @@ def convert():
     _convert._pending = []
     _convert._resolved = resolved
     _convert.clear_cache(reregister=True)
+    # Restoring the methods changes which conversions are the identity, so the
+    # memo of identity conversions has to go with them.
+    plum._function._clear_identity_conversions()
 
 
 @pytest.fixture
--- a/tests/test_promotion.py
+++ b/tests/test_promotion.py
@@ -219,9 +219,9 @@ def _memo():
 @pytest.fixture
 def memo(convert):
     """A cleared memo, restored afterwards along with the conversion methods."""
-    plum._function._clear_identity_conversions()
+    plum._function.identity_conversions.clear()
     yield _memo()
-    plum._function._clear_identity_conversions()
+    plum._function.identity_conversions.clear()
 
 
 def test_identity_conversion_is_memoised(memo, dispatch):
--- a/tests/test_promotion.py
+++ b/tests/test_promotion.py
@@ -204,3 +204,214 @@ def test_self_promotion(convert, promote):
         add_conversion_method(int, str, str)
         add_conversion_method(float, str, str)
         assert promote(1, 1.0) == ("1", "1.0")
+
+
+# The identity-conversion memo (`plum._function.identity_conversions`) lets a dispatched
+# call skip `convert` when its return annotation is already satisfied. These tests pin
+# the conditions that make the skip sound, not its speed. They go through a dispatched
+# function, because that -- not `plum.convert` -- is what reads the memo.
+
+
+def _memo():
+    return plum._function.identity_conversions
+
+
+@pytest.fixture
+def memo(convert):
+    """A cleared memo, restored afterwards along with the conversion methods."""
+    plum._function._clear_identity_conversions()
+    yield _memo()
+    plum._function._clear_identity_conversions()
+
+
+def test_identity_conversion_is_memoised(memo, dispatch):
+    class Base:
+        pass
+
+    @dispatch
+    def f(x: Base) -> Base:
+        return x
+
+    obj = Base()
+    assert f(obj) is obj
+    assert memo[Base, Base] is True
+    # The memoised answer is the same answer.
+    assert f(obj) is obj
+
+
+def test_memo_is_actually_read(memo, dispatch, monkeypatch):
+    """Once a pair is memoised, the call must not reach `convert` at all."""
+
+    class Base:
+        pass
+
+    @dispatch
+    def f(x: Base) -> Base:
+        return x
+
+    obj = Base()
+    assert f(obj) is obj  # Populates the memo.
+
+    def explode(*args, **kw):
+        raise AssertionError("convert was called despite a memoised identity")
+
+    monkeypatch.setattr(plum._function, "_promised_convert", explode)
+    assert f(obj) is obj
+
+
+def test_memo_yields_to_a_later_conversion_method(memo, dispatch):
+    """A conversion method registered *after* a pair is memoised must still win.
+
+    `Sub` is a `Base`, so the first call finds no conversion method and returns the
+    object unchanged. Registering `Sub -> Base` afterwards makes the memoised answer
+    wrong, so `add_conversion_method` has to drop it.
+    """
+
+    class Base:
+        pass
+
+    class Sub(Base):
+        pass
+
+    @dispatch
+    def f(x: Sub) -> Base:
+        return x
+
+    obj = Sub()
+    assert f(obj) is obj
+    assert memo[Sub, Base] is True
+
+    add_conversion_method(Sub, Base, lambda _: "converted")
+    assert f(obj) == "converted"
+
+
+def test_memo_not_used_for_unfaithful_targets(memo, dispatch):
+    """A target whose match depends on the value cannot be answered from the type.
+
+    Both objects are `Thing`s, so they share the key `(Thing, Gate)`, but `Gate`
+    decides membership from the value. Memoising the first call would make the second
+    wrongly succeed -- exactly what the faithfulness gate prevents.
+    """
+
+    class Meta(type):
+        def __instancecheck__(cls, instance):
+            return getattr(instance, "ok", False)
+
+    class Gate(metaclass=Meta):
+        pass
+
+    class Thing:
+        pass
+
+    @dispatch
+    def f(x: Thing) -> Gate:
+        return x
+
+    good, bad = Thing(), Thing()
+    good.ok, bad.ok = True, False
+
+    assert f(good) is good
+    assert memo[(Thing, Gate)] is False
+    with pytest.raises(TypeError):
+        f(bad)
+
+
+def test_memo_not_used_for_literals(memo, dispatch):
+    @dispatch
+    def f(x: int) -> typing.Literal[1]:
+        return x
+
+    assert f(1) == 1
+    assert memo[(int, typing.Literal[1])] is False
+    with pytest.raises(TypeError):
+        f(2)
+
+
+def test_memo_does_not_mask_failures(memo, dispatch):
+    class Base:
+        pass
+
+    @dispatch
+    def f(x: float) -> Base:
+        return x
+
+    with pytest.raises(TypeError):
+        f(1.0)
+    # A raising conversion never reaches the recording step, so nothing is stored.
+    assert (float, Base) not in memo
+    with pytest.raises(TypeError):
+        f(1.0)
+
+
+def test_memo_cleared_by_clear_all_cache(memo, dispatch):
+    class Base:
+        pass
+
+    @dispatch
+    def f(x: Base) -> Base:
+        return x
+
+    assert f(Base()) is not None
+    assert memo
+    plum.clear_all_cache()
+    assert not memo
+
+
+def test_memo_handles_union_targets(memo, dispatch):
+    """A union of faithful types is faithful, so it is memoisable as a whole."""
+
+    class A:
+        pass
+
+    class B:
+        pass
+
+    @dispatch
+    def f(x: A | float) -> A | B:
+        return x
+
+    a = A()
+    assert f(a) is a
+    assert memo[A, A | B] is True
+    with pytest.raises(TypeError):
+        f(1.0)
+
+
+def test_memo_is_bounded(memo, dispatch, monkeypatch):
+    """The memo stops growing at its limit instead of retaining types forever."""
+    monkeypatch.setattr(plum._promotion, "_IDENTITY_CONVERSION_LIMIT", 3)
+
+    @dispatch
+    def f(x: object) -> object:
+        return x
+
+    types = [type(f"C{i}", (), {}) for i in range(10)]
+    for t in types:
+        assert f(t()) is not None
+    assert len(memo) == 3
+
+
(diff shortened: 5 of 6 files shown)


nstarman:
Ping @wesselb, I think this PR is ready.
7-14x faster 😎.

nstarman:
@wesselb Should I add a speed benchmark to prevent future regressions?

wesselb:
@nstarman If there is a simple speed benchmark which prevents future regressions, then that would be fantastic! :)

src/plum/_function.py
```diff
@@ -22,6 +22,14 @@
 SomeExceptionType = TypeVar("SomeExceptionType", bound=Exception)
 
 
+identity_conversions: dict[tuple[Any, Any], bool] = {}
+"""Whether conversion is the identity for a `(type(obj), target_type)` pair.
+
+Written by :func:`plum.convert`, which decides what is recordable; read here because
+this is the hot path. Missing means "not yet analysed". Same staleness contract as
+:attr:`Function._cache`: mutating a type's meaning in place needs `clear_all_cache`."""
```
wesselb: *Comment generated by Claude.*

Annotation and docstring convention.

Tighten the key annotation and open the docstring with the `type: description` form:

```suggestion
identity_conversions: dict[tuple[type, TypeHint], bool] = {}
"""dict[tuple[type, TypeHint], bool]: Whether conversion is the identity for a
`(type(obj), target_type)` pair.

Written by :func:`plum.convert`, which decides what is recordable; read here because
this is the hot path. Missing means "not yet analysed". Same staleness contract as
:attr:`Function._cache`: mutating a type's meaning in place needs `clear_all_cache`."""
```

The key is always `(type(obj), target_type)`, so `tuple[type, TypeHint]` is more precise than `tuple[Any, Any]` (`TypeHint` is already imported); the neighbouring module attributes (`_owner_transfer`, `_promised_convert`) set the docstring form. Separately: those siblings are underscored and this module's `__all__` is `("Function",)` — consider `_identity_conversions` (call sites: `_dispatcher.py`, `_promotion.py`, `tests/conftest.py`, `tests/test_promotion.py`; no suggestion since the rename spans files).

src/plum/_promotion.py
```diff
@@ -43,22 +49,53 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
     Returns:
         object: `obj` converted to type `type_to`.
     """
-    type_to = resolve_type_hint(type_to)
     # TODO: Can we implement this without using `type`?!
-    return _convert.invoke(type(obj), type_to)(obj, type_to)
+    type_from = type(obj)
+    cache = plum._function.identity_conversions
+    # Keyed on the *unresolved* `type_to`, which is what the hot path holds.
```
wesselb: *Comment generated by Claude.*

Comment accuracy.

Change to:

```suggestion
    # Keyed on `type_to` as received, so that a cache hit avoids `resolve_type_hint`.
```

This states the reason for the keying choice, and drops "unresolved": the hot path's `type_to` (`Method.return_type`) has already been resolved once, by `extract_return_type` at registration — what the key is, is the hint *as the reader holds it*.

src/plum/_promotion.py
```diff
@@ -43,22 +49,53 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
...
+    known: bool | None = cache.get((type_from, type_to))
+    if known:
+        return obj
+
+    resolved = resolve_type_hint(type_to)
+    # Call the method directly; `_convert.invoke` builds a wrapper object per call.
+    method, return_type = _convert._resolve_method_with_cache(
+        types=(type_from, resolved)
+    )
+    result = plum._function._convert(method(obj, resolved), return_type)
```
wesselb: *Comment generated by Claude.*

Non-obvious intent; naming.

Add a line saying what the outer `_convert` is for, and rename `resolved` to `type_to_resolved` so the name says what was resolved:

```suggestion
    type_to_resolved = resolve_type_hint(type_to)
    # Call the method directly; `_convert.invoke` builds a wrapper object per call.
    method, return_type = _convert._resolve_method_with_cache(
        types=(type_from, type_to_resolved)
    )
    # As in `invoke`, convert the result per the method's own return annotation. This
    # is `Any` for every current conversion method, so it is effectively free.
    result = plum._function._convert(method(obj, type_to_resolved), return_type)
```

The `result` line reads as converting a second time. It actually inlines what `invoke`'s wrapper did: apply the resolved method's *own return-annotation* conversion — a no-op today, since every method `add_conversion_method` creates (and the fallback) is unannotated. Apply together with my comment below, which rewrites the one remaining use of `resolved`.

src/plum/_promotion.py
```diff
@@ -43,22 +49,53 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
...
+
+    # Two conditions make the outcome a property of the key rather than of `obj`: no
+    # conversion method applied, so the fallback returned `obj` unchanged; and
+    # `is_faithful`, i.e. `isinstance(x, t) == issubclass(type(x), t)`, so that check
+    # depends only on `type_from`. A `Literal` or a custom `__instancecheck__` fails
+    # the second. Negatives are recorded too, or `is_faithful` -- which is not cached
+    # -- would rerun on every call it cannot help.
```
wesselb: *Comment generated by Claude.*

Punctuation.

This code base avoids `--` as an em-dash; a colon reads more cleanly here anyway:

```suggestion
    # the second. Negatives are recorded too: `is_faithful` is not cached and would
    # rerun on every call it cannot help.
```

Same point in `benchmark.py` and the `test_promotion.py` section comment (separate comments).

src/plum/_promotion.py
```diff
@@ -43,22 +49,53 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
...
+    if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
+        cache[type_from, type_to] = method is _identity_conversion and is_faithful(
+            resolved
+        )
```
wesselb: *Comment generated by Claude.*

Companion to my comments above and below.

Change to (uses `_convert._f` from the comment below and `type_to_resolved` from the comment above; the break moves to the assignment instead of splitting `is_faithful`'s argument):

```suggestion
        # `_convert._f` is the fallback that `_convert` wraps, so `method` being it
        # means that no conversion method applied.
        cache[type_from, type_to] = (
            method is _convert._f and is_faithful(type_to_resolved)
        )
```

The three suggestions form one change and only apply together.

src/plum/_promotion.py
```diff
@@ -43,22 +49,53 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
...
+
+    Bound to a module-level name so that :func:`convert` can recognise it by identity
+    and conclude that no conversion method applied.
+    """
     if not is_bearable(obj, resolve_type_hint(type_to)):
         raise TypeError(f"Cannot convert `{obj}` to `{repr_short(type_to)}`.")
     return obj
 
 
+_convert = _dispatch(_identity_conversion)
```
wesselb: *Comment generated by Claude.*

Fallback rename leaks into error messages.

Keep the decorated form and recognise the fallback via `_convert._f`, which *is* the fallback implementation:

```suggestion
@_dispatch
def _convert(obj, type_to):  # type: ignore[no-untyped-def]
    """Fallback conversion: check the type and return `obj` unchanged.

    :func:`convert` recognises this fallback via `_convert._f` to conclude that no
    conversion method applied.
    """
    if not is_bearable(obj, resolve_type_hint(type_to)):
        raise TypeError(f"Cannot convert `{obj}` to `{repr_short(type_to)}`.")
    return obj
```

`Function.__name__` and the resolver's `function_name` come from the wrapped function, so the rename makes lookup errors and the `repr` read `_identity_conversion(...)` where they used to read `_convert(...)`: registering two conversion methods to the same target and converting an object matching both now raises ``AmbiguousLookupError: `_identity_conversion(__main__.C, __main__.T)` is ambiguous.`` (cf. the `NotFoundLookupError` surface seen in #295). Together with the `method is _convert._f` change in my comment above.

tests/benchmark.py
```diff
@@ -99,3 +99,46 @@ def go(self, x: str):
 print("# Class Attribute Calls")
 print(f"Native call: {dur_native:6.2f} us ({1:.1f} x)")
 print(f"Plum call:   {dur_plum:6.2f} us ({factor:.1f} x)")
+print()
+
+
+# Every benchmark above returns through an unannotated method, which plum
+# short-circuits. An annotated return goes through `convert` instead, so it is measured
+# separately here -- and a union is the case that costs the most to check.
```
wesselb: *Comment generated by Claude.*

Punctuation.

As in `_promotion.py`, single hyphen for an em-dash:

```suggestion
# separately here - and a union is the case that costs the most to check.
```

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
         add_conversion_method(int, str, str)
         add_conversion_method(float, str, str)
         assert promote(1, 1.0) == ("1", "1.0")
+
+
+# These pin the conditions that make the identity-conversion memo sound. They go
+# through a dispatched function, because that -- not `plum.convert` -- reads the memo.
```
wesselb: *Comment generated by Claude.*

Punctuation.

As in `_promotion.py`:

```suggestion
# through a dispatched function, because that - not `plum.convert` - reads the memo.
```

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
...
+
+def _memo():
+    return plum._function.identity_conversions
+
+
+@pytest.fixture
+def memo(convert):
+    """A cleared memo, restored afterwards along with the conversion methods."""
+    plum._function.identity_conversions.clear()
+    yield _memo()
```
wesselb: *Comment generated by Claude.*

Unnecessary indirection; naming.

Inline the single-use `_memo` helper, and name the fixture after the dict it manages, mirroring the `convert` fixture:

```suggestion
@pytest.fixture
def identity_conversions(convert):
    """`plum._function.identity_conversions`, cleared before and after the test."""
    plum._function.identity_conversions.clear()
    yield plum._function.identity_conversions
```

The `memo` parameters in the tests follow mechanically; see also the naming remark in the review body.

wesselb (commented):
*Review generated by Claude.*

Very nice speedup, and an unusually well-pinned test set — each guard has a test that fails if it is removed. I checked the memo's soundness argument (the `is_faithful` gate, recording only after a non-raising call, the invalidation hooks) and ran the suite and lints; all good. The one substantive comment is that the `_convert` → `_identity_conversion` rename leaks into user-facing error messages; the rest is naming, punctuation, and test consolidation.

One general remark on naming: the tests call this machinery "the memo" (fixture, `test_memo_*` names, docstrings), but the word appears nowhere in the source — the dict is `identity_conversions` — and the extra vocabulary is confusing on a first read. Please rename towards the source's own terms: the fixture and merged-test suggestions already use `identity_conversions`/"recorded", and the remaining `test_memo_*` names and docstrings can follow the same pattern.

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
...
+        return x
+
+    obj = Base()
+    assert f(obj) is obj  # Populates the memo.
+
+    def explode(*args, **kw):
+        raise AssertionError("convert was called despite a memoised identity")
+
+    monkeypatch.setattr(plum._function, "_promised_convert", explode)
+    assert f(obj) is obj
```
wesselb: *Comment generated by Claude.*

Merge.

The second test subsumes the first: the monkeypatched re-call proves both that the pair was recorded and that the recorded answer is served, which is all the plain re-call shows. One test with the entry assertion covers everything:

```suggestion
def test_identity_conversion_is_recorded_and_read(
    identity_conversions, dispatch, monkeypatch
):
    """Once a pair is recorded, a repeat call must not reach `convert` at all."""

    class Base:
        pass

    @dispatch
    def f(x: Base) -> Base:
        return x

    obj = Base()
    assert f(obj) is obj
    assert identity_conversions[Base, Base] is True

    def explode(*args, **kw):
        raise AssertionError("convert was called despite a recorded identity")

    monkeypatch.setattr(plum._function, "_promised_convert", explode)
    assert f(obj) is obj
```

Uses the fixture name from my comment above; with the fixture kept as `memo`, the same merge applies with the old names.

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
...
+    @dispatch
+    def f(x: Sub) -> Base:
+        return x
+
+    obj = Sub()
+    assert f(obj) is obj
+    assert memo[Sub, Base] is True
+
+    add_conversion_method(Sub, Base, lambda _: "converted")
+    assert f(obj) == "converted"
```
wesselb: *Comment generated by Claude.*

Merge.

This and `test_memo_not_used_when_a_conversion_method_applies` build the identical `Base`/`Sub`/`f`/conversion scaffolding and end in the same state; one test covers both behaviours by asserting what the pair is recorded as after the conversion method wins:

```suggestion
def test_conversion_methods_beat_recorded_identities(identity_conversions, dispatch):
    """A conversion method must win regardless of registration order.

    Registered *after* the pair is recorded as an identity, it must invalidate the
    entry; once it claims the pair, the pair must be recorded as not an identity.
    """

    class Base:
        pass

    class Sub(Base):
        pass

    @dispatch
    def f(x: Sub) -> Base:
        return x

    obj = Sub()
    assert f(obj) is obj
    assert identity_conversions[Sub, Base] is True

    add_conversion_method(Sub, Base, lambda _: "converted")
    assert f(obj) == "converted"
    # The re-analysed pair is no longer an identity, and the call keeps converting.
    assert identity_conversions[Sub, Base] is False
    assert f(obj) == "converted"
```

The other test is then deleted (see my comment there). Its register-first ordering adds no coverage: after the invalidation here, the pair is re-analysed with the conversion method in place — exactly the register-first state.

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
...
+
+def test_memo_not_used_for_literals(memo, dispatch):
+    @dispatch
+    def f(x: int) -> typing.Literal[1]:
+        return x
+
+    assert f(1) == 1
+    assert memo[(int, typing.Literal[1])] is False
+    with pytest.raises(TypeError):
+        f(2)
```
wesselb: *Comment generated by Claude.*

Merge.

Both tests pin the same condition — a value-dependent target must never be recorded as an identity — for the two ways it arises. Merging also fixes the two-line docstring summary:

```suggestion
def test_unfaithful_targets_are_not_recorded(identity_conversions, dispatch):
    """`Gate` and `Literal` match on the value, so their pairs must never be recorded.

    Both `Thing` objects share the key `(Thing, Gate)`: recording the first call
    would make the second wrongly succeed.
    """

    class Meta(type):
        def __instancecheck__(cls, instance):
            return getattr(instance, "ok", False)

    class Gate(metaclass=Meta):
        pass

    class Thing:
        pass

    @dispatch
    def f(x: Thing) -> Gate:
        return x

    good, bad = Thing(), Thing()
    good.ok, bad.ok = True, False

    assert f(good) is good
    assert identity_conversions[(Thing, Gate)] is False
    with pytest.raises(TypeError):
        f(bad)

    @dispatch
    def g(x: int) -> typing.Literal[1]:
        return x

    assert g(1) == 1
    assert identity_conversions[(int, typing.Literal[1])] is False
    with pytest.raises(TypeError):
        g(2)
```

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
...
+    def f(x: Sub) -> Base:
+        return x
+
+    obj = Sub()
+    assert f(obj) == "converted"
+    assert memo[(Sub, Base)] is False
+    # The second call must convert too, rather than take a memoised shortcut.
+    assert f(obj) == "converted"
+
+
```
wesselb: *Comment generated by Claude.*

Merge (companion).

Delete; covered by the merged registration-order test above (see my comment on `test_memo_yields_to_a_later_conversion_method`):

```suggestion
```

tests/test_promotion.py
```diff
@@ -204,3 +204,220 @@ def test_self_promotion(convert, promote):
...
+def test_unhashable_target_still_raises(memo, dispatch):
+    """The memo hashes `(type(obj), type_to)`, but so does the method lookup it
+    replaces, so such a target has never been usable."""
```
wesselb: *Comment generated by Claude.*

Docstring convention.

One-line summary, elaboration below:

```suggestion
    """An unhashable target raises, as it always has.

    The cache hashes `(type(obj), type_to)`, but so does the method lookup it
    replaces, so such a target has never been usable.
    """
```

wesselb (commented):
Yet another speed-up! 🚀 Looking very great, @nstarman. I've left some Claude-assisted comments above together with some general feedback (e.g. about the memo terminology).

src/plum/_function.py
nstarman: *Comment generated by Claude.*

Applied, including the rename. `tuple[type, TypeHint]` is right — the key is built only in `convert`, always as `(type(obj), type_to)`. Renamed to `_identity_conversions` too: all five references are in `src` and `tests` with none in the docs, so it was mechanical.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied. Confirmed the premise: `Method.__init__` fills `return_type` via `extract_return_type`, which ends in `resolve_type_hint`, so the hint the hot path holds has already been resolved once at registration. "Unresolved" was wrong.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied. `--` appeared nowhere else in the code base — the four occurrences were all introduced by this PR, and all four are now gone.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied, with one deviation on the line break. `cache[type_from, type_to] = method is _convert._f and is_faithful(type_to_resolved)` is 91 characters, so at the 88-column limit `ruff format` rewrites the suggested parenthesised form straight back into the `is_faithful(\n    type_to_resolved\n)` split you were avoiding. A local keeps your intent and is stable under the formatter:

```python
        no_method_applied = method is _convert._f
        cache[type_from, type_to] = no_method_applied and is_faithful(type_to_resolved)
```

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Confirmed and applied — this was a real user-facing regression, reproduced before and after:

```
# PR head
AmbiguousLookupError: `_identity_conversion(__main__.C, __main__.T)` is ambiguous.
# master, and now this branch again
AmbiguousLookupError: `_convert(__main__.C, __main__.T)` is ambiguous.
```

`repr(_convert)` was affected the same way. `_convert._f` is the fallback implementation, so the identity check still resolves correctly.

tests/benchmark.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested, using the renamed `_identity_conversions`.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested. Agreed the merged test subsumes both.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested. The reasoning holds: `add_conversion_method` clears the whole dict, so the post-invalidation re-analysis happens with the conversion method already registered — the register-first state exactly.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Deleted, as covered by the merged registration-order test.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

nstarman:
*Comment generated by Claude.*

Thanks — all 14 comments applied, plus the general naming remark, in `342f8ca`.

**On the substantive one:** the `_convert` → `_identity_conversion` rename was a real user-facing regression, and I reproduced it in both directions before fixing. The fallback is now the decorated `@_dispatch def _convert` again, recognised via `_convert._f`.

**On the naming remark:** the "memo" vocabulary is gone — the fixture, the test names, and the docstrings now use the source’s own terms (`identity_conversions`, "recorded"). I also took the parenthesised `_identity_conversions` rename you floated: all five references live in `src` and `tests`, none in the docs, so it was mechanical.

**One deviation,** noted inline: the suggested line break in the recording assignment is 91 characters, so `ruff format` rewrites it back into the `is_faithful(...)` split you were avoiding. A local (`no_method_applied`) preserves the intent and is stable under the formatter.

**Verification.** Since the merges cut the test count from 11 to 8, I re-ran the mutation checks to confirm nothing stopped being pinned — each guard still fails at least one test when removed:

| Mutation | Result |
|---|---|
| drop the `is_faithful` gate | 1 failed |
| drop the fallback-method check | 6 failed |
| drop the `add_conversion_method` invalidation | 1 failed |
| drop the `clear_all_cache` hook | 1 failed |
| drop the size bound | 1 failed |

Full suite green (171 passed; the one failure, `test_methodlist_repr`, fails identically on master). `ruff`, `ruff format`, and `mypy --enable-error-code=no-redef` clean; `pyright` unchanged at exactly the same error count as the unmodified PR head. Benchmarks unchanged, as expected — the hot path (`_identity_conversions.get`) was not touched: annotated return 0.489 µs, union return 0.499 µs.

nstarman:
*Comment generated by Claude.*

> If there is a simple speed benchmark which prevents future regressions, then that would be fantastic! :)

Added in `d77b250` as `tests/test_benchmark.py`. `tests/benchmark.py` only prints, so nothing fails when a fast path goes away; this runs under pytest, and so in CI.

**It asserts a ratio, never an absolute duration** — an annotated-return dispatched call against an unannotated one, timed in the same process. Absolute timings move by an order of magnitude across machines, interpreters and CI load; the ratio does not, because both measurements move together.

Measured bands, nine runs each, spread ~2.5%:

| State | Annotated | Union |
|---|---|---|
| Both short-circuits (healthy) | ~1.2 | ~1.2 |
| `_function._convert` fast path gone | ~2.3 | ~2.3 |
| Cache never consulted (pre-#292) | ~7.4 | ~23.3 |

There are two short-circuits, not one — `_function._convert` skips the call into `convert` entirely, and `convert` itself returns early on a hit. Losing only the first still costs ~2x, so the **1.75x** bound is set to catch that too: ~45% above healthy, ~30% below the nearest failure, on a ~2.5% spread.

**Verified by mutation rather than by assertion:** disabling either short-circuit fails the test, with the observed ratio in the message —

```
AssertionError: an annotated return cost 2.3x an unannotated one, over the 1.75x
bound; the identity-conversion fast path has likely regressed
```

— and restoring it passes.

**On the compiled build:** the ratio is 1.16–1.24 under a mypyc wheel, versus 1.19–1.22 pure, so the cibuildwheel test selection stays green. That insensitivity is also the reason I expect it to survive rebasing onto #288: the ratio barely moved when `_dispatcher`/`_method`/`_type` went native, since numerator and denominator speed up together. Worth a glance after that rebase all the same.

It carries a `benchmark` marker, so `-m 'not benchmark'` drops it if it ever proves flaky on a runner I cannot see.

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+    @dispatch
+    def f(x: Base) -> Base:
+        return x
+
+    obj = Base()
+    assert f(obj) is obj
+    assert identity_conversions[Base, Base] is True
+
+    def explode(*args, **kw):
+        raise AssertionError("convert was called despite a recorded identity")
```
wesselb:
```suggestion
        raise AssertionError("Convert was called despite a recorded identity.")
```

src/plum/_promotion.py
```diff
@@ -16,7 +16,7 @@
 import plum._function
 from ._bear import is_bearable
 from ._dispatcher import Dispatcher
```
wesselb, line 18: *Comment generated by Claude.*

Companion to the `invoke` comment below.

Import the resolver's unwrapping helper, used there to recognise the fallback:

```suggestion
from ._dispatcher import Dispatcher
from ._resolver import _unwrap_invoked_methods
```

`_unwrap_invoked_methods` already exists in `_resolver.py` to undo exactly the wrapping that `invoke` applies.

src/plum/_promotion.py
```diff
@@ -31,6 +31,12 @@
 
 _dispatch = Dispatcher()
 
+_IDENTITY_CONVERSION_LIMIT = 4096
+"""Maximum number of entries in :obj:`plum._function._identity_conversions`.
```
wesselb: *Comment generated by Claude.*

Docstring convention.

Open with the `type: description` form, as the sibling module attributes do:

```suggestion
"""int: Maximum number of entries in :obj:`plum._function._identity_conversions`.
```

`_promised_convert`, `_owner_transfer`, and `_identity_conversions` all follow this form.

src/plum/_promotion.py
```diff
@@ -43,9 +49,36 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
     Returns:
         object: `obj` converted to type `type_to`.
     """
-    type_to = resolve_type_hint(type_to)
     # TODO: Can we implement this without using `type`?!
```
wesselb, line 59: *Comment generated by Claude.*

Explain the caching up front.

Open the body with a plain-language summary of what the cache does and why it is sound:

```suggestion
    # The fallback conversion method only checks that `obj` is an instance of `type_to`
    # and returns `obj` unchanged. Every call of a function with a return annotation
    # goes through here, so this is worth optimising. If `type_to` is faithful, the
    # instance check depends only on `type(obj)`, so whether the conversion is the
    # identity can be cached by `(type(obj), type_to)`. That is what is implemented
    # below.

    # TODO: Can we implement this without using `type`?!
```

With this in place, the comment above the recording step can be much shorter; see my comment there.

src/plum/_promotion.py
```diff
@@ -43,9 +49,36 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
...
+        return obj
+
+    type_to_resolved = resolve_type_hint(type_to)
+    # Call the method directly; `_convert.invoke` builds a wrapper object per call.
+    method, return_type = _convert._resolve_method_with_cache(
+        types=(type_from, type_to_resolved)
+    )
+    # As in `invoke`, convert the result per the method's own return annotation. This
+    # is `Any` for every current conversion method, so it is effectively free.
+    result = plum._function._convert(method(obj, type_to_resolved), return_type)
```
wesselb, line 76: *Comment generated by Claude.*

Use `invoke` for clarity.

`invoke` is the public entry point and already applies the method's return-type conversion, so the private `_resolve_method_with_cache` call and the manual conversion can go:

```suggestion
    type_to_resolved = resolve_type_hint(type_to)
    method = _convert.invoke(type_from, type_to_resolved)
    result = method(obj, type_to_resolved)
```

Building the wrapper is not on the hot path: a recorded identity never reaches this line, so only the first call for a pair and the pairs that genuinely need `convert` pay for it. On my machine that is about 0.5 µs per call on those paths (conversion method registered: 0.73 → 1.22 µs; `Literal` target: 2.65 → 3.22 µs), with the identity path unchanged at 0.21 µs. Goes together with the import above and the fallback check below, which must then unwrap `method`. The PR description's paragraph on calling the method directly, and its second table, would need updating too.

src/plum/_promotion.py
```diff
@@ -43,9 +49,36 @@ def convert(obj: object, type_to: typeTypeTo) -> TypeTo:
...
+    # conversion method applied, so the fallback returned `obj` unchanged; and
+    # `is_faithful`, i.e. `isinstance(x, t) == issubclass(type(x), t)`, so that check
+    # depends only on `type_from`. A `Literal` or a custom `__instancecheck__` fails
+    # the second. Negatives are recorded too: `is_faithful` is not cached and would
+    # rerun on every call it cannot help.
+    if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
+        # `_convert._f` is the fallback that `_convert` wraps, so `method` being it
+        # means that no conversion method applied.
+        no_method_applied = method is _convert._f
+        cache[type_from, type_to] = no_method_applied and is_faithful(type_to_resolved)
```
wesselb: *Comment generated by Claude.*

Companion to the two comments above.

Shorten the explanation now that the top of the function covers it, and recognise the fallback through the wrapper that `invoke` returns:

```suggestion
    # Record whether the conversion is the identity: the fallback applied, and
    # `type_to` is faithful so that its instance check depends only on `type_from`.
    # Negatives are recorded too, because `is_faithful` is not cached and would
    # otherwise run on every call.
    if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
        # `_convert._f` is the fallback that `_convert` wraps.
        no_method_applied = _unwrap_invoked_methods(method) is _convert._f
        cache[type_from, type_to] = no_method_applied and is_faithful(type_to_resolved)
```

`invoke` records the resolved method on its wrapper as `__wrapped_by_plum__`, which is what `_unwrap_invoked_methods` follows; the check therefore keeps working after the rebase onto #288, where `invoke` returns an `_InvokedMethod` instead of a closure. Verified locally: the suite passes with this change on both the PR head and a test merge with master.

wesselb (commented):
*Review generated by Claude.*

Second round, after all first-round comments were applied and the pytest speed guard was added. I re-checked the soundness argument and the applied changes, and ran the suite on the PR head and on a test merge with master, which now compiles `_function` natively via #288: the merge is clean, so the rebase should be routine, but the mypyc job is worth re-running afterwards since the cache and the monkeypatched test then live in a compiled module. Two substantive requests: call `invoke` in `convert` instead of reaching into `_resolve_method_with_cache`, since the wrapper cost is not on the hot path, and open `convert` with a plain-language explanation of what the cache does. The rest is a sweep over the tests for plain language, including one test name that says "not recorded" where the test asserts a recorded `False`.

tests/conftest.py
```diff
@@ -40,6 +40,9 @@ def convert():
     _convert._pending = []
     _convert._resolved = resolved
     _convert.clear_cache(reregister=True)
+    # Restoring the methods changes which conversions are the identity, so the
+    # memo of identity conversions has to go with them.
```
wesselb: *Comment generated by Claude.*

Leftover vocabulary.

The last "memo" in the PR; the source calls these recorded identity conversions:

```suggestion
    # Restoring the methods changes which conversions are the identity, so the
    # recorded identity conversions have to go with them.
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,80 @@
+"""Speed guards for the fast paths that `tests/benchmark.py` reports on.
+
+These assert *ratios* between two dispatched calls timed in the same process, never
+absolute durations: a ratio cancels out machine speed, interpreter version and CI
+load, all of which move absolute timings by an order of magnitude.
+"""
```
wesselb: *Comment generated by Claude.*

Plain language.

"Speed guards" and "fast paths" are new terms; say what the file does:

```suggestion
"""Tests that guard against performance regressions.

These compare the timing of two dispatched calls in the same process. Such a ratio is
robust to machine speed, interpreter version, and CI load; an absolute duration is not.
"""
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+#: State                                Annotated  Union
+#: ===================================  =========  =====
+#: Both short-circuits (healthy)             ~1.2   ~1.2
+#: `_function._convert` fast path gone       ~2.3   ~2.3
+#: Cache never consulted (pre-#292)          ~7.4  ~23.3
+#: ===================================  =========  =====
+#:
+#: The bound sits between the first two rows, so either regression trips it, with
+#: enough margin on a ~2.5% spread that load cannot decide the outcome.
+MAX_ANNOTATED_RETURN_OVERHEAD = 1.75
```
wesselb: *Comment generated by Claude.*

Too much detail; non-standard comment form.

Keep the two numbers that set the bound and drop the table, the spread, and the `#:` form, which appears nowhere else in the tests:

```suggestion
# Cost of a dispatched call with a return annotation relative to one without. A
# satisfied return annotation is skipped in two places: `_function._convert` does not
# call `convert` at all for a recorded identity conversion, and `convert` returns
# early for one. With both in place the ratio is about 1.2; without the first it is
# about 2.3. The bound lies between the two, so losing either trips the test.
MAX_ANNOTATED_RETURN_OVERHEAD = 1.75
```

The measured figures are already drifting (1.08 and 1.27 here for the two cases; 1.16 to 1.24 under a compiled build), and the pre-PR row is in the PR description.

tests/test_benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+@plum.dispatch
+def _annotated_union(x: _R1) -> _R1 | _R2:
+    return x
+
+
+def _time(f, arg, *, n=20_000, repeat=5):
+    """Seconds per call to `f(arg)`, taking the fastest run.
+
+    The minimum is the robust statistic here: noise can only ever add time.
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention.

Imperative summary line:

```suggestion
    """Time `f(arg)` in seconds per call.

    The fastest of `repeat` runs is taken, since noise can only ever add time.
    """
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest.mark.benchmark
+@pytest.mark.parametrize("f", [_annotated, _annotated_union], ids=["plain", "union"])
+def test_annotated_return_is_not_much_slower_than_unannotated(f):
+    """A satisfied return annotation must stay nearly free.
+
+    Fails if the identity-conversion fast path stops being taken, and equally if it
+    is still taken but has become expensive to consult.
+    """
```
wesselb: *Comment generated by Claude.*

Plain language.

The second paragraph restates the constant's comment; one plain line suffices:

```suggestion
    """A return annotation that is already satisfied must add little to a call."""
```

tests/test_benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+    r = _R1()
+    overhead = _time(f, r) / _time(_unannotated, r)
+    assert overhead < MAX_ANNOTATED_RETURN_OVERHEAD, (
+        f"an annotated return cost {overhead:.1f}x an unannotated one, over the "
+        f"{MAX_ANNOTATED_RETURN_OVERHEAD}x bound; the identity-conversion fast path "
+        f"has likely regressed (see plum._function._identity_conversions)"
+    )
```
wesselb, line 63: *Comment generated by Claude.*

Message convention.

Complete sentences, offending values in backticks:

```suggestion
    assert overhead < MAX_ANNOTATED_RETURN_OVERHEAD, (
        f"An annotated return costs `{overhead:.1f}x` an unannotated one, over the "
        f"`{MAX_ANNOTATED_RETURN_OVERHEAD}x` bound. The identity-conversion fast path "
        f"in `plum._function._convert` has likely regressed."
    )
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
         add_conversion_method(int, str, str)
         add_conversion_method(float, str, str)
         assert promote(1, 1.0) == ("1", "1.0")
+
+
+# These pin the conditions that make the identity-conversion cache sound. They go
+# through a dispatched function, because that - not `plum.convert` - reads the cache.
```
wesselb: *Comment generated by Claude.*

Plain language; accuracy.

"Pin the conditions that make ... sound" is jargon, and `plum.convert` does read the cache (it returns early on a recorded identity); what only a dispatched call exercises is the skip in `plum._function._convert`:

```suggestion
# The tests below check the cache of identity conversions. They call a dispatched
# function with a return annotation, which is the path that the cache speeds up.
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+    """`plum._function._identity_conversions`, cleared before and after the test."""
+    plum._function._identity_conversions.clear()
+    yield plum._function._identity_conversions
+    plum._function._identity_conversions.clear()
+
+
+def test_identity_conversion_is_recorded_and_read(
+    identity_conversions, dispatch, monkeypatch
+):
+    """Once a pair is recorded, a repeat call must not reach `convert` at all."""
```
wesselb: *Comment generated by Claude.*

Plain language.

"Pair" needs context; say what is recorded:

```suggestion
    """Once a conversion is recorded as the identity, later calls skip `convert`."""
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+
+    obj = Sub()
+    assert f(obj) is obj
+    assert identity_conversions[Sub, Base] is True
+
+    add_conversion_method(Sub, Base, lambda _: "converted")
+    assert f(obj) == "converted"
+    # The re-analysed pair is no longer an identity, and the call keeps converting.
+    assert identity_conversions[Sub, Base] is False
+    assert f(obj) == "converted"
```
wesselb: *Comment generated by Claude.*

Unclear language.

"Must win regardless of registration order" does not say what the test checks. Name and describe it by what happens, and drop the comment that the docstring then covers:

```suggestion
def test_conversion_method_clears_recorded_identity(identity_conversions, dispatch):
    """Adding a conversion method clears a conversion recorded as the identity.

    The new method is then used, and the conversion is recorded as not the identity.
    """

    class Base:
        pass

    class Sub(Base):
        pass

    @dispatch
    def f(x: Sub) -> Base:
        return x

    obj = Sub()
    assert f(obj) is obj
    assert identity_conversions[Sub, Base] is True

    add_conversion_method(Sub, Base, lambda _: "converted")
    assert f(obj) == "converted"
    assert identity_conversions[Sub, Base] is False
    assert f(obj) == "converted"
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+        f(bad)
+
+    @dispatch
+    def g(x: int) -> typing.Literal[1]:
+        return x
+
+    assert g(1) == 1
+    assert identity_conversions[(int, typing.Literal[1])] is False
+    with pytest.raises(TypeError):
+        g(2)
```
wesselb: *Comment generated by Claude.*

Inaccurate name; unclear docstring.

The test asserts that these conversions *are* recorded, as not the identity, so "not recorded" is wrong; and "match on the value" and "share the key" need context. Also indexes without the tuple parentheses, as the other tests do:

```suggestion
def test_unfaithful_targets_are_not_identities(identity_conversions, dispatch):
    """Conversions to unfaithful types are recorded as not the identity.

    Whether an object is an instance of `Gate` or `Literal[1]` depends on the object
    and not just on its type, so the outcome cannot be cached by type.
    """

    class Meta(type):
        def __instancecheck__(cls, instance):
            return getattr(instance, "ok", False)

    class Gate(metaclass=Meta):
        pass

    class Thing:
        pass

    @dispatch
    def f(x: Thing) -> Gate:
        return x

    good, bad = Thing(), Thing()
    good.ok, bad.ok = True, False

    assert f(good) is good
    assert identity_conversions[Thing, Gate] is False
    with pytest.raises(TypeError):
        f(bad)

    @dispatch
    def g(x: int) -> typing.Literal[1]:
        return x

    assert g(1) == 1
    assert identity_conversions[int, typing.Literal[1]] is False
    with pytest.raises(TypeError):
        g(2)
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+    @dispatch
+    def f(x: float) -> Base:
+        return x
+
+    with pytest.raises(TypeError):
+        f(1.0)
+    # A raising conversion never reaches the recording step, so nothing is stored.
+    assert (float, Base) not in identity_conversions
+    with pytest.raises(TypeError):
+        f(1.0)
```
wesselb, line 323: *Comment generated by Claude.*

Plain language.

"Never reaches the recording step" describes the implementation; a docstring saying what is checked reads more simply:

```suggestion
def test_failed_conversions_are_not_recorded(identity_conversions, dispatch):
    """A conversion that raises is not recorded and keeps raising."""

    class Base:
        pass

    @dispatch
    def f(x: float) -> Base:
        return x

    with pytest.raises(TypeError):
        f(1.0)
    assert (float, Base) not in identity_conversions
    with pytest.raises(TypeError):
        f(1.0)
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+
+    @dispatch
+    def f(x: A | float) -> A | B:
+        return x
+
+    a = A()
+    assert f(a) is a
+    assert identity_conversions[A, A | B] is True
+    with pytest.raises(TypeError):
+        f(1.0)
```
wesselb: *Comment generated by Claude.*

Unclear docstring; re-tests `convert`.

"Recordable as a whole" is unclear, and the `float` leg only re-tests the fallback's type check through a dispatched function, which `test_convert` already covers; the `A | float` argument annotation exists only to reach it:

```suggestion
def test_union_targets_are_recorded(identity_conversions, dispatch):
    """A union of faithful types is faithful, so a conversion to it can be recorded."""

    class A:
        pass

    class B:
        pass

    @dispatch
    def f(x: A) -> A | B:
        return x

    a = A()
    assert f(a) is a
    assert identity_conversions[A, A | B] is True
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+    """Recording stops at the limit instead of retaining types forever."""
+    monkeypatch.setattr(plum._promotion, "_IDENTITY_CONVERSION_LIMIT", 3)
+
+    @dispatch
+    def f(x: object) -> object:
+        return x
+
+    types = [type(f"C{i}", (), {}) for i in range(10)]
+    for t in types:
+        assert f(t()) is not None
```
wesselb: *Comment generated by Claude.*

Plain language; unnecessary assertion.

Say what is checked, without the rationale, and drop the `is not None`:

```suggestion
    """Recording stops once the cache reaches the size limit."""
    monkeypatch.setattr(plum._promotion, "_IDENTITY_CONVERSION_LIMIT", 3)

    @dispatch
    def f(x: object) -> object:
        return x

    types = [type(f"C{i}", (), {}) for i in range(10)]
    for t in types:
        f(t())
```

tests/test_promotion.py
```diff
@@ -204,3 +204,192 @@ def test_self_promotion(convert, promote):
...
+    assert len(identity_conversions) == 3
+
+
+def test_unhashable_target_still_raises(identity_conversions, dispatch):
+    """An unhashable target raises, as it always has.
+
+    The cache hashes `(type(obj), type_to)`, but so does the method lookup it
+    replaces, so such a target has never been usable.
+    """
```
wesselb: *Comment generated by Claude.*

Unused fixtures; plain language.

`dispatch` is unused, and `identity_conversions` only clears a dict this test never writes to, since both calls raise first:

```suggestion
def test_unhashable_target_still_raises():
    """An unhashable `type_to` raises a `TypeError`, as it did before the cache.

    The cache lookup hashes `type_to`, but so did the method lookup already.
    """
```

wesselb (approved):
This also looks great, @nstarman! Thanks a lot. I think this is very clever and like this solution a lot!

Happy for you to merge after the comments have been resolved. :)

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested, and the comment at the recording step is shortened to match.

tests/conftest.py
nstarman: *Comment generated by Claude.*

Applied as suggested — that was indeed the last one; `grep -ri memo` over `src` and `tests` is now empty.

tests/test_benchmark.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_benchmark.py
nstarman: *Comment generated by Claude.*

Applied as suggested. Agreed on dropping the figures that drift — the two that set the bound are the only ones a future reader needs.

tests/test_benchmark.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_benchmark.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_benchmark.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested. The accuracy point is right — `convert` returns early on a recorded identity, so it does read the cache; what only a dispatched call exercises is the skip in `_function._convert`.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested. The name was indeed wrong — the test asserts a recorded `False`, which is a recorded conversion, not an absent one.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested. Confirmed the `float` leg was redundant: with it removed, the five mutation checks on the cache guards all still fail exactly as before, so it was not pinning anything of its own.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested. Kept the trailing `assert len(identity_conversions) == 3`, which is outside the suggested range and is what actually pins the bound — without it, removing the size check no longer fails any test.

tests/test_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Not applied — I measured it, and `invoke` costs more than the estimate. It builds a `functools.wraps` wrapper per call, which is ~1.5 µs here rather than ~0.5 µs, and that is enough to push two paths back above the **pre-PR** baseline.

Medians on this machine, `6840ebb` being the PR's merge base:

| Path | pre-PR | this PR | with `invoke` |
|---|---|---|---|
| identity | 3.67 µs | 0.49 | 0.50 |
| identity, union | 6.83 µs | 0.50 | 0.50 |
| `-> Any` (control) | 0.42 µs | 0.42 | 0.42 |
| conversion method | 2.93 µs | 1.60 | **3.15** |
| `Literal` (unfaithful) | 6.25 µs | 5.59 | **7.20** |

You are right that the hot path is unaffected — a recorded identity never reaches the line, and the first two rows are identical either way. But the last two rows are the ones the PR description's second table advertises as improvements (2.60 → 1.46 and 6.40 → 5.81 there). With `invoke` they become 7.5% and 15% *slower than master*, so the change would turn two claimed wins into regressions rather than costing a little on a cold path.

Three interleaved A/B runs, all within ~1%.

So I have kept the direct call, with a comment recording the trade-off explicitly:

```python
    # Resolve and call the method directly. `_convert.invoke` would read better, but
    # it builds a `functools.wraps` wrapper per call, which costs ~1.5 us on every
    # call that does not hit the cache above.
```

Everything else in this cluster is applied: the plain-language opener at the top of `convert`, and the shortened comment at the recording step. Happy to take `invoke` anyway if you would rather have the clarity and drop those two rows from the description — your call, and it is a one-line change either way.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Not applied, as a consequence of keeping the direct call (see my reply on the `invoke` comment — it measured ~1.5 µs per cache-missing call, enough to put two paths below the pre-PR baseline). With the direct call, `method` is already the resolved method, so `method is _convert._f` works without unwrapping and the import is unused.

Your robustness point is still correct and I checked it: master's `_InvokedMethod.__init__` does set `__wrapped_by_plum__`, and `_unwrap_invoked_methods` follows it, so the unwrap-based check would survive the rebase onto #288. It is only unnecessary here because nothing wraps the method. If you would rather have `invoke`, this import comes back with it.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

The comment rewrite is applied as suggested. The `_unwrap_invoked_methods` line is not, since the direct call is kept — see my reply on the `invoke` comment for the measurements. It reads:

```python
    if known is None and len(cache) < _IDENTITY_CONVERSION_LIMIT:
        # `_convert._f` is the fallback that `_convert` wraps.
        no_method_applied = method is _convert._f
        cache[type_from, type_to] = no_method_applied and is_faithful(type_to_resolved)
```

nstarman:
*Comment generated by Claude.*

Thanks for the second pass — all 20 comments addressed in `f18f382`, 18 applied as suggested.

**One not applied: `_convert.invoke`.** You flagged it as costing ~0.5 µs off the hot path, which would have been fine. Measured here it is ~1.5 µs per cache-missing call, and that is enough to matter:

| Path | pre-PR `6840ebb` | this PR | with `invoke` |
|---|---|---|---|
| identity | 3.67 µs | 0.49 | 0.50 |
| identity, union | 6.83 µs | 0.50 | 0.50 |
| `-> Any` (control) | 0.42 µs | 0.42 | 0.42 |
| conversion method | 2.93 µs | 1.60 | **3.15** |
| `Literal` (unfaithful) | 6.25 µs | 5.59 | **7.20** |

Your reasoning about the hot path is right — the first two rows are identical either way. But the last two are the ones the description's second table advertises as improvements; with `invoke` they end up 7.5% and 15% slower than master, so it converts two claimed wins into regressions. Three interleaved A/B runs, all within ~1%. The direct call stays, now with a comment recording the trade-off, and the `_unwrap_invoked_methods` import goes with it. Say the word if you would rather have the clarity and drop those two rows from the description — it is a one-line change.

I did verify your robustness point independently: master's `_InvokedMethod` does set `__wrapped_by_plum__`, so the unwrap-based check would survive the rebase onto #288 if we go that way.

**Everything else applied,** including both renames — you were right that `test_unfaithful_targets_are_not_recorded` asserted a recorded `False`, so the name said the opposite of the test.

**Verification.** The tests changed a fair amount, so I re-ran the mutation checks rather than assume: each of the five cache guards still fails at least one test when removed (`is_faithful` gate, fallback-method check, `add_conversion_method` invalidation, `clear_all_cache` hook, size bound). That also confirms the dropped `float` leg of the union test was not pinning anything. I kept the trailing `assert len(identity_conversions) == 3` in the bounded test — it sits just outside your suggested range and is the assertion that actually pins the limit.

173 tests pass (the one failure, `test_methodlist_repr`, fails identically on master); `ruff`, `ruff format` and `mypy --enable-error-code=no-redef` clean; benchmarks byte-for-byte unchanged from the previous head on all five paths.

On the rebase: agreed the mypyc job is worth re-running afterwards. Noting for whoever does it that `_IDENTITY_CONVERSION_LIMIT` deliberately lives in the uncompiled `_promotion` so `test_recorded_identities_are_bounded` can still monkeypatch it under a compiled build.

nstarman:
#300 should actually come before this PR so we can clean up some of the code here to use the sped-up `invoke`.

PR #296, opened 2026-08-11
fix: Any must not collide with concrete types under beartype>=0.23

### LINKED ISSUE #295: beartype 0.23's intentional `is_subhint(Any, T) == True` breaks plum's method-redefinition detection, silently deleting registered methods
# `beartype` 0.23.0's intentional `is_subhint(Any, T) == True` change breaks `plum`'s method-redefinition detection, silently deleting registered methods (incl. plum's own identity-conversion fallback)

## Summary

Since `beartype` 0.23.0rc0, `beartype.door.TypeHint(Any) == beartype.door.TypeHint(X)` returns `True` for **every** `X`, not just `Any` itself. `plum` uses `TypeHint.__eq__` to detect "this new method redefines an existing one" in `Resolver.register()`. Because unannotated parameters resolve to `Any`, the very first unannotated method registered on any `plum.Dispatcher` (or `Function`) is now silently **overwritten** by the next method registered with concrete types — no warning, no error, just gone.

This isn't a beartype bug: it's [intentional, upstream-endorsed behavior](https://github.com/beartype/beartype/issues/530) (mypy's `Any`-assignability semantics, merged in [beartype#616](https://github.com/beartype/beartype/pull/616)). `is_subhint` deliberately abandoned antisymmetry for `Any` (`is_subhint(Any, T)` and `is_subhint(T, Any)` are now both `True` for all `T`, which is correct for *assignability* but means `Any` is no longer usable as an operand in `plum`'s `A<=B and B<=A ⇒ A==B` syllogism). The fix belongs in `plum`.

## Minimal repro (no jax, no quax — pure `plum` + `beartype`)

```python
# pip install beartype==0.23.0rc0 plum-dispatch==2.9.0
import plum
plum.convert(1, int)
```

```
plum._resolver.NotFoundLookupError: `_convert(int, int)` could not be resolved.
```

Downgrading to `beartype==0.22.9` (with the identical `plum-dispatch==2.9.0`) fixes it. This is the *only* difference between the two runs.

## Root cause chain

1. `plum/_promotion.py` registers, at import time, an unannotated identity-conversion fallback:
   ```python
   @_dispatch
   def _convert(obj, type_to):
       if not is_bearable(obj, resolve_type_hint(type_to)):
           raise TypeError(...)
       return obj
   ```
   Unannotated parameters resolve to `typing.Any` ([`_signature.py:370-371`](https://github.com/beartype/plum/blob/master/plum/_signature.py#L370-L371)), so this registers as `Signature(Any, Any)`.

2. Immediately afterward, `plum/_promotion.py` registers several concretely-typed converters via `add_conversion_method`, e.g. `add_conversion_method(object, tuple, ...)` → `Signature(object, tuple)`.

3. `Resolver.register()` ([`_resolver.py:298-332`](https://github.com/beartype/plum/blob/master/plum/_resolver.py#L298-L332)) checks whether the new signature already exists via `Signature.__eq__` to decide redefinition-vs-append:
   ```python
   existing = [m.signature == signature for m in self.methods]
   if any(existing):
       self.methods[existing.index(True)] = method   # <-- overwrite in place
   else:
       self.methods.append(method)
   ```

4. `Signature.__eq__` ([`_signature.py:125-139`](https://github.com/beartype/plum/blob/master/plum/_signature.py#L125-L139)) compares each parameter pairwise via `beartype.door.TypeHint(t) == beartype.door.TypeHint(y)`.

5. Under `beartype>=0.23.0rc0`, `TypeHint(Any) == TypeHint(tuple)` is `True` (see below), so `Signature(object, tuple) == Signature(Any, Any)` evaluates `True`. Step 3's `register()` therefore treats the `(object, tuple)` conversion method as a "redefinition" of the `(Any, Any)` identity fallback and **overwrites it in place** — with no `MethodRedefinitionWarning`, because `warn_redefinition` defaults to `False`.

6. Every subsequent `plum.convert(x, T)` call for a `T`/`x` pair not covered by a specific, narrower converter now raises `NotFoundLookupError`, because the universal identity fallback no longer exists in `self.methods`.

We hit this via [`quax`](https://github.com/nstarman/quax) (`plum.convert(jnp.asarray(1.0), jax.Array)` fails the same way — [CI run](https://github.com/nstarman/quax/actions/runs/31364324486)), but the bug has nothing to do with `jax`; the pure-`plum` repro above shows the fallback is gone as soon as `plum` itself finishes importing.

### Why `TypeHint(Any) == TypeHint(X)` is now `True`

`TypeHint._is_equal` ([`doorsuper.py`](https://github.com/beartype/beartype/blob/main/beartype/door/_cls/doorsuper.py)) implements equality via the antisymmetric syllogism `A<=B and B<=A ⇒ A==B`, delegating to `is_subhint`. [beartype#616](https://github.com/beartype/beartype/pull/616) (closing [beartype#530](https://github.com/beartype/beartype/issues/530)) added:

```python
return (
    self._hint is Any or   # <-- new in 0.23.0rc0
    other._hint is Any or  # already existed
    ...
)
```

`other._hint is Any` (pre-existing) correctly implements "`X` is assignable to `Any`" (`X <: Any`, true for all `X`). The new `self._hint is Any` implements "`Any` is assignable to `X`" (`Any <: X`, also true for all `X`, per mypy's assignability rules for `Any` cited in beartype#530). Both are correct *assignability* facts individually — but combined, they make `is_subhint` no longer antisymmetric for `Any`, which breaks the syllogism plum's `_is_equal` depends on. This is called out explicitly, even in `AnyTypeHint._is_subhint_branch`'s own docstring, which still says (now contradicted by the superclass short-circuit): *"typing.Any is a subhint of NO hint other than itself."*

## Why this belongs in `plum`, not `beartype`

`beartype`'s `is_subhint(Any, T) == True` is a deliberate, upstream-endorsed semantic (assignability, matching mypy — see beartype#530/#616), not a defect. `plum` is the one relying on an invariant (antisymmetry ⇒ well-defined equality) that `is_subhint` never actually promised to provide for gradually-typed hints like `Any`. Concretely, `plum` conflates two different notions that need to stay separate:

- **Dispatch matching** (`Signature.match`, `resolve()`'s `<=` check): "does this call satisfy this parameter's hint" — `is_subhint`/assignability semantics are exactly right here.
- **Redefinition/identity detection** (`Resolver.register()`'s `existing = [...]`, `Signature.__eq__`): "is this literally the same registered slot as an existing method" — this needs a stricter, syntactic notion of hint equality, where `Any` only collides with `Any`.

## Suggested fix

In `Signature.__eq__` (or wherever `Resolver.register()` determines "redefinition"), don't derive structural/registration equality from `is_subhint`-based semantic equivalence when either side is `Any` (or, more generally, don't assume `TypeHint.__eq__` is a syntactic identity check going forward — it explicitly isn't anymore for `Any`). A couple of options:

1. Fast-path: in `Resolver.register()`'s existing-check (or in a new comparison used specifically there), treat two parameter hints as "the same slot" only if `x is Any) == (y is Any)` and (both `Any`, or `TypeHint(x) == TypeHint(y)`) — i.e. require literal `Any`-ness to match rather than letting `Any` collapse onto concrete types.
2. Alternatively, compare the raw underlying hints (`self._hint == other._hint`, pre-DOOR) for redefinition purposes, and reserve `beartype.door.TypeHint` comparisons for actual dispatch resolution where assignability semantics are wanted.

Either way, this needs a `plum`-side fix (and probably a regression test pinning `beartype>=0.23.0` compatibility), since the `beartype` change is intentional and unlikely to be reverted.

## Environment

- `plum-dispatch==2.9.0`
- `beartype==0.23.0rc0` (breaks) vs `beartype==0.22.9` (works) — only difference between the two environments
- Reproduced on macOS arm64, Python 3.13.3; the same failure occurs on Linux x86_64 / Python 3.11, 3.13, 3.14 in [quax's CI](https://github.com/nstarman/quax/actions/runs/31364324486) (jax 0.10.2, jaxlib 0.10.2, same plum/beartype versions)

## AI

Created by Claude Opus 5

nstarman:
Fixes #295.

## Problem

Since `beartype` 0.23.0rc0, `is_subhint(Any, T)` is `True` for every `T` — a deliberate change (beartype#616, closing beartype#530) matching mypy's two-way assignability semantics for `Any`. `TypeHint.__eq__` is built from the antisymmetric syllogism `is_subhint(a, b) and is_subhint(b, a) ⇒ a == b`, so this also makes `beartype.door.TypeHint(Any) == TypeHint(T)` `True` for every `T`.

`plum` relies on `TypeHint` equality/subhint checks for two different things that need to stay separate:

- **Dispatch matching** (`Signature.match`, `resolve()`'s value-level checks via `is_bearable`) — assignability semantics are correct here, and are untouched by this PR.
- **Bookkeeping**: "is this the same registered signature as an existing one" (`Resolver.register()`'s redefinition detection, via `Signature.__eq__`) and "is this signature strictly more specific than that one" (`Signature.__le__`'s specificity ordering) — these need `Any` to behave as a distinct, least-specific type, not something that collapses onto (or below) arbitrary concrete types.

Since unannotated parameters resolve to `Any` (`_extract_signature`), the bug surfaces two ways:

1. **Registration silently overwrites the wrong method.** `Resolver.register()` treats a newly-registered concretely-typed method as "redefining" any earlier unannotated method (or vice versa), regardless of registration order, and clobbers it in place — with no `MethodRedefinitionWarning` (`warn_redefinition` defaults to `False`). This is severe enough to destroy plum's own built-in identity-conversion fallback in `_promotion.py` the moment `plum` finishes importing, breaking `plum.convert()` globally for any type without a specific registered converter:
   ```python
   # beartype==0.23.0rc0, plum-dispatch==2.9.0
   import plum
   plum.convert(1, int)
   # plum._resolver.NotFoundLookupError: `_convert(int, int)` could not be resolved.
   ```

2. **Dispatch specificity becomes registration-order-dependent.** A generic `f(x)` fallback and a specific `f(x: int)` overload become mutually "comparable but neither strictly more specific" (`Sig(int) <= Sig(Any)` and `Sig(Any) <= Sig(int)` both `True`), so whichever was registered *last* wins arbitrarily:
   ```python
   import plum

   @plum.dispatch
   def g(x: int):
       return "int"

   @plum.dispatch
   def g(x):
       return "any"

   g(1)  # "any" -- wrong; should be "int"
   ```
   (Registering `g(x)` before `g(x: int)` instead gives the "correct" answer only by accident, because the earlier bug already destroyed one of the two methods on registration.)

Both were originally found via [quax](https://github.com/nstarman/quax), where `plum.convert(jnp.asarray(1.0), jax.Array)` started failing the same way in its "pre-release deps" CI job. Neither reproduction involves `jax`/`quax` at all — see #295 for the full trace.

## Fix

Add two small helpers in `_type.py`:

- `_type_hints_equal(x, y)` — like `TypeHint(x) == TypeHint(y)`, except `Any` is considered equal only to `Any` itself.
- `_type_hint_le(x, y)` — like `TypeHint(x) <= TypeHint(y)`, except `Any` is only a subhint of `Any` itself (so `X <= Any` still holds for all `X`, preserving `Any` as the top/least-specific type, but `Any <= T` no longer holds for concrete `T`).

Both are thin wrappers with no behavior change for anything not involving `Any`, so `beartype`'s (correct, upstream-endorsed) assignability semantics for `Any` are left completely untouched everywhere else — including `Signature.match`/`is_bearable`, plum's own public `plum.issubclass`/`plum.isinstance`, and ordinary type-hint-equivalence cases like `Union[int, bool] == int` (still passes, see `test_equality` in `test_signature.py`).

Use these in the three spots that need bookkeeping-equality rather than assignability:
- `Signature.__eq__` (types + varargs)
- `Signature.__le__`'s two branches (types + varargs, in both the equality fast-path and the subhint fallback)
- `add_promotion_rule`'s reverse-rule-skip check in `_promotion.py`

## Testing

- Added `test_type_hints_equal`/`test_type_hint_le` in `tests/test_type.py`, and `test_register_any_never_collides_with_concrete_type` (parametrized over registration order) in `tests/test_resolver.py`, all directly pinning the scenarios above.
- Ran the **existing** test suite under `beartype==0.23.0rc0` before this fix: **29 tests fail** (`plum._resolver.NotFoundLookupError` and similar) that all pass on `beartype<0.23`. After this fix, under `beartype==0.23.0rc0`: **200 passed**, only the one pre-existing, unrelated `test_methodlist_repr` failure remains (a path-length rendering artifact in my local checkout path, reproduces identically on unmodified `master` regardless of `beartype` version).
- Confirmed identical (200 passed, same one unrelated failure) on `beartype<0.23` too — no regression for the currently-released `beartype`.
- `ruff check`/`ruff format --check` clean; `mypy` error count unchanged (23, matching unmodified `master`) after wrapping the two new comparisons in `bool(...)` to satisfy `no-any-return`.

Happy to adjust naming/placement or split into a smaller PR (e.g. just the `__eq__`/registration fix) if you'd prefer to review the specificity-ordering fix separately — I bundled them because they share the exact same root cause and helper functions.

### DIFF
--- a/src/plum/_parametric.py
+++ b/src/plum/_parametric.py
@@ -17,7 +17,7 @@ from beartype.roar import BeartypeDoorNonpepException
 
 from ._dispatcher import Dispatcher
 from ._function import _owner_transfer
-from ._type import resolve_type_hint
+from ._type import _type_hint_le, resolve_type_hint
 from ._util import TypeHint
 from .repr import repr_short
 
@@ -123,9 +123,11 @@ class ParametricTypeMeta(type):
 
 def _default_le_type_par(p_left: TypeHint | object, p_right: TypeHint | object) -> bool:
     if is_type(p_left) and is_type(p_right):
-        p_left = beartype.door.TypeHint(resolve_type_hint(p_left))
-        p_right = beartype.door.TypeHint(resolve_type_hint(p_right))
-        return p_left <= p_right
+        # `_type_hint_le` rather than comparing `TypeHint`s directly, so that a type
+        # parameter of `Any` stays distinct from a concrete one; see its docstring.
+        # Without this, `P[Any]` and `P[int]` collide under `beartype>=0.23` exactly
+        # as `Any` and `int` do. See https://github.com/beartype/plum/issues/295.
+        return _type_hint_le(resolve_type_hint(p_left), resolve_type_hint(p_right))
     else:
         return p_left == p_right
 
--- a/src/plum/_parametric.py
+++ b/src/plum/_parametric.py
@@ -123,10 +123,8 @@ class ParametricTypeMeta(type):
 
 def _default_le_type_par(p_left: TypeHint | object, p_right: TypeHint | object) -> bool:
     if is_type(p_left) and is_type(p_right):
-        # `_type_hint_le` rather than comparing `TypeHint`s directly, so that a type
+        # Use `_type_hint_le` rather than comparing `TypeHint`s directly so that a type
         # parameter of `Any` stays distinct from a concrete one; see its docstring.
-        # Without this, `P[Any]` and `P[int]` collide under `beartype>=0.23` exactly
-        # as `Any` and `int` do. See https://github.com/beartype/plum/issues/295.
         return _type_hint_le(resolve_type_hint(p_left), resolve_type_hint(p_right))
     else:
         return p_left == p_right
--- a/src/plum/_promotion.py
+++ b/src/plum/_promotion.py
@@ -16,7 +16,7 @@ from beartype.door import TypeHint
 import plum._function
 from ._bear import is_bearable
 from ._dispatcher import Dispatcher
-from ._type import _type_hints_equal, resolve_type_hint
+from ._type import _type_hint_eq, resolve_type_hint
 from .repr import repr_short
 
 T = TypeVar("T")
@@ -149,9 +149,9 @@ def add_promotion_rule(type1: object, type2: object, type_to: object) -> None:
 
     # If the types are the same, we don't need to add the reverse rule. Resolve the
     # types to handle the case where types are equal, but not identical. Use
-    # `_type_hints_equal` rather than comparing `TypeHint`s directly so that `Any`
-    # is never mistaken for an unrelated concrete type (see `_type_hints_equal`).
-    if _type_hints_equal(resolve_type_hint(type1), resolve_type_hint(type2)):
+    # `_type_hint_eq` rather than comparing `TypeHint`s directly so that `Any`
+    # is never mistaken for an unrelated concrete type.
+    if _type_hint_eq(resolve_type_hint(type1), resolve_type_hint(type2)):
         return  # Escape early.
 
     @_promotion_rule.dispatch
--- a/src/plum/_promotion.py
+++ b/src/plum/_promotion.py
@@ -16,7 +16,7 @@ from beartype.door import TypeHint
 import plum._function
 from ._bear import is_bearable
 from ._dispatcher import Dispatcher
-from ._type import resolve_type_hint
+from ._type import _type_hints_equal, resolve_type_hint
 from .repr import repr_short
 
 T = TypeVar("T")
@@ -148,8 +148,10 @@ def add_promotion_rule(type1: object, type2: object, type_to: object) -> None:
         return type_to
 
     # If the types are the same, we don't need to add the reverse rule. Resolve the
-    # types to handle the case where types are equal, but not identical.
-    if TypeHint(resolve_type_hint(type1)) == TypeHint(resolve_type_hint(type2)):
+    # types to handle the case where types are equal, but not identical. Use
+    # `_type_hints_equal` rather than comparing `TypeHint`s directly so that `Any`
+    # is never mistaken for an unrelated concrete type (see `_type_hints_equal`).
+    if _type_hints_equal(resolve_type_hint(type1), resolve_type_hint(type2)):
         return  # Escape early.
 
     @_promotion_rule.dispatch
--- a/src/plum/_signature.py
+++ b/src/plum/_signature.py
@@ -15,7 +15,7 @@ from rich.segment import Segment
 from beartype.peps import resolve_pep563 as beartype_resolve_pep563
 
 from ._bear import is_bearable
-from ._type import _type_hint_le, _type_hints_equal, is_faithful, resolve_type_hint
+from ._type import _type_hint_eq, _type_hint_le, is_faithful, resolve_type_hint
 from ._util import (
     Comparable,
     Missing,
@@ -131,18 +131,20 @@ class Signature(Comparable):
         # We don't need to check faithfulness, because that is automatically
         # derived from the arguments.
         #
-        # We use `_type_hints_equal` rather than comparing `TypeHintWrapper`s
-        # directly, because `beartype.door.TypeHint(Any) == TypeHint(T)` is `True`
-        # for every `T` (see `_type_hints_equal`'s docstring). Comparing directly
-        # would make two signatures with unrelated concrete types spuriously equal
-        # whenever one of them has an unannotated (and hence `Any`-typed) parameter.
+        # Use `_type_hint_eq` rather than comparing `TypeHintWrapper`s directly
+        # so that `Any` only equals `Any` itself; see its docstring.
         types_equal = all(
-            _type_hints_equal(x, y)
-            for x, y in zip(self.types, other.types, strict=True)
+            _type_hint_eq(x, y) for x, y in zip(self.types, other.types, strict=True)
         )
-        varargs_equal = (self.varargs is Missing) == (other.varargs is Missing) and (
-            self.varargs is Missing or _type_hints_equal(self.varargs, other.varargs)
-        )
+        if self.varargs is Missing:
+            # Both must be missing.
+            varargs_equal = other.varargs is Missing
+        elif other.varargs is Missing:
+            # Neither must be missing.
+            varargs_equal = False
+        else:
+            # And they must be equal.
+            varargs_equal = _type_hint_eq(self.varargs, other.varargs)
         return types_equal and varargs_equal and self.precedence == other.precedence
 
     def __hash__(self) -> int:
@@ -180,14 +182,10 @@ class Signature(Comparable):
         # one place.
         self_types = self.expand_varargs(len(other.types))
         other_types = other.expand_varargs(len(self.types))
-        # As in `__eq__`, use `_type_hints_equal` here rather than comparing
-        # `TypeHintWrapper`s directly, so an unannotated (`Any`-typed) parameter
-        # is never mistaken for an unrelated concrete type.
+        # As in `__eq__`, use `_type_hint_eq` so that `Any` only equals `Any`
+        # itself; see its docstring.
         if all(
-            [
-                _type_hints_equal(x, y)
-                for x, y in zip(self_types, other_types, strict=True)
-            ]
+            [_type_hint_eq(x, y) for x, y in zip(self_types, other_types, strict=True)]
         ):
             if self.has_varargs and other.has_varargs:
                 return _type_hint_le(self.varargs, other.varargs)
@@ -201,10 +199,8 @@ class Signature(Comparable):
             else:
                 return True
 
-        # As above, use `_type_hint_le` rather than comparing `TypeHintWrapper`s
-        # directly: `beartype.door.is_subhint(Any, T)` is `True` for every `T`
-        # since `beartype` 0.23, which would make an unannotated (`Any`-typed)
-        # parameter appear exactly as specific as any concrete overload.
+        # As above, use `_type_hint_le` so that `Any` is only a subhint of `Any`
+        # itself; see its docstring.
         elif all(
             [_type_hint_le(x, y) for x, y in zip(self_types, other_types, strict=True)]
         ):
--- a/src/plum/_signature.py
+++ b/src/plum/_signature.py
@@ -185,7 +185,7 @@ class Signature(Comparable):
         # As in `__eq__`, use `_type_hint_eq` so that `Any` only equals `Any`
         # itself; see its docstring.
         if all(
-            [_type_hint_eq(x, y) for x, y in zip(self_types, other_types, strict=True)]
+            _type_hint_eq(x, y) for x, y in zip(self_types, other_types, strict=True)
         ):
             if self.has_varargs and other.has_varargs:
                 return _type_hint_le(self.varargs, other.varargs)
@@ -202,7 +202,7 @@ class Signature(Comparable):
         # As above, use `_type_hint_le` so that `Any` is only a subhint of `Any`
         # itself; see its docstring.
         elif all(
-            [_type_hint_le(x, y) for x, y in zip(self_types, other_types, strict=True)]
+            _type_hint_le(x, y) for x, y in zip(self_types, other_types, strict=True)
         ):
             # In this case, we have that `other >= self` is `False`, so returning `True`
             # gives that `other < self` and returning `False` gives that `other` cannot
--- a/src/plum/_signature.py
+++ b/src/plum/_signature.py
@@ -12,11 +12,10 @@ from typing_extensions import Self
 from rich.console import Console, ConsoleOptions
 from rich.segment import Segment
 
-from beartype.door import TypeHint as TypeHintWrapper
 from beartype.peps import resolve_pep563 as beartype_resolve_pep563
 
 from ._bear import is_bearable
-from ._type import is_faithful, resolve_type_hint
+from ._type import _type_hint_le, _type_hints_equal, is_faithful, resolve_type_hint
 from ._util import (
     Comparable,
     Missing,
@@ -126,17 +125,25 @@ class Signature(Comparable):
         if not isinstance(other, Signature):
             return False
 
+        if len(self.types) != len(other.types):
+            return False
+
         # We don't need to check faithfulness, because that is automatically
         # derived from the arguments.
-        return (
-            tuple(TypeHintWrapper(t) for t in self.types),
-            Missing if self.varargs is Missing else TypeHintWrapper(self.varargs),
-            self.precedence,
-        ) == (
-            tuple(TypeHintWrapper(t) for t in other.types),
-            Missing if other.varargs is Missing else TypeHintWrapper(other.varargs),
-            other.precedence,
+        #
+        # We use `_type_hints_equal` rather than comparing `TypeHintWrapper`s
+        # directly, because `beartype.door.TypeHint(Any) == TypeHint(T)` is `True`
+        # for every `T` (see `_type_hints_equal`'s docstring). Comparing directly
+        # would make two signatures with unrelated concrete types spuriously equal
+        # whenever one of them has an unannotated (and hence `Any`-typed) parameter.
+        types_equal = all(
+            _type_hints_equal(x, y)
+            for x, y in zip(self.types, other.types, strict=True)
         )
+        varargs_equal = (self.varargs is Missing) == (other.varargs is Missing) and (
+            self.varargs is Missing or _type_hints_equal(self.varargs, other.varargs)
+        )
+        return types_equal and varargs_equal and self.precedence == other.precedence
 
     def __hash__(self) -> int:
         return hash((Signature, *self.types, self.varargs))
@@ -173,16 +180,17 @@ class Signature(Comparable):
         # one place.
         self_types = self.expand_varargs(len(other.types))
         other_types = other.expand_varargs(len(self.types))
+        # As in `__eq__`, use `_type_hints_equal` here rather than comparing
+        # `TypeHintWrapper`s directly, so an unannotated (`Any`-typed) parameter
+        # is never mistaken for an unrelated concrete type.
         if all(
             [
-                TypeHintWrapper(x) == TypeHintWrapper(y)
+                _type_hints_equal(x, y)
                 for x, y in zip(self_types, other_types, strict=True)
             ]
         ):
             if self.has_varargs and other.has_varargs:
-                self_varargs = TypeHintWrapper(self.varargs)
-                other_varargs = TypeHintWrapper(other.varargs)
-                return bool(self_varargs <= other_varargs)
+                return _type_hint_le(self.varargs, other.varargs)
 
             # Having variable arguments makes you slightly larger.
             elif self.has_varargs:
@@ -193,11 +201,12 @@ class Signature(Comparable):
             else:
                 return True
 
+        # As above, use `_type_hint_le` rather than comparing `TypeHintWrapper`s
+        # directly: `beartype.door.is_subhint(Any, T)` is `True` for every `T`
+        # since `beartype` 0.23, which would make an unannotated (`Any`-typed)
+        # parameter appear exactly as specific as any concrete overload.
         elif all(
-            [
-                TypeHintWrapper(x) <= TypeHintWrapper(y)
-                for x, y in zip(self_types, other_types, strict=True)
-            ]
+            [_type_hint_le(x, y) for x, y in zip(self_types, other_types, strict=True)]
         ):
             # In this case, we have that `other >= self` is `False`, so returning `True`
             # gives that `other < self` and returning `False` gives that `other` cannot
@@ -211,9 +220,7 @@ class Signature(Comparable):
                 #       is `1.0`, then reasonably the variable arguments should be
                 #       ignored and `(int, *A)` should be considered more specific
                 #       than `(Number, *B)`.
-                self_varargs = TypeHintWrapper(self.varargs)
-                other_varargs = TypeHintWrapper(other.varargs)
-                return bool(self_varargs <= other_varargs)
+                return _type_hint_le(self.varargs, other.varargs)
 
             elif self.has_varargs:
                 # Previously, this returned `False`, which would implement the subset
--- a/src/plum/_type.py
+++ b/src/plum/_type.py
@@ -19,7 +19,6 @@ from typing import (
     Literal,
     TypeGuard,
     TypeVar,
-    Union,
     cast,
     final,
     get_args,
@@ -294,117 +293,84 @@ def resolve_type_hint(x: object, /) -> object:
     return x
 
 
-@lru_cache(maxsize=4096)
-def _substitute_nested_any(hint: object, /) -> object:
-    """Replace every `Any` strictly below the root of `hint` with `object`.
+def _substitute_any(hint: object, /) -> object:
+    """Replace every `Any` in `hint` with `object`.
 
-    `_type_hint_le` handles a root `Any` itself; this handles the parameters of a
-    hint, where `beartype>=0.23` collapses `list[Any]` onto `list[int]` in exactly
-    the way it collapses `Any` onto `int`. Substituting `object` hands the whole
-    comparison, variance included, back to `beartype`: `object` is a subhint of
-    nothing but itself, and everything is a subhint of it, which is precisely the
-    ordering `Any` is meant to have.
-
-    A hint that cannot be rebuilt is returned unchanged, leaving `beartype`'s own
-    semantics in place rather than failing the comparison.
-
-    `_type_hint_le`/`_type_hint_eq` call this on every non-`Any` signature
-    comparison, i.e. on essentially every dispatch resolution and method
-    registration -- but the set of distinct type hints in a program is small and
-    fixed (the annotations written in its `@dispatch`-decorated signatures), while
-    the number of *comparisons* between them is not: the same hints get walked
-    over and over. `lru_cache` turns that from "recurse and rebuild every time"
-    into "recurse once per distinct hint, then a dict lookup", including for
-    hints nested inside others (`int` inside `list[int]` inside `dict[str,
-    list[int]]`, ...) -- self-recursive calls hit the cache too. Hints reachable
-    here are already required to be hashable elsewhere (`Signature.__hash__`
-    hashes them directly), so this adds no new constraint.
+    For Plum's signature bookkeeping, `object` has exactly the ordering that `Any`
+    should have: it is a subhint of nothing but itself, and everything is a subhint
+    of it. Rewriting `Any` to `object` therefore leaves nesting and variance to
+    `beartype`. A hint that cannot be rebuilt is returned unchanged.
 
     Args:
         hint (object): Already-resolved type hint.
 
     Returns:
-        object: `hint` with every nested `Any` replaced by `object`.
+        object: `hint` with every `Any` replaced by `object`.
     """
+    # The same hints are compared over and over, so cache the rewrite. Not every hint
+    # is hashable, e.g. `Annotated[int, {"a": 1}]`, so only cache when possible.
+    if _hashable(hint):
+        return _substitute_any_cached(hint)
+    return _substitute_any_uncached(hint)
+
+
+def _substitute_any_uncached(hint: object, /) -> object:
+    if hint is Any:
+        return object
+    if isinstance(hint, list):
+        # The parameters of a `Callable`.
+        return [_substitute_any(arg) for arg in hint]
     origin = get_origin(hint)
     # `Literal`'s arguments are values, not hints, so they must not be rewritten.
     if origin is None or origin is Literal:
         return hint
     args = get_args(hint)
-    if not args:
-        return hint
     new_args = tuple(_substitute_any(arg) for arg in args)
     if new_args == args:
         return hint
     try:
-        if origin is Union or origin is UnionType:
-            # `reduce(or_, ...)` rather than `Union[...]`: the arguments are only
-            # known at runtime, and `X | Y` is the modern spelling.
+        if origin is UnionType:
+            # `types.UnionType` cannot be subscripted.
             return reduce(or_, new_args)
-        if origin is Callable:
-            # `Callable`'s arguments are `(parameters, return)`, where `parameters`
-            # is a list or `...`, so it does not rebuild by subscripting a tuple.
-            return Callable[new_args[0], new_args[1]]  # pyright: ignore[reportInvalidTypeForm]
         return origin[new_args]
-    except Exception:  # noqa: BLE001
+    except Exception:
         return hint
 
 
-def _substitute_any(arg: object, /) -> object:
-    """Replace `Any` with `object` in a single argument of a type hint."""
-    if arg is Any:
-        return object
-    if isinstance(arg, list):
-        # The parameter list of a `Callable`.
-        return [_substitute_any(x) for x in arg]
-    if arg is Ellipsis or arg is None:
-        return arg
-    return _substitute_nested_any(arg)
+_substitute_any_cached = lru_cache(maxsize=4096)(_substitute_any_uncached)
 
 
 def _type_hint_le(x: object, y: object, /) -> bool:
     """Check whether `x` is a subhint of `y`, where `Any` is only a subhint of itself.
 
-    This check is used for Plum's signature bookkeeping: redefinition detection
-    (`Signature.__eq__`, via `_type_hint_eq`) and specificity ordering
-    (`Signature.__le__`). It differs from `beartype.door.TypeHint(x) <= TypeHint(y)`
-    in exactly one respect: `Any` is only a subhint of `Any` itself, preserving `Any`
-    as Plum's unique least specific type. Since `beartype` 0.23, `is_subhint(Any, T)`
-    is `True` for every `T`, which would otherwise make an unannotated (`Any`-typed)
-    parameter compare equal to, and as specific as, any concrete type. See
-    https://github.com/beartype/plum/issues/295.
-
-    This holds at every depth: a nested `Any` is rewritten to `object` by
-    `_substitute_nested_any`, so `list[Any]` and `list[int]` stay distinct. The one
-    approximation is that a nested `Any` and a nested `object` then compare equal,
-    even though a root `Any` remains strictly less specific than a root `object`.
+    Since `beartype` 0.23, `is_subhint(Any, T)` is `True` for every `T`, which would
+    make an unannotated (`Any`-typed) parameter compare equal to, and as specific as,
+    any concrete type. See https://github.com/beartype/plum/issues/295. For Plum's
+    signature bookkeeping, `Any` must instead be the unique least specific type. This
+    check therefore differs from `beartype.door.TypeHint(x) <= TypeHint(y)` in exactly
+    one respect: a root `Any` is only a subhint of `Any`, and a nested `Any` is
+    rewritten to `object` by `_substitute_any`, which reproduces `beartype<0.23`.
 
     Args:
         x (object): First, already-resolved type hint.
         y (object): Second, already-resolved type hint.
 
     Returns:
-        bool: Whether `x` is a subhint of `y` for these bookkeeping purposes.
+        bool: Whether `x` is a subhint of `y`.
     """
     if x is Any:
         return y is Any
     if y is Any:
         return True
     return bool(
-        TypeHintWrapper(_substitute_nested_any(x))
-        <= TypeHintWrapper(_substitute_nested_any(y))
+        TypeHintWrapper(_substitute_any(x)) <= TypeHintWrapper(_substitute_any(y))
     )
 
 
 def _type_hint_eq(x: object, y: object, /) -> bool:
-    """Check whether two already-resolved type hints are the same, for the same
-    signature-bookkeeping purposes as `_type_hint_le` (and with the same `Any`
-    and nested-`Any` handling).
+    """Check whether `x` and `y` are the same hint, where `Any` equals only itself.
 
-    This does not call `_type_hint_le(x, y) and _type_hint_le(y, x)`: each call
-    would independently substitute and wrap both `x` and `y`, doubling that work
-    for no benefit, since `beartype.door.TypeHint.__eq__` already performs (and
-    caches) the equivalent bidirectional subhint check in a single pass.
+    See `_type_hint_le`.
 
     Args:
         x (object): First, already-resolved type hint.
@@ -413,13 +379,12 @@ def _type_hint_eq(x: object, y: object, /) -> bool:
     Returns:
         bool: Whether `x` and `y` denote the same type.
     """
-    x_is_any = x is Any
-    y_is_any = y is Any
-    if x_is_any or y_is_any:
-        return x_is_any and y_is_any
+    if x is Any or y is Any:
+        return x is y
+    # Do not derive this from `_type_hint_le`: `TypeHintWrapper.__eq__` already checks
+    # both directions in one pass.
     return bool(
-        TypeHintWrapper(_substitute_nested_any(x))
-        == TypeHintWrapper(_substitute_nested_any(y))
+        TypeHintWrapper(_substitute_any(x)) == TypeHintWrapper(_substitute_any(y))
     )
 
 
--- a/src/plum/_type.py
+++ b/src/plum/_type.py
@@ -284,23 +284,38 @@ def resolve_type_hint(x: object, /) -> object:
     return x
 
 
-def _type_hints_equal(x: object, y: object, /) -> bool:
-    """Check whether two already-resolved type hints denote the *same* type, for
-    bookkeeping purposes such as detecting whether a newly registered method
-    redefines an existing one, or comparing the specificity of two signatures.
+def _type_hint_le(x: object, y: object, /) -> bool:
+    """Subhint check for plum's signature bookkeeping: redefinition detection
+    (`Signature.__eq__`, via `_type_hints_equal`) and specificity ordering
+    (`Signature.__le__`).
 
-    This differs from `beartype.door.TypeHint(x) == TypeHint(y)` in exactly one
-    respect: `typing.Any` is considered equal only to `typing.Any` itself.
-    Beartype's `TypeHint` equality is built from
-    ``is_subhint(x, y) and is_subhint(y, x)``, and, since ``is_subhint(Any, T)``
-    is `True` for every `T` (`Any` is two-way assignable to everything, per
-    :pep:`484`; see `beartype#530 <https://github.com/beartype/beartype/issues/530>`_
-    and `beartype#616 <https://github.com/beartype/beartype/pull/616>`_), that
-    makes ``TypeHint(Any) == TypeHint(T)`` also `True` for every `T`. That is
-    correct for `Any`'s assignability semantics, but wrong here: an unannotated
-    parameter (which resolves to `Any`) must never be mistaken for an unrelated
-    concrete type, or method registration silently overwrites the wrong method
-    (see `plum#295 <https://github.com/beartype/plum/issues/295>`_).
+    This differs from `beartype.door.TypeHint(x) <= TypeHint(y)` in exactly one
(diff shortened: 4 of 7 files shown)


src/plum/_signature.py
```diff
@@ -126,17 +125,25 @@ def __eq__(self, other: Any, /) -> bool:
...
+            _type_hints_equal(x, y)
+            for x, y in zip(self.types, other.types, strict=True)
+        )
+        varargs_equal = (self.varargs is Missing) == (other.varargs is Missing) and (
+            self.varargs is Missing or _type_hints_equal(self.varargs, other.varargs)
         )
+        return types_equal and varargs_equal and self.precedence == other.precedence
 
     def __hash__(self) -> int:
         return hash((Signature, *self.types, self.varargs))
```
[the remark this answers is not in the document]
nstarman: Good catch, but this predates #296 and isn't something this PR introduces or worsens — it's reproducible on unmodified `master` with stable `beartype==0.22.9`, nothing to do with the `Any` fix here:

```python
from typing import Union
from plum import Signature as Sig

s1, s2 = Sig(Union[int, bool]), Sig(int)
s1 == s2                    # True  (test_signature.py::test_equality already asserts this)
hash(s1) == hash(s2)        # False -- hash/eq contract violated
```

It's inherited straight from `beartype.door.TypeHint` itself: `TypeHint.__hash__` is `hash(self._hint)`, not consistent with its own `__eq__` (`TypeHint(Union[int, bool]) == TypeHint(int)` is `True`, but the two hash differently). Routing `Signature.__hash__` through `TypeHintWrapper(...)` wouldn't fix anything since it reduces to the identical call. A real fix needs a canonical form for semantically-equivalent-but-structurally-different hints, which is a separate, bigger problem than this PR's scope (and arguably belongs upstream in `beartype.door.TypeHint`, or requires `plum` to maintain its own hint canonicalization).

For what it's worth, this PR doesn't make it worse — for the specific `Any` case, it actually removes one instance of the exact same violation: under `beartype>=0.23` before this fix, `Signature(Any) == Signature(int)` was `True` while their hashes still differed (an even broader case of this bug, since `Any` collided with *every* type).

Filed as its own tracking issue so it doesn't get lost: #297.

src/plum/_promotion.py
```diff
@@ -148,8 +148,10 @@ def rule(t1: type1, t2: type2):
         return type_to
 
     # If the types are the same, we don't need to add the reverse rule. Resolve the
-    # types to handle the case where types are equal, but not identical.
-    if TypeHint(resolve_type_hint(type1)) == TypeHint(resolve_type_hint(type2)):
+    # types to handle the case where types are equal, but not identical. Use
+    # `_type_hints_equal` rather than comparing `TypeHint`s directly so that `Any`
+    # is never mistaken for an unrelated concrete type (see `_type_hints_equal`).
```
wesselb: *Comment generated by Claude.*

Redundant cross-reference.

The parenthetical "(see `_type_hints_equal`)" repeats the function name already given in the same sentence:

```suggestion
    # types to handle the case where types are equal, but not identical. Use
    # `_type_hints_equal` rather than comparing `TypeHint`s directly so that `Any`
    # is never mistaken for an unrelated concrete type.
```

src/plum/_signature.py
```diff
@@ -126,17 +125,25 @@ def __eq__(self, other: Any, /) -> bool:
...
-        ) == (
-            tuple(TypeHintWrapper(t) for t in other.types),
-            Missing if other.varargs is Missing else TypeHintWrapper(other.varargs),
-            other.precedence,
+        #
+        # We use `_type_hints_equal` rather than comparing `TypeHintWrapper`s
+        # directly, because `beartype.door.TypeHint(Any) == TypeHint(T)` is `True`
+        # for every `T` (see `_type_hints_equal`'s docstring). Comparing directly
+        # would make two signatures with unrelated concrete types spuriously equal
+        # whenever one of them has an unannotated (and hence `Any`-typed) parameter.
```
wesselb: *Comment generated by Claude.*

Duplicated rationale.

This restates the full story that already lives in `_type_hints_equal`'s docstring (which it also points to). A one-line cross-reference suffices; the fact then lives in one place and cannot drift:

```suggestion
        #
        # Use `_type_hints_equal` rather than comparing `TypeHintWrapper`s directly
        # so that `Any` only equals `Any` itself; see its docstring.
```

src/plum/_signature.py
```diff
@@ -193,11 +201,12 @@ def __le__(self, other: object, /) -> bool:
             else:
                 return True
 
+        # As above, use `_type_hint_le` rather than comparing `TypeHintWrapper`s
+        # directly: `beartype.door.is_subhint(Any, T)` is `True` for every `T`
+        # since `beartype` 0.23, which would make an unannotated (`Any`-typed)
+        # parameter appear exactly as specific as any concrete overload.
```
wesselb: *Comment generated by Claude.*

Duplicated rationale.

The `beartype` 0.23 behaviour is already documented in `_type_hint_le`'s docstring; a cross-reference keeps the fact in one place (cf. the analogous comment in `__eq__`):

```suggestion
        # As above, use `_type_hint_le` so that `Any` is only a subhint of `Any`
        # itself; see its docstring.
```

src/plum/_type.py
```diff
@@ -283,6 +284,49 @@ def resolve_type_hint(x: object, /) -> object:
...
+    This differs from `beartype.door.TypeHint(x) <= TypeHint(y)` in exactly one
+    respect: `typing.Any` is only a subhint of `typing.Any` itself, preserving
+    `Any` as plum's unique *least specific* type. Since `beartype` 0.23,
+    ``is_subhint(Any, T)`` is `True` for every `T` (`Any` is two-way assignable
+    to everything, per :pep:`484`; see
+    `beartype#530 <https://github.com/beartype/beartype/issues/530>`_ /
+    `beartype#616 <https://github.com/beartype/beartype/pull/616>`_), which
+    would otherwise make an unannotated (`Any`-typed) parameter compare equal
+    to, and as specific as, any concrete type (see
+    `plum#295 <https://github.com/beartype/plum/issues/295>`_).
```
wesselb: *Comment generated by Claude.*

Docstring convention and length.

Three things in one suggestion: (1) the code base's convention (Google style) is a one-line imperative summary ending in a period, with elaboration below; (2) the rationale paragraph is denser than the surrounding style — the code base references issues with plain URLs (cf. the warning in `resolve_type_hint`) rather than reST links and `:pep:` roles, and the linked issue already tells the full history; (3) "plum" is capitalised in prose elsewhere ("the current implementation of Plum", "Plum's autoreload hack"). Also includes a note recording the nested-`Any` limitation (see my other comment on this function):

```suggestion
    """Check whether `x` is a subhint of `y`, where `Any` is only a subhint of itself.

    This check is used for Plum's signature bookkeeping: redefinition detection
    (`Signature.__eq__`, via `_type_hints_equal`) and specificity ordering
    (`Signature.__le__`). It differs from `beartype.door.TypeHint(x) <= TypeHint(y)`
    in exactly one respect: `Any` is only a subhint of `Any` itself, preserving `Any`
    as Plum's unique least specific type. Since `beartype` 0.23, `is_subhint(Any, T)`
    is `True` for every `T`, which would otherwise make an unannotated (`Any`-typed)
    parameter compare equal to, and as specific as, any concrete type. See
    https://github.com/beartype/plum/issues/295.

    Note that only a top-level `Any` is special-cased: parameterised hints containing
    `Any`, such as `list[Any]`, still follow `beartype`'s assignability semantics, so
    `list[Any]` and `list[int]` still compare equal under `beartype>=0.23`.
```

src/plum/_type.py
```diff
@@ -283,6 +284,49 @@ def resolve_type_hint(x: object, /) -> object:
...
+
+    Args:
+        x (object): First, already-resolved type hint.
+        y (object): Second, already-resolved type hint.
+
+    Returns:
+        bool: Whether `x` is a subhint of `y` for these bookkeeping purposes.
+    """
+    if x is Any:
+        return y is Any
```
wesselb, line 362: *Comment generated by Claude.*

Nested `Any` still collides.

The `x is Any` check only catches a top-level `Any`. Under `beartype==0.23.0rc0`, the same collision this PR fixes still occurs one level down, because `TypeHint(list[Any]) == TypeHint(list[int])` is `True` there:

```python
@plum.dispatch
def h(x: list[Any]): return "list[Any]"

@plum.dispatch
def h(x: list[int]): return "list[int]"

h(["a"])  # NotFoundLookupError -- `h(x: list[Any])` was silently overwritten.
```

This works correctly on `beartype<0.23` (returns `"list[Any]"`), so it remains a regression that this PR does not cover. A fully general fix would need to recurse structurally with variance handling — essentially reimplementing `is_subhint` — so restricting the fix to top-level `Any` (by far the common case, via unannotated parameters) seems a reasonable scope. But the limitation should at least be recorded: the suggestion in my comment on the docstring above includes a note to that effect. Possibly also worth a follow-up issue.

src/plum/_type.py
```diff
@@ -283,6 +284,49 @@ def resolve_type_hint(x: object, /) -> object:
...
+    return bool(TypeHintWrapper(x) <= TypeHintWrapper(y))
+
+
+def _type_hints_equal(x: object, y: object, /) -> bool:
```
wesselb: *Comment generated by Claude.*

Naming consistency.

`_type_hint_le` (singular) vs `_type_hints_equal` (plural) — both take two hints. Consider renaming `_type_hints_equal` to `_type_hint_eq`, matching `_type_hint_le`. No code suggestion since the rename spans `_type.py`, `_signature.py`, `_promotion.py`, and `tests/test_type.py`; my other comments quote the current name and would follow the rename.

src/plum/_type.py
```diff
@@ -283,6 +284,49 @@ def resolve_type_hint(x: object, /) -> object:
...
+    """Check whether two already-resolved type hints are the same, for signature
+    bookkeeping (see `_type_hint_le`, on which this is built via the usual
+    antisymmetry syllogism: `x <= y and y <= x` implies `x == y`).
```
wesselb: *Comment generated by Claude.*

Docstring convention.

Same as for `_type_hint_le`: one-line summary, elaboration below. Also, "antisymmetry" alone is the established term — "syllogism" adds no information:

```suggestion
    """Check whether two already-resolved type hints are the same.

    This equality check is used for signature bookkeeping and is built on
    `_type_hint_le` via antisymmetry: `x <= y and y <= x` implies `x == y`.
```

tests/test_type.py
```diff
@@ -279,3 +282,36 @@ def test_is_faithful_literal(recwarn):
     assert not is_faithful(Literal[1])
     # There should be no warnings.
     assert len(recwarn) == 0
+
+
+def test_type_hints_equal():
+    """Regression test for https://github.com/beartype/plum/issues/295: `Any`
+    must compare equal only to `Any`, never to an unrelated concrete type.
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention.

One-line summary, elaboration below (cf. `test_register_any_never_collides_with_concrete_type`, which already follows this shape):

```suggestion
    """Regression test for https://github.com/beartype/plum/issues/295.

    `Any` must compare equal only to `Any`, never to an unrelated concrete type.
    """
```

wesselb (commented):
*Review generated by Claude.*

Thanks for the thorough PR — the claims all check out. I reproduced both failure modes on `master` under `beartype==0.23.0rc0`, confirmed this PR fixes them, and ran the full test suite under both `beartype==0.22.8` and `beartype==0.23.0rc0` (passes; the remaining failures are unrelated environment artifacts). The comments below are mostly minor style points; the one substantive remark is that parameterised hints containing `Any` (e.g. `list[Any]`) still collide under `beartype>=0.23`.

tests/test_type.py
```diff
@@ -279,3 +282,36 @@ def test_is_faithful_literal(recwarn):
...
+
+    # Equivalent but not identical types should still compare equal.
+    assert _type_hints_equal(Union[int, bool], int)  # noqa: UP007
+
+
+def test_type_hint_le():
+    """Regression test for https://github.com/beartype/plum/issues/295: `Any`
+    must stay the unique least-specific type -- a subhint of nothing but itself,
+    even though everything else remains a subhint of it.
+    """
```
wesselb: *Comment generated by Claude.*

Docstring convention.

As in `test_type_hints_equal` above:

```suggestion
    """Regression test for https://github.com/beartype/plum/issues/295.

    `Any` must stay the unique least-specific type - a subhint of nothing but
    itself, even though everything else remains a subhint of it.
    """
```

src/plum/_signature.py
```diff
@@ -126,17 +125,25 @@ def __eq__(self, other: Any, /) -> bool:
...
+        # would make two signatures with unrelated concrete types spuriously equal
+        # whenever one of them has an unannotated (and hence `Any`-typed) parameter.
+        types_equal = all(
+            _type_hints_equal(x, y)
+            for x, y in zip(self.types, other.types, strict=True)
+        )
+        varargs_equal = (self.varargs is Missing) == (other.varargs is Missing) and (
+            self.varargs is Missing or _type_hints_equal(self.varargs, other.varargs)
         )
+        return types_equal and varargs_equal and self.precedence == other.precedence
```
wesselb:
```suggestion
        if self.varargs is Missing:
            # Both must be missing.
            varargs_equal = other.varargs is Missing
        else:
            # Neither must be missing.
            varargs_equal = other.varargs is not Missing
            # And they must be equal.
            varargs_equal &= _type_hints_equal(self.varargs, other.varargs)
        return types_equal and varargs_equal and self.precedence == other.precedence
```

This is correct, but I had to stare at it for a bit. Perhaps a  little more handholding is nice.

tests/test_resolver.py
```diff
@@ -157,6 +157,46 @@ def f(*xs):
...
 
+@pytest.mark.parametrize("register_any_first", [True, False])
+def test_register_any_never_collides_with_concrete_type(register_any_first):
+    """Regression test for https://github.com/beartype/plum/issues/295.
+
+    An unannotated parameter resolves to `Signature(Any)`. `Any` must never be
+    treated as "the same registered signature" as a concrete type -- regardless
+    of registration order -- or one of the two methods silently overwrites the
+    other and the resolver ends up with only one of them registered.
+    """
```
wesselb:
```suggestion
    An unannotated parameter resolves to `Signature(Any)`. `Any` must never be
    treated as "the same registered signature" as a concrete type - regardless
    of registration order - or one of the two methods silently overwrites the
    other and the resolver ends up with only one of them registered.
    """
```

Let's stick to the convention that em-dashes are single hyphens. Double hyphens is a bit confusing, but those correspond to en-dashes in LaTeX.

wesselb (commented):
Thanks for the fix @nstarman! I think this largely works, but there is a deeper issue which is to do with nested `Any`s. This might be a tricky one to get right...

By the way, I've used Claude to help with the review. Let me know if you're happy with the comments.

src/plum/_promotion.py
nstarman: *Comment generated by Claude.*

Applied as suggested (with the `_type_hint_eq` rename from your other comment).

src/plum/_signature.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_signature.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Applied — one-line imperative summary, plain URL, "Plum" capitalised.

One change from your text: the nested-`Any` note is not the limitation you drafted, because that case is now fixed (see my reply on the `x is Any` comment). The docstring instead records what actually remains — that a *nested* `Any` and a *nested* `object` compare equal, even though a root `Any` stays strictly less specific than a root `object`.

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Applied. Renamed across `_type.py`, `_signature.py`, `_promotion.py` and `tests/test_type.py`; no references to the old name remain.

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_type.py
nstarman: *Comment generated by Claude.*

Applied as suggested, with the function renamed to `test_type_hint_eq`.

tests/test_type.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_resolver.py
nstarman: *Comment generated by Claude.*

Applied. Also fixed the same `--` in `test_type_hint_le`'s docstring.

src/plum/_signature.py
nstarman: *Comment generated by Claude.*

Applied, with one correction — the suggestion as written raises.

`&=` is bitwise, not logical, so it does not short-circuit: when `self.varargs` is present and `other.varargs` is `Missing`, the second line still runs and calls `_type_hint_eq(self.varargs, Missing)`, giving ``TypeError: `Missing` has no boolean value``. It fails `test_equality` and `test_varargs_tie_breaking`. The original `and` spelling short-circuited, which is what hid it.

An `elif` keeps the hand-holding you were after and short-circuits properly:

```python
if self.varargs is Missing:
    # Both must be missing.
    varargs_equal = other.varargs is Missing
elif other.varargs is Missing:
    # Neither must be missing.
    varargs_equal = False
else:
    # And they must be equal.
    varargs_equal = _type_hint_eq(self.varargs, other.varargs)
```

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Confirmed, and now fixed rather than documented — the recursion turned out to be avoidable.

**The trick is to substitute rather than recurse.** `object` already has exactly the ordering `Any` is meant to have for bookkeeping: a subhint of nothing but itself, and a superhint of everything. So rewriting every nested `Any` to `object` and handing the result straight to `beartype` gets the right answer without reimplementing `is_subhint` — variance, unions and nesting stay `beartype`'s job:

```
is_subhint(list[object], list[int]) = False     # list[Any] </= list[int]
is_subhint(list[int], list[object]) = True      # list[int] <=  list[Any]
```

The root keeps its own guard, so a root `Any` stays *strictly* less specific than a root `object` — which is the `beartype<0.23` behaviour, and something a blanket substitution would have destroyed (`Sig(Any) == Sig(object)` would wrongly become `True`).

Your repro now gives `"list[Any]"` on `0.23.0rc0`, matching `0.22.8`.

**The same bug was also live in Plum's own parametric types,** via a different route — `_default_le_type_par` compared type parameters with a raw `TypeHint` comparison, so `P[Any]` and `P[int]` collided and `P[str]` stopped resolving at all:

```python
@plum.dispatch
def k(x: P[Any]): return "any"
@plum.dispatch
def k(x: P[int]): return "int"

k(P[str]("a"))   # 0.23.0rc0: NotFoundLookupError; 0.22.8: "any"
```

That one is a genuine one-liner: route it through `_type_hint_le`. It restores `issubclass(P[Any], P[int]) is False`, matching `beartype<0.23`. (No import cycle — `_parametric` already imports from `_type`.)

**What remains,** now recorded in the docstring: a *nested* `Any` and a *nested* `object` compare equal, since both become `object` below the root. Distinguishing them would need a hint that is a strict supertype of `object`, which cannot exist nominally. It is a much smaller wart than the collision it replaces, and no test or behaviour depends on it.

Verified on `Annotated` (validator preserved, nested `Any` still rewritten), `Literal` (arguments are values, left alone), both `Callable` forms, unions, and deep nesting. 173 tests pass under both `0.22.8` and `0.23.0rc0`.

nstarman:
*Comment generated by Claude.*

All 12 review comments applied in `d644a7b`, with a reply on each thread. Summary of the two that were more than wording:

**The nested-`Any` limitation is fixed, not documented.** The suggestion was to record it and open a follow-up, on the grounds that a general fix would mean reimplementing `is_subhint` with variance handling. It turns out to be avoidable: `object` already has exactly the ordering `Any` needs for bookkeeping — a subhint of nothing but itself, a superhint of everything. Rewriting every nested `Any` to `object` and handing the result to `beartype` gets the right answer while leaving variance, unions and nesting to `beartype`:

```
is_subhint(list[object], list[int]) = False     # list[Any] </= list[int]
is_subhint(list[int], list[object]) = True      # list[int] <=  list[Any]
```

The substitution deliberately applies only *below* the root. A blanket `Any -> object` would make `Sig(Any) == Sig(object)` true and destroy the strict `object < Any` ordering — I checked that against `0.22.8` before settling on the design.

**The same bug was live in Plum's own parametric types,** through a different route: `_default_le_type_par` compared type parameters with a raw `TypeHint` comparison. It is more severe than the `list[Any]` case, since the method vanishes outright — `k(P[str]("a"))` raises `NotFoundLookupError` on `0.23.0rc0` and returns `"any"` on `0.22.8`. One line to route it through `_type_hint_le`, restoring `issubclass(P[Any], P[int]) is False`.

**One suggestion needed a correction.** The `__eq__` rewrite raises as written: `&=` is bitwise and does not short-circuit, so with varargs on one side only it calls `_type_hint_eq(varargs, Missing)` and gives ``TypeError: `Missing` has no boolean value``, failing `test_equality` and `test_varargs_tie_breaking`. The original `and` spelling had been hiding it. An `elif` branch keeps the intended readability.

**What remains,** now in the docstring rather than as a follow-up: a *nested* `Any` and a *nested* `object` compare equal, since both become `object` below the root. Distinguishing them would need a hint that is a strict supertype of `object`, which cannot exist nominally.

**Verification.** Checked the rewrite against `Annotated` (validator preserved, nested `Any` still rewritten), `Literal` (arguments are values, left alone), both `Callable` forms, unions and deep nesting. 173 tests pass under both `beartype==0.22.8` and `beartype==0.23.0rc0`, with only the pre-existing `test_methodlist_repr` path-length failure. `ruff` clean, `mypy` clean, `pyright` back at its baseline count.

nstarman:
## Performance follow-up: `lru_cache` + `mypyc`, and how they relate

`_substitute_nested_any` (added for the nested-`Any` fix) runs on essentially every non-`Any` `_type_hint_le`/`_type_hint_eq` comparison — i.e. most dispatch resolutions and method registrations. Two things landed for this in `cb58a7a`/`23f9ba3`, and I want to make sure both are visible here rather than only in the commit bodies, since they're easy to mistake for redundant:

**`@lru_cache(maxsize=4096)` on `_substitute_nested_any`** (the actual code change): the set of distinct type hints in a program is small and fixed (the annotations in `@dispatch`-decorated signatures), but the number of *comparisons* between them isn't — the same hints get walked repeatedly. Benchmarked against a realistic repeated-comparison workload:
- `_substitute_nested_any` itself: **~22x faster** warm vs. cold
- End to end at `_type_hint_le`/`_type_hint_eq` (including `beartype`'s own `TypeHintWrapper` construction/comparison, which caching can't touch): **~1.7x faster**

**`mypyc`** (no code change needed — already configured at the project level): `src/plum/_type.py` is already in `pyproject.toml`'s `[tool.hatch.build.targets.wheel.hooks.mypyc]` `include` list, and `[tool.cibuildwheel.environment]` sets `HATCH_BUILD_HOOKS_ENABLE = "1"` for the `mypyc-wheels` job in `publish.yml`, which runs on every `release: published` and ships mypyc-compiled wheels to PyPI alongside the pure-Python wheel/sdist in the same job. So this module is *already* targeted for compilation on every real release — nothing to opt into here.

I compiled `_type.py` standalone with `mypyc` (using this project's configured `mypy-args`) to check what it actually buys on top of the cache, rather than assume: **~1.4x speedup on the cache-*miss* path only** (first computation of a given hint); no measurable difference on the cache-*hit* path, since that's already dominated by `lru_cache`'s own C-implemented dict lookup rather than this module's bytecode.

So: caching reduces *how often* the cold path runs at all (dominant effect for realistic workloads, where the same hints recur); compilation makes the cold path itself faster *when* it does run (smaller, but real, and already happening automatically). Complementary, not alternatives.

nstarman:
@wesselb ready for another review!

src/plum/_parametric.py
```diff
@@ -123,9 +123,11 @@ def type_parameter(cls) -> object:
 
 def _default_le_type_par(p_left: TypeHint | object, p_right: TypeHint | object) -> bool:
     if is_type(p_left) and is_type(p_right):
-        p_left = beartype.door.TypeHint(resolve_type_hint(p_left))
-        p_right = beartype.door.TypeHint(resolve_type_hint(p_right))
-        return p_left <= p_right
+        # `_type_hint_le` rather than comparing `TypeHint`s directly, so that a type
+        # parameter of `Any` stays distinct from a concrete one; see its docstring.
+        # Without this, `P[Any]` and `P[int]` collide under `beartype>=0.23` exactly
+        # as `Any` and `int` do. See https://github.com/beartype/plum/issues/295.
```
wesselb: *Comment generated by Claude.*

Duplicated rationale.

The other call sites settled on a two-line cross-reference; the `beartype>=0.23` story and the issue link already live in `_type_hint_le`'s docstring:

```suggestion
        # Use `_type_hint_le` rather than comparing `TypeHint`s directly so that a type
        # parameter of `Any` stays distinct from a concrete one; see its docstring.
```

src/plum/_type.py
```diff
@@ -283,6 +294,135 @@ def resolve_type_hint(x: object, /) -> object:
...
+def _substitute_any(arg: object, /) -> object:
+    """Replace `Any` with `object` in a single argument of a type hint."""
+    if arg is Any:
+        return object
+    if isinstance(arg, list):
+        # The parameter list of a `Callable`.
+        return [_substitute_any(x) for x in arg]
+    if arg is Ellipsis or arg is None:
+        return arg
+    return _substitute_nested_any(arg)
```
wesselb: *Comment generated by Claude.*

`lru_cache` raises on unhashable hints, and the helper pair can be one function.

Two things, one suggestion.

1. `lru_cache` hashes its argument, but type hints need not be hashable. `Annotated[int, {"a": 1}]` (or `Annotated[int, ["a"]]`) dispatches fine on `master` and raises `TypeError: unhashable type: 'dict'` on this branch, under both `beartype` versions. The docstring's argument that `Signature.__hash__` already requires hashability does not hold: that method exists, but nothing on the dispatch path calls it. The existing `_hashable` helper in this module is the natural guard.

2. The `_substitute_nested_any`/`_substitute_any` split exists only to leave the root alone, but both callers already return before reaching this when either root is `Any`, so a single function that replaces every `Any`, root included, gives identical results with less machinery. The `Union` and `Callable` special cases are unnecessary too: `origin[new_args]` rebuilds `Union[int, object]`, `Callable[[int, object], object]`, `Callable[..., object]`, and `Annotated[...]` alike, so only `types.UnionType` needs `reduce(or_, ...)`, exactly as in `resolve_type_hint`. The `Ellipsis`/`None` branch and the `if not args` guard are redundant for the same reason. Also: the `lru_cache` paragraph is an implementation detail that belongs in a comment, `--` should be `-`, and `BLE001` is not an enabled rule.

```suggestion
def _substitute_any(hint: object, /) -> object:
    """Replace every `Any` in `hint` with `object`.

    For Plum's signature bookkeeping, `object` has exactly the ordering that `Any`
    should have: it is a subhint of nothing but itself, and everything is a subhint
    of it. Rewriting `Any` to `object` therefore leaves nesting and variance to
    `beartype`. A hint that cannot be rebuilt is returned unchanged.

    Args:
        hint (object): Already-resolved type hint.

    Returns:
        object: `hint` with every `Any` replaced by `object`.
    """
    # The same hints are compared over and over, so cache the rewrite. Not every hint
    # is hashable, e.g. `Annotated[int, {"a": 1}]`, so only cache when possible.
    if _hashable(hint):
        return _substitute_any_cached(hint)
    return _substitute_any_uncached(hint)


def _substitute_any_uncached(hint: object, /) -> object:
    if hint is Any:
        return object
    if isinstance(hint, list):
        # The parameters of a `Callable`.
        return [_substitute_any(arg) for arg in hint]
    origin = get_origin(hint)
    # `Literal`'s arguments are values, not hints, so they must not be rewritten.
    if origin is None or origin is Literal:
        return hint
    args = get_args(hint)
    new_args = tuple(_substitute_any(arg) for arg in args)
    if new_args == args:
        return hint
    try:
        if origin is UnionType:
            # `types.UnionType` cannot be subscripted.
            return reduce(or_, new_args)
        return origin[new_args]
    except Exception:
        return hint


_substitute_any_cached = lru_cache(maxsize=4096)(_substitute_any_uncached)
```

Follow-ups outside these lines: `Union` becomes an unused import, the two call sites become `_substitute_any(x)`/`_substitute_any(y)` (after which `ruff format` joins each `bool(...)` comparison onto one line), and `test_substitute_nested_any_rebuilds_hints` becomes a test of `_substitute_any` in which `_substitute_any(Any) is object`. I ran this variant against the full suite under `beartype==0.22.8` and `0.23.0rc0`, plus `mypy` and `ruff`. If you would rather not cache at all, the guard and the two extra names go away as well.

src/plum/_type.py
```diff
@@ -283,6 +294,135 @@ def resolve_type_hint(x: object, /) -> object:
...
+    approximation is that a nested `Any` and a nested `object` then compare equal,
+    even though a root `Any` remains strictly less specific than a root `object`.
+
+    Args:
+        x (object): First, already-resolved type hint.
+        y (object): Second, already-resolved type hint.
+
+    Returns:
+        bool: Whether `x` is a subhint of `y` for these bookkeeping purposes.
+    """
```
wesselb, line 360: *Comment generated by Claude.*

Docstring: stale call-site list and duplicated rationale.

The list of users is already out of date (`_parametric.py` and `_promotion.py` call this too), and the nested-`Any` paragraph repeats the helper's docstring. The "approximation" is also not one: `beartype<0.23` already had `TypeHint(list[Any]) == TypeHint(list[object])`, so the nested behaviour matches the old behaviour exactly.

```suggestion
    """Check whether `x` is a subhint of `y`, where `Any` is only a subhint of itself.

    Since `beartype` 0.23, `is_subhint(Any, T)` is `True` for every `T`, which would
    make an unannotated (`Any`-typed) parameter compare equal to, and as specific as,
    any concrete type. See https://github.com/beartype/plum/issues/295. For Plum's
    signature bookkeeping, `Any` must instead be the unique least specific type. This
    check therefore differs from `beartype.door.TypeHint(x) <= TypeHint(y)` in exactly
    one respect: a root `Any` is only a subhint of `Any`, and a nested `Any` is
    rewritten to `object` by `_substitute_any`, which reproduces `beartype<0.23`.

    Args:
        x (object): First, already-resolved type hint.
        y (object): Second, already-resolved type hint.

    Returns:
        bool: Whether `x` is a subhint of `y`.
    """
```

src/plum/_type.py
```diff
@@ -283,6 +294,135 @@ def resolve_type_hint(x: object, /) -> object:
...
+        x (object): First, already-resolved type hint.
+        y (object): Second, already-resolved type hint.
+
+    Returns:
+        bool: Whether `x` and `y` denote the same type.
+    """
+    x_is_any = x is Any
+    y_is_any = y is Any
+    if x_is_any or y_is_any:
+        return x_is_any and y_is_any
```
wesselb: *Comment generated by Claude.*

Docstring convention and implementation detail.

The summary runs over three lines again (the one-line form from round 1 was lost in the later perf commit), and the second paragraph explains why the body is not derived from `_type_hint_le`, which is a comment, not interface. The `Any` check also collapses to `x is y`:

```suggestion
    """Check whether `x` and `y` are the same hint, where `Any` equals only itself.

    See `_type_hint_le`.

    Args:
        x (object): First, already-resolved type hint.
        y (object): Second, already-resolved type hint.

    Returns:
        bool: Whether `x` and `y` denote the same type.
    """
    if x is Any or y is Any:
        return x is y
    # Do not derive this from `_type_hint_le`: `TypeHintWrapper.__eq__` already checks
    # both directions in one pass.
```

tests/test_resolver.py
```diff
@@ -157,6 +158,77 @@ def f(*xs):
...
+    if register_any_first:
+        r.register(m_any)
+        r.register(m_int)
+    else:
+        r.register(m_int)
+        r.register(m_any)
+
+    assert len(r) == 2
+    assert r.resolve(([1],)) == m_int
+    assert r.resolve((["a"],)) == m_any
```
wesselb: *Comment generated by Claude.*

Duplicated test bodies.

The two tests differ only in the hints and the values, so parametrise over those. (The nested case is also pinned at the `_type_hint_eq` level in `test_type.py`, so dropping it here instead would be fine too.)

```suggestion
@pytest.mark.parametrize("register_any_first", [True, False])
@pytest.mark.parametrize(
    "any_hint, int_hint, any_value, int_value",
    [(typing.Any, int, 1.5, 1), (list[typing.Any], list[int], ["a"], [1])],
)
def test_register_any_never_collides_with_concrete_type(
    register_any_first, any_hint, int_hint, any_value, int_value
):
    """Regression test for https://github.com/beartype/plum/issues/295.

    `Any`, which an unannotated parameter resolves to, must never be treated as "the
    same registered signature" as a concrete type, also when nested and regardless
    of registration order. Otherwise one of the two methods silently overwrites the
    other and the resolver ends up with only one of them registered.
    """
    r = Resolver()

    def f(x):
        return x

    m_any = Method(f, plum.Signature(any_hint))
    m_int = Method(f, plum.Signature(int_hint))

    if register_any_first:
        r.register(m_any)
        r.register(m_int)
    else:
        r.register(m_int)
        r.register(m_any)

    # Both methods must survive registration, and the more specific one must be
    # selected regardless of registration order.
    assert len(r) == 2
    assert r.resolve((int_value,)) == m_int
    assert r.resolve((any_value,)) == m_any
```

wesselb (commented):
*Review generated by Claude.*

Round 2. All round-1 comments were applied, and the claims check out: the full suite, `ruff`, and `mypy` pass at the PR head under both `beartype==0.22.8` and `beartype==0.23.0rc0`, the nested-`Any` rewrite reproduces the `beartype<0.23` ordering for every case I probed, and `_type.py` still compiles with `mypyc`. One regression: `lru_cache` makes dispatch raise on hints with unhashable metadata, such as `Annotated[int, {"a": 1}]`, which works on `master`. See the comment on `_substitute_nested_any`, which also suggests a simpler shape for the helper. The remaining comments are docstring and test-organisation points.

wesselb (approved):
Did another Claude-assisted review, but this looks great to me! Happy for you to merge it after the comments. :)

src/plum/_parametric.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Applied as suggested. The correction is right and I verified it — on `beartype==0.22.8`, `TypeHint(list[Any]) == TypeHint(list[object])` is already `True` (while `TypeHint(Any) == TypeHint(object)` is `False`). So the nested rewrite reproduces the old ordering exactly rather than approximating it; my "one approximation" wording was wrong.

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Applied as suggested.

tests/test_resolver.py
nstarman: *Comment generated by Claude.*

Applied as suggested — merged rather than dropped, so the nested case stays pinned at the resolver level as well as in `test_type.py`.

src/plum/_type.py
nstarman: *Comment generated by Claude.*

Both applied — and the `lru_cache` crash is real. Reproduced it before fixing, on `fix/any-redefinition-equality` vs a clean `master` worktree, under both `beartype` versions:

```
this branch:  Annotated[int, {"a":1}] -> RAISED TypeError: unhashable type: 'dict'
              Annotated[int, ["a"]]   -> RAISED TypeError: unhashable type: 'list'
master:       both dispatch fine
```

Your point about the docstring's justification is also right: `Signature.__hash__` exists but nothing on the dispatch path calls it, so it never imposed hashability. Now behind `_hashable`, with a test for the unhashable case.

On the simplification, I checked each claim rather than taking it on trust, and they all hold:

- Merging the two functions is a no-op, as you say — both callers return before reaching the helper when either root is `Any`, so rewriting the root too is unobservable.
- `origin[new_args]` does rebuild `Union[int, object]`, `Callable[[int, object], object]`, `Callable[..., object]` and `Annotated[list[object], Is[...]]`. (I briefly thought `Annotated` was a counter-example, but that was my test passing one argument where the real rewrite passes the metadata through as well.) Only `types.UnionType` needs `reduce(or_, ...)`, on 3.10–3.13 where `get_origin(int | str)` is not `typing.Union`.
- The `Ellipsis`/`None` branch and `if not args` were both dead: `get_origin` is `None` for all three, so the generic path already returned them unchanged.

Kept the cache rather than dropping it, since the earlier perf commit added it deliberately. Measured cost of the guard: registration of 4 methods 120 µs vs 122 µs and `g(1)` dispatch 0.407 µs vs 0.417 µs — unchanged — with ~6% added to a tight loop of raw `_type_hint_le`/`_type_hint_eq` calls, which does not surface in registration.

174 tests pass under `beartype==0.22.8` and `0.23.0rc0`; the compiled wheel builds with the module-level `_substitute_any_cached` assignment, `assert COMPILED` passes, and the mypyc selection is green.

nstarman:
*Comment generated by Claude.*

Thanks — all five applied in `aa79930`.

**The `lru_cache` regression was real.** Reproduced before fixing, on this branch against a clean `master` worktree, under both `beartype` versions:

```
this branch:  Annotated[int, {"a": 1}] -> TypeError: unhashable type: 'dict'
              Annotated[int, ["a"]]    -> TypeError: unhashable type: 'list'
master:       both dispatch fine
```

Your point about the justification holds too: `Signature.__hash__` exists but nothing on the dispatch path calls it, so hashability was never actually required. The rewrite now sits behind the module's existing `_hashable` guard, with a test for the unhashable case.

**The simplification checks out,** and I verified each part rather than taking it on trust. Merging the two helpers is unobservable, because both callers return before reaching it when either root is `Any`. `origin[new_args]` really does rebuild `Union`, both `Callable` forms and `Annotated`, leaving only `types.UnionType` needing `reduce(or_, ...)` on 3.10–3.13. And the `Ellipsis`/`None` branch and `if not args` were dead — `get_origin` is `None` for all three, so the generic path already returned them unchanged. Net −68 lines across the four files.

**The docstring correction is right,** and it is worth stating plainly since I wrote the wrong version: on `beartype==0.22.8`, `TypeHint(list[Any]) == TypeHint(list[object])` is already `True`, while `TypeHint(Any) == TypeHint(object)` is `False`. So the nested rewrite *reproduces* the pre-0.23 ordering rather than approximating it.

**Kept the cache** rather than dropping it, since the earlier commit added it deliberately. The guard costs nothing where it matters: registering 4 methods is 120 µs vs 122 µs and `g(1)` dispatch 0.407 µs vs 0.417 µs; a tight loop of raw `_type_hint_le`/`_type_hint_eq` calls is ~6% slower, which does not surface in registration.

**Verification.** 174 tests pass under both `beartype==0.22.8` and `0.23.0rc0` — only the pre-existing `test_methodlist_repr` path-length failure remains, identical on `master`. The compiled wheel builds with the module-level `_substitute_any_cached` assignment, `assert COMPILED` passes, and the mypyc test selection is green. Both original fixes still hold: `h(["a"])` gives `"list[Any]"` and `k(P[str]("a"))` gives `"any"` on `0.23.0rc0`. `ruff` and `mypy` clean.

Issue #290, opened 2026-07-28
Design: cacheability aspects — separate `is_faithful` / `is_cacheable` / `cache_key`, with a soundness proof and a seam for `Literal` + generics


nstarman:
Design proposal, opening for discussion. This supersedes the approach in #282 (the very rework @wesselb requested) and subsumes the generics work in #272.

Context: #282 and @wesselb's review (comments `4715632727`, `5095844190`).

> **Updated twice**, after building a draft implementation of the whole sequence (branch `feat/cacheability-aspects` on my fork — see the report comment below). Everywhere the code contradicted the design I have corrected the design in place and said so rather than quietly editing: the `VALUE` extractor for `Literal`, the per-resolver key binding, and — the big one — **user-defined generics turned out to be a Tier-1 aspect, not a Tier-2 consumer as this document originally argued**. Cache-retention was added to Risks.

## Problem

PR #282 made dispatch on `type[X]` cacheable by keying class arguments on identity. To
do so it made `_is_faithful(type[X])` return `True`. @wesselb's objection: `type[X]` is
**not** faithful, so `is_faithful` now lies, and the caching argument lacks a general
soundness proof. He proposed separating two concepts — **faithfulness** and
**cacheability** — with an explicit `cache_key(x)` and `is_cacheable(t)`.

To be precise about the baseline, since it is easy to misread: on `master`,
`is_faithful(type[X])` is already (correctly) `False` — `type[X]` is simply
**uncacheable** there, and one such method disables the cache for the whole function. The
lying predicate is specific to #282's approach, which this design supersedes. Measured
against `master`, the change is *uncacheable → cacheable*, with `is_faithful` keeping its
honest meaning throughout.

Separately, this is the first of a planned **sequence of PRs**. Later PRs widen the set
of cacheable types (notably `Literal`) and pursue common-path speed. This PR must lay a
seam those followups slot into additively, with everything tested along the way.

## Goals

- Restore `is_faithful` to its honest meaning; add `is_cacheable` alongside it (both
  public). Neither is a hand-written predicate that can drift from the other.
- Provide a clear, general **soundness proof** for cache-by-`cache_key`, and freeze the
  one external assumption it rests on with a test.
- **Maintain or improve every measured path** — a hard, benchmarked gate on this and
  every PR in the sequence (no regressions, not even intermediate ones). Concretely: the
  hot (faithful) path is made branchless via a per-resolver bound keyer, registration is
[...]

PR #301, opened 2026-09-03
perf(type): memoise the wrapper used by signature comparison

nstarman:
## Summary

`Signature.__eq__` and `Signature.__le__` compare hints through `_type_hint_eq` /
`_type_hint_le`, and each of those rebuilds `TypeHintWrapper(_substitute_any(h))` from
scratch. The hints are fixed per signature, so all of that work repeats.

It repeats a lot. For a function no cache can serve — one dispatching on `Literal`,
`list[int]`, `Callable[...]` — full resolution runs on **every call**, so those two
helpers run ~14 times per call. Under `cProfile`, `Signature.__le__` is 75% of an
uncached dispatch, and `_substitute_any` plus its `_hashable` probe are ~28% of it.

This memoises the finished wrapper rather than just the rewrite. Unhashable hints
(`Annotated[int, {"a": 1}]`) fall back to building it, exactly as `_substitute_any`
already does.

```python
def _wrapped(hint):
    try:
        return _wrapped_cached(hint)
    except TypeError:  # Unhashable hint.
        return _wrapped_uncached(hint)
```

## It also repairs a regression

Uncached `Literal` dispatch cost **27.7 µs before #296 and 35.7 µs after it**. I
bisected all 17 commits from `fdf6232` to `c44e0cf`, swapping only `src/plum` against a
fixed environment:

| commit | | ns/call |
|---|---|---:|
| `fdf6232` | Remove deprecated `Val` class (#269) | 27,659 |
| *13 commits* | incl. #268, #281, #287, #288 | 27,428 – 28,463 |
| `15dcdbe` | build(mypyc): make `_BoundFunction` native (#289) | 27,750 |
| `c44e0cf` | fix: `Any` must not collide with concrete types (#296) | **35,651** |

Flat for fifteen commits, then one 29% step. The `mypyc` work is not involved; the
added `_substitute_any` call in each comparison is the whole difference.

The memo takes the path *below* the pre-#296 figure rather than merely back to it,
because it also skips `beartype`'s own wrapper lookup, which was there all along.

## Benchmarks

Medians from `tests/benchmarks` (`pytest --benchmark-enable`), same process, same
machine:

| path | before | after | |
|---|---:|---:|---|
| `Literal` dispatch (uncached) | 34,959 ns | 15,750 ns | **2.22× faster** |
| registration — faithful | 46,667 ns | 43,500 ns | 1.07× |
| registration — union | 60,750 ns | 57,333 ns | 1.06× |
| registration — `type[X]` | 58,000 ns | 53,625 ns | 1.08× |
| registration — `Literal` | 54,250 ns | 50,083 ns | 1.08× |
| registration — parametric | 57,917 ns | 53,834 ns | 1.08× |

Every cached path is unchanged, within run-to-run noise: faithful dispatch 388 ns,
faithful 2-arg 410 ns, `type[X]` 3,709 ns, parametric 3,750 ns, `invoke` 119 ns,
annotated return 481 ns, `convert` 583 ns, native baseline 45 ns.

Registration improves because `Resolver.register` compares signatures too.

## Testing

- Two new tests in `tests/test_type.py`, both **mutation-checked**: removing the
  unhashable fallback fails `test_wrapped_handles_an_unhashable_hint`; dropping
  `_substitute_any` from the memoised body fails
  `test_wrapped_matches_the_unmemoised_wrapper`.
- Full suite green — 424 passed, with only the pre-existing `test_methodlist_repr`
  failure, which reproduces identically on unmodified `master` (it is a checkout-path
  length rendering artifact).
- `ruff check`, `ruff format --check` clean.

## Invalidation

Like `_substitute_any`'s own cache, this one is not cleared by `clear_all_cache`. Both
are keyed on an **already-resolved** hint, so `type_mapping` entries and
late-delivered `ModuleType`s are handled by `resolve_type_hint` before either cache is
consulted. Both are bounded at `maxsize=4096`.

## Context

Found while building a `pytest-benchmark` harness for the PR series in #290, which is
what made the per-commit bisection cheap. This PR stands alone — it touches only
`_type.py` and does not depend on #292, #300 or that series.

### DIFF
--- a/src/plum/_type.py
+++ b/src/plum/_type.py
@@ -307,14 +307,6 @@ def _substitute_any(hint: object, /) -> object:
     Returns:
         object: `hint` with every `Any` replaced by `object`.
     """
-    # The same hints are compared over and over, so cache the rewrite. Not every hint
-    # is hashable, e.g. `Annotated[int, {"a": 1}]`, so only cache when possible.
-    if _hashable(hint):
-        return _substitute_any_cached(hint)
-    return _substitute_any_uncached(hint)
-
-
-def _substitute_any_uncached(hint: object, /) -> object:
     if hint is Any:
         return object
     if isinstance(hint, list):
@@ -337,7 +329,32 @@ def _substitute_any_uncached(hint: object, /) -> object:
         return hint
 
 
-_substitute_any_cached = lru_cache(maxsize=4096)(_substitute_any_uncached)
+def _wrap_type_hint_uncached(hint: object, /) -> TypeHintWrapper:
+    return TypeHintWrapper(_substitute_any(hint))
+
+
+_wrap_type_hint_cached = lru_cache(maxsize=4096)(_wrap_type_hint_uncached)
+
+
+def _wrap_type_hint(hint: object, /) -> TypeHintWrapper:
+    """Wrap `hint` for comparison, replacing every `Any` by `object`.
+
+    `Signature` comparison wraps the same fixed hints on every uncached dispatch, so
+    the rewrite and `beartype`'s own wrapper lookup would otherwise repeat per call.
+
+    Args:
+        hint (object): Already-resolved type hint.
+
+    Returns:
+        TypeHintWrapper: Wrapped hint.
+    """
+    # The same hints are compared over and over, so cache the wrapper. Not every hint
+    # is hashable, e.g. `Annotated[int, {"a": 1}]`, so only cache when possible.
+    try:
+        return _wrap_type_hint_cached(hint)
+    except TypeError:
+        # Cannot hash `hint`.
+        return _wrap_type_hint_uncached(hint)
 
 
 def _type_hint_le(x: object, y: object, /) -> bool:
@@ -362,9 +379,7 @@ def _type_hint_le(x: object, y: object, /) -> bool:
         return y is Any
     if y is Any:
         return True
-    return bool(
-        TypeHintWrapper(_substitute_any(x)) <= TypeHintWrapper(_substitute_any(y))
-    )
+    return bool(_wrap_type_hint(x) <= _wrap_type_hint(y))
 
 
 def _type_hint_eq(x: object, y: object, /) -> bool:
@@ -383,9 +398,7 @@ def _type_hint_eq(x: object, y: object, /) -> bool:
         return x is y
     # Do not derive this from `_type_hint_le`: `TypeHintWrapper.__eq__` already checks
     # both directions in one pass.
-    return bool(
-        TypeHintWrapper(_substitute_any(x)) == TypeHintWrapper(_substitute_any(y))
-    )
+    return bool(_wrap_type_hint(x) == _wrap_type_hint(y))
 
 
 def is_faithful(x: object, /) -> bool:
--- a/tests/test_type.py
+++ b/tests/test_type.py
@@ -10,10 +10,12 @@ from plum._type import (
     ModuleType,
     PromisedType,
     ResolvableType,
+    TypeHintWrapper,
     _is_hint,
     _substitute_any,
     _type_hint_eq,
     _type_hint_le,
+    _wrap_type_hint,
     is_faithful,
     resolve_type_hint,
     type_mapping,
@@ -362,7 +364,9 @@ def test_substitute_any_rebuilds_hints():
     assert _substitute_any(Callable[..., Any]) == Callable[..., object]
 
 
-def test_substitute_any_unhashable_metadata():
-    """An unhashable hint must not reach `lru_cache`; it is rewritten uncached."""
+def test_wrap_type_hint_unhashable_hint():
+    """An unhashable hint cannot be cached but must still be wrapped."""
     hint = Annotated[list[Any], {"a": 1}]
-    assert _substitute_any(hint) == Annotated[list[object], {"a": 1}]
+    assert _wrap_type_hint(hint) == TypeHintWrapper(Annotated[list[object], {"a": 1}])
+    # A hashable one is cached, which is the point of the helper.
+    assert _wrap_type_hint(list[Any]) is _wrap_type_hint(list[Any])


nstarman:
Fixed — good catch. The fallback ran `_substitute_any`, whose own `_hashable` probe
hashed the hint a second time only to fail again, when the `lru_cache` above had
already settled the question. It now calls `_substitute_any_uncached` directly.

src/plum/_type.py
```diff
@@ -340,6 +340,26 @@ def _substitute_any_uncached(hint: object, /) -> object:
...
+    """`TypeHintWrapper(_substitute_any(hint))`, memoised on the hint.
+
+    `Signature` comparison wraps the same fixed hints on every uncached dispatch, so
+    the rewrite, the hashability probe and `beartype`'s own wrapper lookup all repeat
+    per call. Memoising the finished wrapper collapses all three to one dict lookup.
+    """
+    try:
+        return _wrapped_cached(hint)
+    except TypeError:  # Unhashable hint, e.g. `Annotated[int, {"a": 1}]`.
+        return _wrapped_uncached(hint)
```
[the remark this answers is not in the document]
nstarman: Fixed — the fallback now calls `_substitute_any_uncached` directly, so the hint is hashed once by `lru_cache` and not again by `_substitute_any`'s own probe.

tests/test_type.py
```diff
@@ -366,3 +368,26 @@ def test_substitute_any_unhashable_metadata():
...
+        assert _wrapped(hint) == TypeHintWrapper(_substitute_any(hint))
+
+
+def test_wrapped_handles_an_unhashable_hint():
+    """An unhashable hint cannot be memoised; it must be wrapped uncached."""
+    hint = Annotated[list[Any], {"a": 1}]
+    assert _wrapped(hint) == TypeHintWrapper(Annotated[list[object], {"a": 1}])
+    # And the comparisons built on it must still work for such a hint.
+    assert _type_hint_eq(hint, hint)
+    assert _type_hint_le(hint, Any)
```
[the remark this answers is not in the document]
nstarman: Half right, and the half that was right is fixed.

`_type_hint_le(hint, Any)` does short-circuit on `y is Any` and never reached `_wrapped` — I instrumented it: **0 calls**. That assertion was vacuous and now uses concrete targets (`list`, `int`), which reach it twice each.

`_type_hint_eq(hint, hint)` has no identity early-return, though: it goes straight to `_wrapped(x) == _wrapped(y)`, measured at **2 calls**. That one was already exercising the path.

Mutation-checked: forcing the memo without the fallback fails the test.

src/plum/_type.py
```diff
@@ -340,6 +340,29 @@ def _substitute_any_uncached(hint: object, /) -> object:
 _substitute_any_cached = lru_cache(maxsize=4096)(_substitute_any_uncached)
```
wesselb: Redundant cache.

With the wrapper cached, this cache is only consulted on a miss of the wrapper cache, i.e. once per hint. Suggest dropping it: fold `_substitute_any_uncached` into `_substitute_any`, remove this line and the `_hashable` check, and let `_wrapped` be the one place that caches:

```python
def _wrapped_uncached(hint: object, /) -> TypeHintWrapper:
    return TypeHintWrapper(_substitute_any(hint))


_wrapped_cached = lru_cache(maxsize=4096)(_wrapped_uncached)


def _wrapped(hint: object, /) -> TypeHintWrapper:
    """Wrap `hint` for comparison, replacing every `Any` by `object`.

    Args:

    Returns:
        TypeHintWrapper: Wrapped hint.
    """
    # The same hints are compared over and over, so cache the wrapper. Not every hint
    # is hashable, e.g. `Annotated[int, {"a": 1}]`, so only cache when possible.
    try:
        return _wrapped_cached(hint)
    except TypeError:
        # Cannot hash `hint`.
        return _wrapped_uncached(hint)
```

`test_substitute_any_unhashable_metadata` then goes too; the `_wrapped` test covers the unhashable case.

This also rewords the `_wrapped` docstring and comments: imperative summary with `Args:` and `Returns:`, and "cache" rather than "memoise", as in `_substitute_any`.

src/plum/_type.py
```diff
@@ -340,6 +340,29 @@ def _substitute_any_uncached(hint: object, /) -> object:
...
 
 
+def _wrapped_uncached(hint: object, /) -> TypeHintWrapper:
+    return TypeHintWrapper(_substitute_any(hint))
+
+
+_wrapped_cached = lru_cache(maxsize=4096)(_wrapped_uncached)
+
+
+def _wrapped(hint: object, /) -> TypeHintWrapper:
```
wesselb: Naming.

Consider `_wrap_type_hint`, a verb like `_substitute_any` and `resolve_type_hint`, with the `_cached` and `_uncached` helpers renamed to match.

tests/test_type.py
```diff
@@ -366,3 +368,29 @@ def test_substitute_any_unhashable_metadata():
...
+def test_wrapped_handles_an_unhashable_hint():
+    """An unhashable hint cannot be memoised; it must be wrapped uncached."""
+    hint = Annotated[list[Any], {"a": 1}]
+    assert _wrapped(hint) == TypeHintWrapper(Annotated[list[object], {"a": 1}])
+    # And the comparisons built on it must still work for such a hint. Both targets
+    # are deliberately concrete: `_type_hint_le(hint, Any)` short-circuits on `Any`
+    # and would never reach `_wrapped` at all.
+    assert _type_hint_eq(hint, hint)
+    assert _type_hint_le(hint, list)
+    assert not _type_hint_le(hint, int)
```
wesselb: Redundant tests.

The only new behaviour is the fallback for an unhashable hint, so one test for that is enough:

```suggestion
def test_wrapped_unhashable_hint():
    """An unhashable hint cannot be cached but must still be wrapped."""
    hint = Annotated[list[Any], {"a": 1}]
    assert _wrapped(hint) == TypeHintWrapper(Annotated[list[object], {"a": 1}])
```

The first test checks `_substitute_any` a second time, through `_wrapped`, on hint forms that `test_substitute_any_rebuilds_hints` already covers, and that the rewrite reaches the comparison is already checked by `test_type_hint_eq_nested_any`. The `_type_hint_eq` and `_type_hint_le` asserts test one-liners over `_wrapped` that have their own tests. If you want the cache itself pinned, one line does it: `assert _wrapped(list[Any]) is _wrapped(list[Any])`.

wesselb (approved):
Thanks, @nstarman!! Looks great. :) Happy to merge after the comments.

src/plum/_type.py
nstarman: Done — folded `_substitute_any_uncached` back into `_substitute_any`, dropped `_substitute_any_cached` and the `_hashable` probe, and `_wrap_type_hint` is now the one place that caches. That also dissolves the earlier Copilot comment about a double hashability probe: with no probe left in `_substitute_any`, the fallback is just `_wrap_type_hint_uncached`, as you wrote it.

Measured before deciding, since the cache did still serve *shared subtrees* across distinct top-level hints. Fresh process per measurement, interleaved, min of 7:

| | sub-cache kept | dropped |
|---|---|---|
| `Signature.__le__` (nested hint) | 2807 / 2483 ns | 2515 / 2520 ns |
| `is_comparable` | 3833 / 3913 ns | 3892 / 3954 ns |
| registration (µs/method, nested hints) | 87.4 / 86.4 | 90.5 / 87.1 |
| cold wrap of 6 shared-subtree hints | 22.3 µs | 27.9 µs |
| re-wrap under cache-eviction pressure (>4096 distinct hints) | 190 / 230 ns | 188 / 197 ns |

The hot paths are unchanged, and even under eviction pressure the two are the same. The only cost is a one-off ~25% on the very first wrap of hints that share subtrees — paid once per distinct hint. Worth it to lose a cache, a function and a probe.

Docstring reworded as suggested: imperative summary with `Args:`/`Returns:`, and "cache" rather than "memoise".

src/plum/_type.py
nstarman: Renamed: `_wrap_type_hint`, `_wrap_type_hint_cached`, `_wrap_type_hint_uncached`. Agreed it reads better alongside `_substitute_any` and `resolve_type_hint`.

tests/test_type.py
nstarman: Taken, including the cache-pinning line — it is the one assertion that fails if the cache is silently removed, which is what this PR is for:

```python
def test_wrap_type_hint_unhashable_hint():
    """An unhashable hint cannot be cached but must still be wrapped."""
    hint = Annotated[list[Any], {"a": 1}]
    assert _wrap_type_hint(hint) == TypeHintWrapper(Annotated[list[object], {"a": 1}])
    # A hashable one is cached, which is the point of the helper.
    assert _wrap_type_hint(list[Any]) is _wrap_type_hint(list[Any])
```

`test_substitute_any_unhashable_metadata` is gone too — with the `_hashable` probe removed from `_substitute_any`, there was nothing left there to test that this does not cover. Mutation-checked: dropping the `except TypeError` fallback fails this test.

PR #316, opened 2026-09-04
fix(function): do not cache a resolution a concurrent clear invalidated

nstarman:
## The bug

`clear_cache` empties `_cache` while holding `_lock`, but `_resolve_method_with_cache` resolves and stores **outside** it. A registration landing between the two writes the superseded method back into the cache *after* the clear was supposed to discard it — and `_pending` is empty by then, so nothing ever invalidates it again. Every later call gets the old method for the life of the process.

Reproduced by parking `resolve_method` between resolving and storing, with a plain `@dispatch def f(x: int)` redefined from another thread:

```
before:  f(1) -> "v1"   (the superseded method, permanently)
after:   f(1) -> "v2"
```

This is pre-existing on `master`, not introduced by anything in flight. It surfaced while reviewing #300, which adds a second cache with the same shape; that PR now uses the same pattern this one introduces.

## The fix

`clear_cache` swaps in a **fresh dict** instead of clearing in place, and the miss path captures the dict *before* resolving and stores into that one:

```python
cache = self._cache                              # capture
method, return_type = self.resolve_method(args)  # may be overtaken by a clear
if self._resolver.is_faithful:
    cache[types] = method, return_type           # store into the captured dict
```

A resolution overtaken by a clear therefore writes into a dict nothing can reach any more, and the live cache is untouched. No counter, no extra lock, nothing new on the instance — and it is how `_verify_cache` is already invalidated elsewhere in this area.

### Why not hold the lock across resolve + store

That is also correct, and was measured. It costs the same single-threaded, but it serialises invalidation behind resolution and holds `_lock` across `beartype` and arbitrary user `__instancecheck__` / `__eq__`. With a resolution artificially stretched to 1 s, a concurrent `clear_all_cache()` blocks for:

| | blocking |
|---|---|
| dict swap (this PR) | **0.1 ms** |
| holding the lock | 949.9 ms |

Rejected on those grounds, not on single-threaded speed.

## Cost

Free on the hot path: the capture is one local load, and only on a cache **miss**.

| | cache hit | cache miss |
|---|---|---|
| dict swap (this PR) | 0.365 µs | 4.708 µs |
| generation counter | 0.366 µs | 4.697 µs |
| holding the lock | 0.360 µs | 4.760 µs |

## Testing

One regression test, +54 lines of diff in total. Mutation-checked both ways: restoring `.clear()` in `clear_cache`, or storing into `self._cache` instead of the captured one, each fails it with `assert 'v1' == 'v2'`.

- It parks `resolve_method` between resolving and storing and registers a second method from the main thread, then asserts the superseded one did not survive.
- Marked `incompatible_with_mypyc`: it patches `resolve_method`, which a compiled module binds at compile time. It scopes the park by *identity* rather than by setting an attribute, since `Function` is a native class with no instance `__dict__`.
- The call runs on a `ThreadPoolExecutor` future rather than a bare `Thread`, so `result(5)` re-raises whatever the call raised and times out if it never finishes — both of which a `Thread` silently swallows — and `monkeypatch` handles the restore.
- Full suite green pure and compiled; `mypyc` wheel built and tested.
- `mypy` clean on 3.10–3.14; `ruff` clean.

`test_methodlist_repr` fails identically on unmodified `master` in my environment — a repr line-wrapping artifact of the long checkout path, unrelated.

### DIFF
--- a/src/plum/_function.py
+++ b/src/plum/_function.py
@@ -346,7 +346,9 @@ class Function(NativeBase):
         # Serialise against concurrent resolution: the `reregister` branch swaps
         # `_pending`/`_resolved`/`_resolver` in multiple steps. See GitHub issue #274.
         with self._lock:
-            self._cache.clear()
+            # A fresh dict, not `.clear()`: a resolution already in flight holds the
+            # old one and stores into that. See `_resolve_method_with_cache`.
+            self._cache = {}
 
             if reregister:
                 # Add all resolved to pending.
@@ -546,12 +548,14 @@ class Function(NativeBase):
             if args is None:
                 args = Signature(*(resolve_type_hint(t) for t in types))
 
-            # Cache miss. Run the resolver based on the arguments.
+            # Cache miss. Capture the dict *before* resolving: a `clear_cache` that
+            # overtakes us swaps in a new one, leaving this store unreachable.
+            cache = self._cache
             method, return_type = self.resolve_method(args)
             # If the resolver is faithful, then we can perform caching using the types
             # of the arguments. If the resolver is not faithful, then we cannot.
             if self._resolver.is_faithful:
-                self._cache[types] = method, return_type
+                cache[types] = method, return_type
             return method, return_type
 
     def invoke(self, *types: TypeHint) -> Callable[..., Any]:
--- a/tests/test_function.py
+++ b/tests/test_function.py
@@ -5,6 +5,7 @@ import textwrap
 import threading
 import typing
 import weakref
+from concurrent.futures import ThreadPoolExecutor
 
 import pytest
 
@@ -715,3 +716,51 @@ def test_weakref_and_doc_assignment(wrap):
     assert weakref.ref(g)() is g
     g.__doc__ = "Replaced."
     assert g.__doc__ == "Replaced."
+
+
+@pytest.mark.incompatible_with_mypyc
+def test_resolution_overtaken_by_a_clear_is_not_cached(monkeypatch, dispatch):
+    """A registration landing mid-resolution must not leave a stale method cached.
+
+    `_resolve_method_with_cache` resolves and stores outside `_lock`, so a method
+    resolved before a `clear_cache` could be written after it. `_pending` is empty by
+    then, so every later call would get the superseded method. See issue #274.
+    """
+
+    @dispatch
+    def f(x: int):
+        return "v1"
+
+    f._resolve_pending_registrations()
+
+    resolved, invalidated = threading.Event(), threading.Event()
+    original, target = Function.resolve_method, f
+
+    def parked(self, *args, **kw_args):
+        out = original(self, *args, **kw_args)
+        # Scoped by identity: `Function` is a `mypyc` native class, so there is no
+        # instance `__dict__` to hang a flag on, and other functions resolve here too.
+        if self is target:
+            resolved.set()
+            # Park between resolving and storing, which is where the clear lands.
+            # Asserted: on a timeout the park would end early and the store would no
+            # longer race the clear, so the test would pass without exercising it.
+            assert invalidated.wait(5), "the clear never landed"
+        return out
+
+    monkeypatch.setattr(Function, "resolve_method", parked)
+    # The call runs on a future so `result` re-raises whatever it raised and times
+    # out if it never finishes; neither would surface from a bare `Thread`.
+    with ThreadPoolExecutor(1) as pool:
+        call = pool.submit(f, 1)
+        assert resolved.wait(5), "the parked thread never reached the park"
+
+        @dispatch
+        def f(x: int):  # noqa: F811
+            return "v2"
+
+        f._resolve_pending_registrations()
+        invalidated.set()
+        call.result(5)
+
+    assert f(1) == "v2"


tests/test_function.py
```diff
@@ -699,3 +700,49 @@ def worker(f=f, barrier=barrier, errors=errors):
...
+    original, target = Function.resolve_method, f
+
+    def parked(self, *args, **kw_args):
+        out = original(self, *args, **kw_args)
+        # Scoped by identity: `Function` is a `mypyc` native class, so there is no
+        # instance `__dict__` to hang a flag on, and other functions resolve here too.
+        if self is target:
+            resolved.set()
+            # Park between resolving and storing, which is where the clear lands.
+            invalidated.wait(5)
```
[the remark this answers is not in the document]
nstarman: Good catch — fixed. It is now `assert invalidated.wait(5), "the clear never landed"`.

Worth noting the assertion is inside the parked function, which runs on the worker thread, so it only surfaces because the call is submitted to a `ThreadPoolExecutor` and awaited with `result()`. A bare `Thread` would have swallowed it and left the same silent pass. Verified by forcing the timeout:

```
E           AssertionError: the clear never landed
E           assert False
E            +  where False = wait(1)
```

tests/test_function.py
```diff
@@ -699,3 +700,49 @@ def worker(f=f, barrier=barrier, errors=errors):
             assert f(1.0) == "float"
     finally:
         sys.setswitchinterval(old_interval)
+
+
+@pytest.mark.incompatible_with_mypyc
+def test_call_does_not_cache_a_result_a_clear_invalidated(monkeypatch, dispatch):
```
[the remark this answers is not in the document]
nstarman: Renamed to `test_resolution_overtaken_by_a_clear_is_not_cached`. You are right that the old name was missing a word — it wanted to read "does not cache a result *that* a clear invalidated" — and the new one names the mechanism rather than the sentence, so it greps.

nstarman:
@wesselb this is the start of the PR chain :)

src/plum/_function.py
```diff
@@ -346,7 +346,9 @@ def clear_cache(self, reregister: bool = True) -> None:
         # Serialise against concurrent resolution: the `reregister` branch swaps
         # `_pending`/`_resolved`/`_resolver` in multiple steps. See GitHub issue #274.
         with self._lock:
-            self._cache.clear()
+            # A fresh dict, not `.clear()`: a resolution already in flight holds the
+            # old one and stores into that. See `_resolve_method_with_cache`.
+            self._cache = {}
```
wesselb: Unclear wording.

Change to:

```suggestion
            # Replace the dictionary rather than emptying it in place: a call that is
            # still resolving a method must not store its result in the live cache. See
            # `_resolve_method_with_cache` for why.
            self._cache = {}
```

"In flight" and "stores into that" are hard to parse without already knowing the bug. The full explanation lives at the other site, so this one only needs to say what is different about the line and point there.

src/plum/_function.py
```diff
@@ -546,12 +548,14 @@ def _resolve_method_with_cache(
             if args is None:
                 args = Signature(*(resolve_type_hint(t) for t in types))
 
-            # Cache miss. Run the resolver based on the arguments.
+            # Cache miss. Capture the dict *before* resolving: a `clear_cache` that
+            # overtakes us swaps in a new one, leaving this store unreachable.
+            cache = self._cache
```
wesselb: Unclear wording.

Change to:

```suggestion
            # Cache miss. If `clear_cache` runs while the method is being resolved, the
            # result may be outdated and must not end up in the cache. `clear_cache`
            # replaces `self._cache` with a new dictionary, so take `self._cache` now:
            # afterwards, `cache` is either still the live cache, or the old dictionary,
            # which nothing reads any more, so storing into it is harmless.
            cache = self._cache
```

"Capture", "overtakes us", and "store unreachable" describe the mechanism in shorthand. Spelling out the two cases makes the line self-explanatory.

tests/test_function.py
```diff
@@ -715,3 +716,51 @@ def f(x):
     assert weakref.ref(g)() is g
     g.__doc__ = "Replaced."
     assert g.__doc__ == "Replaced."
+
+
+@pytest.mark.incompatible_with_mypyc
```
wesselb, line 706: Test placement.

Consider moving this test directly after `test_resolve_pending_registrations_is_thread_safe` so the two concurrency tests sit together, rather than after the weakref test.

tests/test_function.py
```diff
@@ -715,3 +716,51 @@ def f(x):
...
+def test_resolution_overtaken_by_a_clear_is_not_cached(monkeypatch, dispatch):
+    """A registration landing mid-resolution must not leave a stale method cached.
+
+    `_resolve_method_with_cache` resolves and stores outside `_lock`, so a method
+    resolved before a `clear_cache` could be written after it. `_pending` is empty by
+    then, so every later call would get the superseded method. See issue #274.
+    """
```
wesselb: Unclear name and docstring.

Change to:

```suggestion
def test_method_replaced_during_resolution_is_not_cached(monkeypatch, dispatch):
    """Test that a method replaced while a call is resolving it is not cached.

    `_resolve_method_with_cache` resolves and then writes to the cache without holding
    the lock. If a method is registered in between, the cache is cleared before the
    write, so the old method would be written into the cleared cache and, since nothing
    is pending any more, be returned by every later call. See PR #316.
    """
```

"Overtaken", "landing mid-resolution", "stale", and "superseded" all say the same thing in different jargon; "replaced" covers it. The summary now uses the imperative "Test that ...", matching the other tests in this file. Issue #274 is the first-call `__annotations__` race, which is a different bug, so the reference should point at this PR instead.

tests/test_function.py
```diff
@@ -715,3 +716,51 @@ def f(x):
...
+        out = original(self, *args, **kw_args)
+        # Scoped by identity: `Function` is a `mypyc` native class, so there is no
+        # instance `__dict__` to hang a flag on, and other functions resolve here too.
+        if self is target:
+            resolved.set()
+            # Park between resolving and storing, which is where the clear lands.
+            # Asserted: on a timeout the park would end early and the store would no
+            # longer race the clear, so the test would pass without exercising it.
+            assert invalidated.wait(5), "the clear never landed"
+        return out
```
wesselb: Unclear comments; the `mypyc` justification does not apply.

Change to:

```suggestion
    resolved, cleared = threading.Event(), threading.Event()
    original, target = Function.resolve_method, f

    def paused(self, *args, **kw_args):
        out = original(self, *args, **kw_args)
        # The patch applies to every `Function`, so pause only `f`.
        if self is target:
            resolved.set()
            # Pause between resolving and writing to the cache, which is where the clear
            # must happen. Assert rather than only wait: if the wait timed out, the
            # write would happen before the clear, and the test would pass without
            # testing the race.
            assert cleared.wait(5), "The clear never happened."
        return out
```

The test is marked `incompatible_with_mypyc`, so it never runs against the compiled class, and the pure-Python `Function` does have an instance `__dict__`. The identity check is still the simplest way to scope the patch, but the reason is just that the patch is class-wide. "Park", "store", "lands", and "invalidated" are replaced by plain words, and the assertion message becomes a sentence. See the next comment for the matching renames further down.

tests/test_function.py
```diff
@@ -715,3 +716,51 @@ def f(x):
...
+    with ThreadPoolExecutor(1) as pool:
+        call = pool.submit(f, 1)
+        assert resolved.wait(5), "the parked thread never reached the park"
+
+        @dispatch
+        def f(x: int):  # noqa: F811
+            return "v2"
+
+        f._resolve_pending_registrations()
+        invalidated.set()
```
wesselb: Unclear comment.

Change to:

```suggestion
    monkeypatch.setattr(Function, "resolve_method", paused)
    # Use a `ThreadPoolExecutor` so that `call.result` propagates any error raised in
    # the call and times out if the call hangs.
    with ThreadPoolExecutor(1) as pool:
        call = pool.submit(f, 1)
        assert resolved.wait(5), "The call never reached the pause."

        @dispatch
        def f(x: int):  # noqa: F811
            return "v2"

        f._resolve_pending_registrations()
        cleared.set()
```

The executor comment now just names the mechanism: `call.result` is what re-raises an error from the worker and what times out. Also the assertion message as a sentence, and the renames from the previous comment applied here.

tests/test_function.py
```diff
@@ -715,3 +716,51 @@ def f(x):
...
+        call.result(5)
```
wesselb: Discarded result.

Change to:

```suggestion
        assert call.result(5) == "v1"
```

Documents that the call which started before the redefinition returns the method it resolved, and that only later calls see the new one.

wesselb (approved):
Thanks, @nstarman! Very good catch. Left a few comments. Happy to merge after those are resolved. :)

src/plum/_function.py
nstarman: Done, applied as suggested.

src/plum/_function.py
nstarman: Done, applied as suggested.

tests/test_function.py
nstarman: Done, applied as suggested.

tests/test_function.py
nstarman: Done, applied as suggested.

tests/test_function.py
nstarman: Done, applied as suggested.

tests/test_function.py
nstarman: Done, applied as suggested.

tests/test_function.py
nstarman: Done, applied as suggested.

PR #319, opened 2026-09-07
fix(mypyc): keep `Function` weak-referenceable on the compiled wheels

### LINKED ISSUE #317: 2.10.0: assigning `__doc__` on a `Function` fails on the compiled wheels
Since 2.10.0, setting `__doc__` on a `Function` instance raises on the mypyc-compiled wheels:

```python
from plum import Function

def f(x): return x
fn = Function(f)
fn.__doc__ = "Some docstring"
# AttributeError: 'Function' object attribute '__doc__' is read-only
```

The same code works with the pure-Python wheel (for example on macOS arm64, where no compiled wheel is published), because the instance dictionary takes the assignment. It fails with the manylinux and other cp3x wheels.

The cause is the `_DocDescriptor` attached in `_function.py`:

```python
setattr(Function, "__doc__", _DocDescriptor())
```

It defines `__get__` only. On a native class there is no instance dictionary to fall back to, so the assignment has nowhere to go. The comment on the class says `functools.wraps` may copy `__doc__` onto instances, which suggests assignment was meant to keep working.

This breaks fastcore's `add_docs`, which sets `__doc__` on methods after class creation, and through it fastai's import on Linux CI. 2.9.0 works.

A `__set__` on the descriptor that stores into `instance._doc` would restore the behaviour. Happy to send a PR if that approach is acceptable.

Environment: plum-dispatch 2.10.0, Python 3.12, manylinux x86_64 wheel. Possibly related: #315.

### LINKED ISSUE #318: 2.10.0: compiled `Function` is not weak-referenceable, so `jax.jit` over a dispatched function fails
Since 2.10.0, every `jax.jit(<dispatched function>)` raises on first call when the compiled wheels are installed:

```python
import jax, jax.numpy as jnp, plum

d = plum.Dispatcher()

@d
def double(x: jax.Array) -> jax.Array:
    return x * 2

jax.jit(double)(jnp.arange(3.0))
# TypeError: cannot create weak reference to 'Function' object
```

The `jax.jit(...)` call itself succeeds; the failure is on the first invocation, when JAX registers the wrapped callable in its cache.

## Cause

2.10.0 is the first release to ship mypyc-compiled wheels, and `_function.py` is in the compile set, so `Function` is a mypyc **native** class. A native class gets neither `__dict__` nor `__weakref__`. JAX holds a weak reference to the callable it wraps (`jax/_src/pjit.py`, `_cpp_pjit`), so a non-weak-referenceable callable cannot be jitted at all.

`__weakref__` cannot simply be declared on the class. `mypyc/codegen/emitclass.py` emits the `__weakref__` member only when `cl.has_dict`, and on CPython >= 3.12 `has_managed_dict` takes the other branch and drops it. I checked the neighbouring escape hatches:

- `@mypyc_attr(native_class=False)` on `Function` and `_BoundFunction` does not build — codegen fails with `KeyError: 'object'` at `_function.py:657`. It would also give up the native-class win on the hottest class in the library.
- A `native_class=False` base declared in the *same* compiled module still yields no `__weakref__`. I compiled that case to be sure: the subclass gets `__dict__` but `weakref.ref` still raises.

What does work is inheriting from a base in a module that is **not** in the mypyc include list. mypyc then builds the class through CPython's normal type machinery, which grants `__dict__` and `__weakref__` for free, and the class stays native.

## Scope

Only cp310-cp313 on manylinux, Windows and macOS x86_64. There is no cp314 wheel and no macOS arm64 wheel, so those installs fall back to `py3-none-any` and are unaffected. That platform split is what makes this easy to miss: it does not reproduce on an arm64 Mac, and a CI matrix leg on 3.14 stays green.

## Same root cause as #317

#317 (`fn.__doc__ = ...` raising `AttributeError: ... is read-only`) is the same missing-`__dict__` problem: `_DocDescriptor` defines only `__get__`, so on a native class the assignment has nowhere to land. Restoring the instance dictionary fixes both — I verified `__doc__` assignment on a compiled build before and after.

## Impact

Any downstream library that exposes dispatched functions to JAX. In [unxt](https://github.com/GalacticDynamics/unxt) this took out 7 tests plus 2 errors on a compiled build; `unxt` has excluded 2.10.0 in [#917](https://github.com/GalacticDynamics/unxt/pull/917) for now. #317 reports the same 2.10.0 regression reaching fastai's import through fastcore.

## Verification

Built v2.10.0 with `HATCH_BUILD_HOOK_ENABLE_MYPYC=1` on Python 3.13 to reproduce, then again with the fix:

| | baseline 2.10.0 | with fix |
|---|---|---|
| `weakref.ref(Function)` | `TypeError` | OK |
| `jax.jit(f)(x)` | `TypeError` | OK |
| `fn.__doc__ = ...` (#317) | `AttributeError` | OK |
| plum test suite | 2 pre-existing failures | same 2, no new |
| cached-dispatch hot path | 333 ns | 320 ns (noise; a rerun gave 314 vs 315) |

No measurable dispatch cost. PR follows.

nstarman:
Closes #318. Closes #317.

## The problem

2.10.0 is the first release to ship `mypyc`-compiled wheels, with `_function.py` in the compile set, so `Function` is a native class. A native class has neither `__weakref__` nor `__dict__`, and both absences are load-bearing for downstream code.

**No `__weakref__` (#318).** `jax.jit` registers the callable it wraps in a weakly-keyed cache, so every `jax.jit(<dispatched function>)` raises on first call:

```python
jax.jit(double)(jnp.arange(3.0))
# TypeError: cannot create weak reference to 'Function' object
```

**No `__dict__` (#317).** `_DocDescriptor` defines only `__get__`, so `f.__doc__ = ...` has nowhere to land and raises `AttributeError: ... is read-only`. This is what breaks fastcore's `add_docs`, and through it fastai's import.

Only cp310–cp313 on manylinux, Windows and macOS x86_64 are affected. There is no cp314 wheel and no macOS arm64 wheel, so those installs fall back to `py3-none-any`. That split is why neither issue reproduces in local development on Apple silicon.

## The fix

`__weakref__` cannot simply be declared on a native class. `mypyc/codegen/emitclass.py` emits the member only when `cl.has_dict`, and on CPython ≥ 3.12 `has_managed_dict` takes the other branch and drops it even then.

I compiled the two nearby escape hatches to check them rather than reasoning about them:

- **`@mypyc_attr(native_class=False)` on `Function`** — does not build at all; codegen fails with `KeyError: 'object'` at `_function.py:657`. It would also forfeit the native-class win on the hottest class in the library.
- **A `native_class=False` base in the *same* compiled module** — the subclass gets `__dict__`, but `weakref.ref` still raises.

What does work is inheriting from a base in a module that is **not** in the compile set. `mypyc` then builds the class through CPython's normal type machinery, which grants `__dict__` and `__weakref__`, and the class stays native. Hence the new interpreted `_weakrefable` module, applied to both `Function` and `_BoundFunction`, plus a comment on the `include` list so it does not get swept in later.

## Verification

Built with `HATCH_BUILD_HOOK_ENABLE_MYPYC=1` on Python 3.13 and compared against a compiled build of `master`:

| | `master`, compiled | this branch, compiled |
|---|---|---|
| `weakref.ref(Function)` | `TypeError` | OK |
| `weakref.ref(_BoundFunction)` | `TypeError` | OK |
| `f.__doc__ = ...` (#317) | `AttributeError` | OK |
| `jax.jit(f)(x)` (#318) | `TypeError` | OK |
| cached dispatch | 328 ns | 309 ns |

The three new tests in `tests/test_function.py` fail on a compiled `master` and pass here; they are no-ops on the pure-Python wheel, so the `test-mypyc` CI job is what actually exercises them.

The dispatch numbers are noise in both directions — a separate run of the same comparison gave 314 ns against 315 ns. There is no measurable hot-path cost.

Full suite on a compiled build (`pytest tests/ -k "not incompatible_with_mypyc" --ignore=tests/advanced`): `1 failed, 164 passed`. The one failure is `test_methodlist_repr`, which fails identically on a compiled `master` and is unrelated to this change.

## Downstream

[unxt](https://github.com/GalacticDynamics/unxt) lost 7 tests plus 2 errors to #318 and has excluded 2.10.0 in [GalacticDynamics/unxt#917](https://github.com/GalacticDynamics/unxt/pull/917) meanwhile; that exclusion comes back out once this ships.


nstarman:
@wesselb this should be a bug fix release v2.10.1.

nstarman:
Ping @jph00

nstarman:
Second commit: applied the same mechanism to the rest of the file, per review.

`_InvokedMethod` and `_BoundInvokedMethod` were `@mypyc_attr(native_class=False)` for exactly one reason, which `_InvokedMethod`'s own docstring spelled out — `functools.wraps` needs an instance `__dict__` to copy `__name__`/`__doc__` into, and a native class has none. That is the same problem `Weakrefable` solves, so the opt-out is now redundant. Both are native again, and `Function.invoke(...)` goes from ~2600 ns to ~2420 ns (~7%, consistent across three alternating runs).

The other `native_class=False` sites are a *different* problem and stay as they are: `_ModuleDescriptor` subclasses `str`, and `ResolvableType` / `PromisedType` / `ModuleType` in `_type.py` subclass `type`. What forces those is a builtin or metaclass base, not a missing `__dict__`, so `Weakrefable` has nothing to offer them.

`_DocDescriptor` is not a workaround at all — it computes the method-list docstring — and stays.

`_wraps` also stays, but its docstring was stale on both counts, so I corrected it. It claimed `functools.wraps` "writes the read-only native `__module__` and updates a `__dict__` that native instances lack"; with `Weakrefable` the `__dict__` exists and `__module__` is writable. The real reason to keep the helper is subtler: `functools.wraps` copies `__doc__` and `__module__`, and `Function`/`_BoundFunction` serve both through **non-data** descriptors, which an instance attribute silently shadows. Checked rather than assumed — running `functools.wraps` over a two-method `Function` collapses `__doc__` from

```
Doc from the int method.
------------------------
f(x: str) -> str
Doc from the str method.
```

down to just `Doc from the int method.`, and pins `__module__` instead of tracking `_f.__module__`.

pyproject.toml
```diff
@@ -87,6 +87,7 @@ enable-by-default = false
 dependencies = ["hatch-mypyc>=0.16.0", "mypy>=1.18.2"]
 require-runtime-dependencies = true
 include = [ # TODO: include more files
+    # Never add `_weakrefable.py`; it must stay interpreted. See #318.
```
wesselb: Follows from the `_weakrefable.py` comment.

If the base moves into `_mypyc.py`, point this guard at that module:

```suggestion
    # Never add `_mypyc.py`; it must stay interpreted. See `NativeBase` there.
```

src/plum/_weakrefable.py
```diff
@@ -0,0 +1,11 @@
+"""Interpreted base for `mypyc` native classes. Keep out of the `mypyc` include list."""
```
wesselb: Module placement.

Consider putting the base in `_mypyc.py` rather than in a new module. `_mypyc.py` already exists for this kind of `mypyc` accommodation, is interpreted, and is imported by the compiled modules, so this is a move plus one line in its module docstring saying it must stay out of the compile set. One module fewer, and one guard in `pyproject.toml` rather than one per interpreted helper module.

src/plum/_weakrefable.py
```diff
@@ -0,0 +1,11 @@
...
+
+__all__ = ["Weakrefable"]
+
+
+class Weakrefable:
+    """Base granting `mypyc` native subclasses `__dict__` and `__weakref__`.
+
+    A native class has neither. One inheriting from an interpreted base is built by
+    CPython's normal type machinery, which grants both, and stays native. See #318.
+    """
```
wesselb: Naming and docstring.

`Weakrefable` names one of the two things the base grants. Suggest `NativeBase`, with the docstring stating the constraint first and then what native subclasses gain:

```suggestion
class NativeBase:
    """Base for `mypyc` native classes. Must stay interpreted.

    A native class has neither `__dict__` nor `__weakref__`, but one inheriting from a
    class outside the compile set gains both and stays native. Its instances can then be
    weakly referenced, which `jax.jit` needs (#318), and accept undeclared attributes,
    such as `__doc__` (#317) and those :func:`functools.wraps` copies.
    """
```

`__all__` here and the uses in `_function.py` (the import, the four bases, and the `_InvokedMethod` docstring) follow.

tests/test_function.py
```diff
@@ -699,3 +700,38 @@ def worker(f=f, barrier=barrier, errors=errors):
...
+
+def test_doc_can_be_assigned():
+    """`__doc__` is a non-data descriptor, so assignment needs a `__dict__` (#317)."""
+
+    def f(x):
+        """Original."""
+
+    g = Function(f)
+    g.__doc__ = "Replaced."
+    assert g.__doc__ == "Replaced."
```
wesselb: Test economy.

The three tests share one setup and one motivation. A single parametrised test covers both classes and both regressions in a third of the lines:

```suggestion
@pytest.mark.parametrize(
    "wrap", [Function, lambda f: _BoundFunction(Function(f), None)]
)
def test_weakref_and_doc_assignment(wrap):
    """Both must be weak-referenceable (#318) and accept `__doc__` assignment (#317)."""

    def f(x):
        """Original."""

    g = wrap(f)
    assert weakref.ref(g)() is g
    g.__doc__ = "Replaced."
    assert g.__doc__ == "Replaced."
```

Constructing `_BoundFunction` directly, as `test_class_doc_and_module` uses the class directly, needs no `Dispatcher` or dummy class, and `__doc__` assignment then gets checked on `_BoundFunction` too.

wesselb (approved):
Thanks, @nstarman! Looks good. Left an approval with a few comments. Please feel free to merge after you've resolved the comments. I'll then immediately release a new version.

src/plum/_function.py
wesselb, line 1: Stale rationale, in the comment above the `setattr` calls at lines 578-581 (outside the diff, hence a file-level comment):

https://github.com/beartype/plum/blob/5ffffe3bafdebf19fa7a8c1e643ceaebe782a9fb/src/plum/_function.py#L578-L581

With the interpreted base, instances have a `__dict__`, so `__module__` is no longer read-only on them; the reason for keeping the descriptors now lives in `_wraps`'s docstring. Drop the clause:

```python
# Attach `__doc__`/`__module__` here, not in the class body: `mypyc` replaces a class
# `__doc__` with a filler. These descriptors serve instance access (`f.__doc__`,
# `f.__module__`). `setattr` also stops `mypy` treating these as class variables.
```

src/plum/_weakrefable.py
nstarman: Done — the base is now `NativeBase` in `_mypyc.py`, and `_weakrefable.py` is gone. You are right that the module already exists for exactly this, and its docstring now opens with "This module must stay out of the compile set" rather than leaving that implicit in one guard comment.

Verified on a `mypyc` wheel built from this branch that the move preserves the mechanism, since it only works while the base is interpreted:

```
compiled _function: True
interpreted _mypyc : True
plum.COMPILED      : True
```

src/plum/_weakrefable.py
nstarman: Applied verbatim, including the docstring. `Weakrefable` did name only half of it, and the `__dict__` half is the one two of the four subclasses are there for.

The rename reaches further than the diff you saw: both commit messages referred to `Weakrefable` as well, and the second commit's argument turns on the `__dict__`, so those are updated too rather than left pointing at a name that no longer exists.

pyproject.toml
nstarman: Done, with your wording.

tests/test_function.py
nstarman: Applied. Three tests to one parametrised pair, and constructing `_BoundFunction` directly does remove the `Dispatcher` and the dummy class along with the `__func__` indirection.

The part I had not appreciated: it also closes a real gap rather than just shortening things. `__doc__` assignment was only ever checked on `Function`, never on `_BoundFunction` — and `_BoundFunction` carries its own `_DocDescriptor`, so it was the case most likely to break independently.

src/plum/_function.py
nstarman: Fixed, and I checked the premise rather than taking it on reading, since the clause is a claim about compiled behaviour. Built a `mypyc` wheel from this branch and assigned to `__module__` on an instance of each class:

```
--- Function
   weakref       : True
   __doc__ set   : True
   __module__ set: True
--- _BoundFunction
   weakref       : True
   __doc__ set   : True
   __module__ set: True
```

So the clause was stale exactly as you describe: with the interpreted base the assignment lands in the instance `__dict__`. The descriptors stay for the reason now stated in `_wraps`'s docstring — they are non-data, so an instance attribute silently shadows them.

Full suite green against that wheel as well: 177 passed, `-k "not incompatible_with_mypyc"`.

jph00:
Many thanks @nstarman @wesselb

PR #320, opened 2026-09-07
Build macOS arm64 mypyc wheels

nstarman:
## Problem

No `macosx_arm64` mypyc wheels have ever been published. Apple Silicon users silently fall back to the pure-Python wheel.

**Root cause:** the wheels matrix is generated on `ubuntu-latest`, and cibuildwheel's default `archs = "auto"` resolves to the *generator host's* architecture. Cross-listing macOS identifiers from an x86_64 Linux runner only ever emits `*-macosx_x86_64`:

```console
$ cibuildwheel --print-build-identifiers --platform macos   # on x86_64 Linux
cp310-macosx_x86_64
cp311-macosx_x86_64
cp312-macosx_x86_64
cp313-macosx_x86_64
```

## Fix

Pass `--archs x86_64,arm64` explicitly to the macOS identifier listing. The matrix goes from 4 to 8 macOS identifiers, all on the existing `macos-14` runner — arm64 natively, x86_64 cross-compiled, exactly as today.

Also drops `*-macosx_arm64` from `test-skip`. That entry was written back when macOS wheels were built on Intel runners; on `macos-14` it would silently skip testing every arm64 wheel we now build. cibuildwheel already auto-skips tests for cross-compiled x86_64 wheels, so the remaining `*-macosx_universal2:arm64` entry is enough.

## Verification

Built the mypyc wheel natively on an M-series Mac: `macosx_26_0_arm64`, `COMPILED=True`, 157 tests pass.

The one failure (`test_methodlist_repr`) is pre-existing and unrelated — it's a `repr` line-wrapping issue on Python 3.14 that reproduces identically on a pure-Python install, and 3.14 isn't in the cibuildwheel build list anyway.

## Note

`linux/aarch64` and `windows/ARM64` are missing from the matrix for the same reason. Left out of this PR to keep it to the reported bug — happy to add them here or in a follow-up.


nstarman:
@wesselb — could we get this into a 2.10 bugfix release? Apple Silicon is currently the default Mac, and every user on it has been silently getting the pure-Python wheel since mypyc wheels were introduced, so the compiled speedup isn't reaching most Mac users at all.

The change is confined to CI config (matrix generation + `test-skip`); no library code is touched.

wesselb (approved):

wesselb:
@nstarman absolutely! Merging now.

PR #321, opened 2026-09-07
docs(mypyc): say what a `NativeBase` weak reference does not do

nstarman:
Follow-up to #319. Docstring only; no behaviour change.

## The problem

#319 added `NativeBase` so a `mypyc` native class gets `__dict__` and `__weakref__`. Its docstring says instances "can then be weakly referenced, which `jax.jit` needs (#318)" and stops there — which reads as *weak references work now*. They do not, and the way they fail is quiet.

The compiled deallocator omits `PyObject_ClearWeakRefs()`. A reference reads as cleared, but its **callback never runs**, so anything built on that callback — `weakref.WeakSet`, `WeakValueDictionary` — keeps a stale entry, and the dangling reference can segfault the interpreter at shutdown.

Measured on a `mypyc` wheel, holding a `Function` in a plain `WeakSet` and dropping it:

| | with an interpreted handle | plain `WeakSet` |
|---|---|---|
| `ref() is None` | True | True ← looks fine |
| callback fired | True | **False** |
| `len(WeakSet)` | 0 | **1** ← entry never dropped |
| `list(WeakSet)` | `[]` | `[]` |
| exit status | 0 | **139** ← segfault at teardown |

Reproduced 3/3 and 0/3 respectively, from a minimal case: create the object, put it in a `WeakSet`, drop it, collect. The crash lands *after* the script finishes.

## Upstream

This is [mypyc/mypyc#1102](https://github.com/mypyc/mypyc/issues/1102) — specifically the second of the two problems Jukka separates in that thread, "native subclasses of non-native classes can trigger segfaults when using weakrefs", and independently reported there by @Dreamsorcerer with the same `PyObject_ClearWeakRefs()` diagnosis. [python/mypy#19056](https://github.com/python/mypy/pull/19056) is the fix in progress.

Worth noting the interpreted-base pattern `NativeBase` uses is itself the workaround recommended in #1102 — so the caveat travels with it.

## Why it matters here

`jax.jit` is unaffected: it holds the reference itself rather than asking to be told when the referent dies, which is why #318 is fixed and stays fixed. But the next person to read `NativeBase` and reach for a `WeakSet` gets the silent version. Plum has one such place already, `_function._LiveFunctions`, which routes through an interpreted handle for exactly this reason.

The measurement above deliberately stays out of the docstring and lives with `_LiveFunctions`, the one class that exists because of this — one canonical account rather than two to drift apart. The docstring is four lines: the caveat, the cause, the consequence, the links.

### DIFF
--- a/src/plum/_mypyc.py
+++ b/src/plum/_mypyc.py
@@ -36,8 +36,9 @@ class NativeBase:
 
     Such a weak reference can be *taken*, but not relied on to notify: the compiled
     deallocator omits `PyObject_ClearWeakRefs()`, so its callback never runs and a
-    `weakref.WeakSet` or `weakref.WeakValueDictionary` keeps a stale entry that can segfault the
-    interpreter at shutdown. See mypyc/mypyc#1102; python/mypy#19056 is the fix.
+    `weakref.WeakSet` or `weakref.WeakValueDictionary` keeps a stale entry that can
+    segfault the interpreter at shutdown. See mypyc/mypyc#1102;
+    python/mypy#19056 is the fix.
 
     `jax.jit` is unaffected: it holds the reference itself rather than asking to be
     told when the referent dies.
--- a/src/plum/_mypyc.py
+++ b/src/plum/_mypyc.py
@@ -36,7 +36,7 @@ class NativeBase:
 
     Such a weak reference can be *taken*, but not relied on to notify: the compiled
     deallocator omits `PyObject_ClearWeakRefs()`, so its callback never runs and a
-    `WeakSet` or `WeakValueDictionary` keeps a stale entry that can segfault the
+    `weakref.WeakSet` or `weakref.WeakValueDictionary` keeps a stale entry that can segfault the
     interpreter at shutdown. See mypyc/mypyc#1102; python/mypy#19056 is the fix.
 
     `jax.jit` is unaffected: it holds the reference itself rather than asking to be
--- a/src/plum/_mypyc.py
+++ b/src/plum/_mypyc.py
@@ -33,4 +33,12 @@ class NativeBase:
     class outside the compile set gains both and stays native. Its instances can then be
     weakly referenced, which `jax.jit` needs (#318), and accept undeclared attributes,
     such as `__doc__` (#317) and those :func:`functools.wraps` copies.
+
+    Such a weak reference can be *taken*, but not relied on to notify: the compiled
+    deallocator omits `PyObject_ClearWeakRefs()`, so its callback never runs and a
+    `WeakSet` or `WeakValueDictionary` keeps a stale entry that can segfault the
+    interpreter at shutdown. See mypyc/mypyc#1102; python/mypy#19056 is the fix.
+
+    `jax.jit` is unaffected: it holds the reference itself rather than asking to be
+    told when the referent dies.
     """


src/plum/_mypyc.py
```diff
@@ -33,4 +33,13 @@ class NativeBase:
     class outside the compile set gains both and stays native. Its instances can then be
     weakly referenced, which `jax.jit` needs (#318), and accept undeclared attributes,
     such as `__doc__` (#317) and those :func:`functools.wraps` copies.
+
+    Such a weak reference can be *taken*, but not relied on to notify: the compiled
+    deallocator omits `PyObject_ClearWeakRefs()`, so its callback never runs and a
+    `weakref.WeakSet` or `weakref.WeakValueDictionary` keeps a stale entry that can
+    segfault the interpreter at shutdown. See mypyc/mypyc#1102;
+    python/mypy#19056 is the fix.
```
wesselb: Plainer wording.

```suggestion
    Weak references work, but their callbacks never run: a native class frees its
    instances without calling `PyObject_ClearWeakRefs()`. A `weakref.WeakSet` or
    `weakref.WeakValueDictionary` therefore keeps a stale entry, which can crash the
    interpreter at shutdown. See mypyc/mypyc#1102; python/mypy#19056 is the fix.
```

"Can be *taken*, but not relied on to notify" makes the reader work out what is being notified of what; the plain fact is that the callback never runs. "The compiled deallocator" assumes the reader knows `mypyc` generates one, and in "so its callback never runs" the "its" grammatically attaches to the deallocator rather than the weak reference. "Segfault" is fine but "crash" is the everyday word, and the closing line breaks a sentence early.

src/plum/_mypyc.py
```diff
@@ -33,4 +33,13 @@ class NativeBase:
...
+
+    `jax.jit` is unaffected: it holds the reference itself rather than asking to be
+    told when the referent dies.
```
wesselb: Simpler wording.

```suggestion
    `jax.jit` is unaffected: it does not rely on the callback.
```

"Asking to be told when the referent dies" is a roundabout way of saying "registering a callback", and "referent" is jargon. After the previous paragraph, "the callback" is understood.

wesselb (approved):
Thanks, @nstarman! This is helpful. :)

src/plum/_mypyc.py
nstarman: Applied in 1541b5c, verbatim. Agreed on all four — the fact goes first, the antecedent problem disappears with "the compiled deallocator", and the links now close a sentence instead of splitting one.

src/plum/_mypyc.py
nstarman: Applied in 1541b5c, verbatim. "The callback" is well established by the paragraph above it.

src/plum/_mypyc.py
```diff
@@ -33,4 +33,12 @@ class NativeBase:
     class outside the compile set gains both and stays native. Its instances can then be
     weakly referenced, which `jax.jit` needs (#318), and accept undeclared attributes,
     such as `__doc__` (#317) and those :func:`functools.wraps` copies.
+
+    Such a weak reference can be *taken*, but not relied on to notify: the compiled
+    deallocator omits `PyObject_ClearWeakRefs()`, so its callback never runs and a
+    `WeakSet` or `WeakValueDictionary` keeps a stale entry that can segfault the
+    interpreter at shutdown. See mypyc/mypyc#1102; python/mypy#19056 is the fix.
```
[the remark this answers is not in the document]
nstarman: Already the case on the reviewed commit and still so: both are written `weakref.WeakSet` and `weakref.WeakValueDictionary`.

src/plum/_mypyc.py
```diff
@@ -33,4 +33,12 @@ class NativeBase:
     class outside the compile set gains both and stays native. Its instances can then be
     weakly referenced, which `jax.jit` needs (#318), and accept undeclared attributes,
     such as `__doc__` (#317) and those :func:`functools.wraps` copies.
+
+    Such a weak reference can be *taken*, but not relied on to notify: the compiled
+    deallocator omits `PyObject_ClearWeakRefs()`, so its callback never runs and a
+    `weakref.WeakSet` or `weakref.WeakValueDictionary` keeps a stale entry that can segfault the
+    interpreter at shutdown. See mypyc/mypyc#1102; python/mypy#19056 is the fix.
```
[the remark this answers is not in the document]
nstarman: Fixed before the human review; 1541b5c keeps it so. The longest line in the docstring is 86 characters, and `ruff check` passes.

PR #324, opened 2026-09-10
Add integration test for beartype `TypeHint` contract

PhilipVinc:
Follow-up to #295 / #296, from the downstream side.

The existing suite pins the `Any` handling at the level of `_type_hint_eq` / `_type_hint_le` and for a two-method `Resolver`, but nothing exercises the public `@dispatch` path for the shape that actually broke in the wild: several concrete overloads plus a single `Any` fallback. This adds `tests/test_integration_beartype.py` covering that end to end (flat, nested, unannotated, multi-parameter), plus direct assertions on the `TypeHint` invariants Plum depends on — `__eq__` symmetry, equal-implies-hash-equal, and `Any` as a supertype of concrete hints. It deliberately does *not* assert on `TypeHint(Any) <= TypeHint(T)`, since `beartype>=0.23` changes that on purpose and `_substitute_any` already normalises it.

Motivation: Plum's whole suite passes against beartype 0.23.0rc0/rc1 — see beartype/beartype#682 — while that build still breaks real downstream dispatch. This is the repro that took out NetKet:

```python
@dispatch
def f(x: Number): ...
@dispatch
def f(x: Array): ...
@dispatch
def f(x: None): ...
@dispatch
def f(x: type): ...
@dispatch
def f(x: Any): ...     # fallback
f(1)  # AssertionError: The added method `Method(..., signature=Signature(typing.Any), ...)`
      #   is equal to 4 existing methods. This should never happen.
```

Verified matrix (`pytest tests/test_integration_beartype.py -q`):

| | beartype 0.22.8 / 0.22.9 | 0.23.0rc0 (PyPI) | 0.23.0rc1 (`main`, e3074dd) |
|---|---|---|---|
| plum `master` | 22 passed | 5 failed, 17 passed | 8 failed, 14 passed |
| plum at `c44e0cf^` (pre-#296) | 22 passed | — | 13 failed, 9 passed |

On `master` the behavioural tests stay green under the broken beartype (#296 does its job) and only the contract assertions light up, each with a message naming the invariant it lost. On the pre-#296 tree the behavioural tests fail too, which is the regression these lock down. Full suite green on beartype 0.22.8 and 0.22.9: 244 passed, 6 skipped, 2 xfailed.

**Heads-up on CI:** because `--pre beartype` currently resolves to 0.23.0rc0, the two `Test 3.13-pre-beartype` / `Test 3.14-pre-beartype` jobs will go red on this PR — asymmetric `TypeHint.__eq__` and equal-but-not-hash-equal wrappers. That is the alarm this PR is meant to raise (those jobs exist to catch exactly this), and it clears itself once beartype/beartype#682 lands. If you would rather not carry a red prerelease job until then, happy to wrap the two contract tests in an `xfail` keyed on `beartype.__version__`, or drop them and keep only the behavioural ones — just say which.


PhilipVinc:
Correction to the CI heads-up above: the `pre-beartype` jobs came back **green**, and the reason is worth flagging on its own — those two jobs are currently a no-op, and never actually run against a prerelease `beartype`.

From the `Test 3.14-pre-beartype` log on this PR:

```
08:55:44.647  uv pip install --upgrade --pre beartype
              - beartype==0.22.8
              + beartype==0.23.0rc0          <- prerelease installed
08:55:44.657  ##[group]Run uv run --frozen nox -s typecheck
08:55:45.220  Installed 13 packages in 117ms
              + beartype==0.22.8             <- reverted, before pytest ever starts
```

`--frozen` only means "don't update `uv.lock`"; `uv run` still syncs the environment to the lock, which pins `beartype==0.22.8`, so the `extra-install` step is undone by the very next step. That is why the whole suite passes on "prerelease beartype" while 0.23.0rc0 genuinely breaks downstream dispatch — the alarm was wired up, but the wire was cut.

Adding `--no-sync` to the two `uv run` steps (or `UV_NO_SYNC: 1` on the job) should be enough to make the prerelease stick. With that in place, this PR's contract tests fail on 0.23.0rc0 exactly as intended:

```
5 failed, 17 passed
FAILED test_beartype_typehint_equality_is_symmetric[list1]
FAILED test_beartype_typehint_equality_agrees_with_hash[int, Number, list, str]
```

I have deliberately left the CI change out of this PR to keep it to the one new test file — happy to add it here, or send it separately, whichever you prefer.

nstarman:
*Requested review, done by Claude.*

Checked the behavioural tests have real teeth: against pre-#296 plum (`c44e0cf^`) all four `@dispatch` tests plus the ordering one go red on rc1. Those are good to go. Three things on the contract half.

**1. The hash test can't fail.** `TypeHint` is memoized, so `x, y = TypeHint(hint), TypeHint(hint)` makes `x is y` and the first three asserts compare an object with itself. Only the trailing conditional has teeth, and it covers `hint` vs `Any` at the root only — so the file reports 22 passed on rc1 while rc1 (and `main`) still violate the contract:

```python
TypeHint(tuple[int, Any]) == TypeHint(tuple[int, str])  # True, hashes differ
TypeHint(tuple[Any])      == TypeHint(tuple[str])       # True, hashes differ
TypeHint(int | Any)       == TypeHint(int | str)        # True, hashes differ
```

Stating the contract over distinct pairs is shorter and catches them (8 pass on 0.22.8, 3 fail on rc1):

```python
_EQ_PAIRS = [
    (int, int), (list[int], list[int]), (Any, Any), (int, str), (list[Any], list[int]),
    (tuple[int, Any], tuple[int, str]), (tuple[Any], tuple[str]), (int | Any, int | str),
]

@pytest.mark.parametrize(("left", "right"), _EQ_PAIRS)
def test_equality_agrees_with_hash(left, right):
    if TypeHint(left) == TypeHint(right):
        assert hash(TypeHint(left)) == hash(TypeHint(right))
```

That reddens the pre-beartype jobs until upstream fixes tuple/union — which is what those jobs are for, but worth saying in the description.

**2. CI heads-up is stale.** `--pre` now resolves to rc1, so those jobs are green as submitted and the `xfail` isn't needed. The matrix was measured against beartype `main` at `e3074dd`, pre-`84f4ef3`; on released rc1 I get master 22 pass / pre-#296 5 fail. Worth a re-run so it doesn't read as "master is broken on rc1".

**3. Two dead lines.** `Signature(typing.Any) == Signature(Any)` is tautological (`from typing import Any`, so they're the same object) — different spelling intended? And `_ = TypeHint(Any) <= TypeHint(int)` discards its result; the comment above carries the meaning.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
+"""End-to-end tests for the `beartype` behaviour that Plum's resolver depends on.
+
+`tests/test_type.py` pins the `Any` handling at the level of `_type_hint_eq` and
+`_type_hint_le`, and `tests/test_resolver.py` pins it for a two-method `Resolver`.
+These tests instead exercise the whole public path - `@dispatch`, signature
+registration, method resolution, and the call cache - for the overload shape that
+broke in the wild: a set of concrete types with a single `Any` fallback.
```
wesselb, line 7: Terminology and wording.

```suggestion
These tests instead exercise the whole public path, from `@dispatch` to the call
cache, for several concrete methods with a single `Any` fallback.
```

Plum's term for these is "method", not "overload". The same applies to the first test's name and docstring, the comment inside it, and the name of the nested test.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+
+def test_any_fallback_in_large_overload_set(dispatch: plum.Dispatcher):
+    """Regression test for https://github.com/beartype/plum/issues/295.
+
+    An `Any` fallback must register as its own method next to every concrete type in
+    the set, and must only ever be selected for arguments that match nothing else.
+    When `Any` compares equal to the concrete types, all of these overloads collapse
+    onto one another and registration fails with an `AssertionError` from the
+    resolver. Taken from `netket.utils.numbers`, which this broke.
+    """
```
wesselb, line 29: Name and docstring.

```suggestion
def test_any_fallback(dispatch: plum.Dispatcher):
    """Regression test for https://github.com/beartype/plum/issues/295.

    An `Any` fallback must register as its own method next to every concrete method
    and must only be selected for arguments that match none of them.
    """
```

Plum calls these methods rather than overloads. How registration used to fail is history that the linked issue records, and the downstream module name is not something a reader of this test needs.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @dispatch
+    def f(x: type):
+        return "type"
+
+    @dispatch
+    def f(x: Any):
+        return "Any"
+
+    # Every concrete overload must survive alongside the fallback.
```
wesselb, line 51: Terminology.

```suggestion
    # Every concrete method must survive alongside the fallback.
```

See the comment on the module docstring.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+    def f(x: int):
+        return "int"
+
+    @dispatch
+    def f(x):
+        return "fallback"
+
+    assert len(f.methods) == 2
+    assert f(1) == "int"
+    assert f("x") == "fallback"
```
wesselb, line 82: Also cover redefining the fallback.

```suggestion
    assert len(f.methods) == 2
    assert f(1) == "int"
    assert f("x") == "fallback"

    # An explicit `Any` is the same signature, so it replaces the fallback.
    @dispatch
    def f(x: Any):
        return "updated fallback"

    assert len(f.methods) == 2
    assert f(1) == "int"
    assert f("x") == "updated fallback"
```

An unannotated parameter and an explicit `Any` must be one signature, so the second registration has to replace the fallback rather than add a third method, and calls made before it must not keep serving the old one.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def test_nested_any_fallback_in_overload_set(dispatch: plum.Dispatcher):
```
wesselb, line 85: Terminology.

```suggestion
def test_nested_any_fallback(dispatch: plum.Dispatcher):
```

See the comment on the module docstring; this also matches `test_any_fallback` above.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+
+@pytest.mark.parametrize("hint", [int, Number, list, list[int], str, type(None)])
+def test_beartype_typehint_equality_is_symmetric(hint):
+    """Plum relies on this contract from `beartype`.
+
+    Plum considers two signatures identical when each is a subhint of the other, so
+    an asymmetric `__eq__` on `TypeHint` makes "identical" depend on registration
+    order and silently drops methods. `Any` against a concrete type is the case that
+    regressed; see https://github.com/beartype/beartype/issues/682.
+    """
```
wesselb, line 135: Say precisely where symmetry matters.

```suggestion
    """`TypeHint.__eq__` must be symmetric.

    `Resolver.register` decides whether a new method replaces an existing one by
    evaluating `existing.signature == new.signature`, which compares the hints with
    `TypeHint.__eq__`. Were that asymmetric, whether two methods count as the same
    would depend on registration order, and one order would silently overwrite a
    method. Plum handles a root `Any` in `_type_hint_eq` itself, but `Any` against a
    concrete type is where symmetry broke upstream, so it is checked as well; see
    https://github.com/beartype/beartype/issues/682.
    """
```

"Each is a subhint of the other" describes how `beartype` implements `__eq__`, not anything Plum does, and "Plum relies on this contract" has not been true for `Any` since #296.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+@pytest.mark.parametrize("hint", [Any, int, Number, list[int], str])
+def test_beartype_typehint_equality_agrees_with_hash(hint):
+    """Plum relies on this contract from `beartype`.
+
+    Plum keys its signature bookkeeping on `TypeHint`, so wrappers that compare equal
+    must hash equally. A wrapper that is equal but not hash-equal breaks every
+    hash-based container the resolver puts it in.
+    """
```
wesselb, line 147: The docstring's claim does not hold.

```suggestion
    """`TypeHint.__eq__` must agree with `TypeHint.__hash__`.

    Plum hashes raw hints, never the wrappers, so the resolver does not depend on this
    directly. A wrapper that is equal but not hash-equal is, however, how `Any`
    collapsing onto every concrete hint first surfaced upstream, so it is a cheap
    early warning; see https://github.com/beartype/beartype/issues/682.
    """
```

`Signature.__hash__` and `Method.__hash__` hash the raw hints, and the cache in `_wrap_type_hint` is keyed on the raw hint too, so an equal-but-not-hash-equal wrapper cannot break a container in the resolver.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+    x, y = TypeHint(hint), TypeHint(hint)
+    assert x == y
+    assert y == x
+    assert hash(x) == hash(y)
+
+    # `Any` hashes differently from every concrete hint, so it must not compare equal
+    # to one either.
+    if hash(TypeHint(hint)) != hash(TypeHint(Any)):
+        assert TypeHint(hint) != TypeHint(Any)
+        assert TypeHint(Any) != TypeHint(hint)
```
wesselb, line 157: Assertion guarded by the property under test.

```suggestion
    # `Any` hashes differently from every concrete hint, so it must not compare equal
    # to one either.
    if hint is not Any:
        assert hash(x) != hash(TypeHint(Any))
        assert x != TypeHint(Any)
        assert TypeHint(Any) != x
```

The comment states the hash inequality as a fact, so assert it rather than branch on it. As written the branch is only ever skipped for `hint=Any`, which is what the new condition says directly.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+def test_beartype_orders_concrete_types_below_any(hint):
+    """Plum relies on this ordering from `beartype`.
+
+    Every concrete hint must be a subhint of `Any`. Plum normalises the *other*
+    direction itself, in `_substitute_any`, because `beartype>=0.23` additionally
+    makes `Any` a subhint of everything. If this assertion ever fails, `Any` has
+    stopped being a supertype at all, and the resolver's notion of "least specific"
+    goes with it.
+    """
+    assert TypeHint(hint) <= TypeHint(Any)
```
wesselb, line 170: Docstring does not match `_type_hint_le`; add the `object` ordering.

```suggestion
    """Every concrete hint must be below `object`, and `object` strictly below `Any`.

    The resolver relies only on the `object` ordering: `_substitute_any` rewrites a
    nested `Any` to `object`, and `_type_hint_le` handles a root `Any` itself.
    """
    assert TypeHint(hint) <= TypeHint(Any)
    assert TypeHint(hint) <= TypeHint(object)
    assert TypeHint(object) <= TypeHint(Any)
    assert not (TypeHint(Any) <= TypeHint(object))
```

`_type_hint_le(x, Any)` returns `True` without consulting `beartype`, and a nested `Any` becomes `object` before `beartype` sees it, so `TypeHint(hint) <= TypeHint(object)` is the ordering Plum actually uses. The last assertion is the one that `beartype>=0.23` changes on purpose. Assert it anyway: the `pre-beartype` jobs exist to show exactly that.

tests/test_integration_beartype.py
```diff
@@ -0,0 +1,198 @@
...
+def test_dispatch_is_independent_of_beartype_any_ordering(dispatch: plum.Dispatcher):
+    """Whichever way `beartype` orders `Any`, Plum's own ordering must not move.
+
+    `beartype<0.23` and `beartype>=0.23` disagree on `TypeHint(Any) <= TypeHint(int)`,
+    which is why Plum normalises `Any` rather than deferring to `beartype`. This test
+    reads the raw ordering without asserting on it, and pins Plum's.
+    """
+    # Read, but deliberately do not assert on, `beartype`'s raw answer: it is `False`
+    # before 0.23 and `True` from 0.23 onwards, and Plum must work either way.
+    _ = TypeHint(Any) <= TypeHint(int)
```
wesselb, line 182: Docstring, and what the assignment is for.

```suggestion
    """Plum's ordering of `Any` must not depend on `beartype`'s.

    `_type_hint_le` and `_type_hint_eq` handle `Any` themselves rather than deferring
    to `beartype`, whose answer changed in 0.23. Whichever answer `beartype` gives,
    `Any` must stay the unique least specific type in `Signature` comparison and in
    dispatch.
    """
    # `beartype` answers `False` before 0.23 and `True` from 0.23 onwards, so only
    # check that the comparison runs, not what it returns.
    _ = TypeHint(Any) <= TypeHint(int)
```

"Normalises" does not say where; the two helpers do. The comment on the assignment now says why the result is kept out of an assertion. Separately, `plum.Signature(typing.Any) == plum.Signature(Any)` further down compares `Any` with itself and could go, together with `import typing`.

wesselb (approved):
Thanks a lot, @PhilipVinc! Left a few comments. Happy to merge after those are resolved. :)

PR #326, opened 2026-09-14
fix(type): fall back to an opaque comparison for hints beartype can't wrap

nstarman:
Fixes #298.

`Signature` comparison wraps every type hint in `beartype.door.TypeHint` to get `Any`/subhint semantics right (#296). An object that looks class-like but isn't a real type or typing construct — e.g. a Sphinx `autodoc_mock_imports` placeholder — makes that constructor raise, uncaught, the moment a second `@dispatch` overload gets registered and compared against one carrying such a hint (`Resolver.register` → `Signature.__eq__` → `_type_hint_eq` → `_wrap_type_hint`).

`_wrap_type_hint_uncached` now catches that failure and falls back to `_OpaqueHint`, which compares equal only to an identical opaque hint and is a subhint of nothing but itself. No new warning is added — `Signature.__init__` already runs `is_faithful` on every type and warns there when a hint can't be classified, so the user is already told something's off before comparison ever runs.

## Test plan
- [x] New regression test (`test_wrap_type_hint_unwrappable_hint`) exercises `_type_hint_eq`/`_type_hint_le` with an unwrappable hint on either side, including the reflected `>=` path.
- [x] `pytest tests/ --ignore=tests/advanced` passes (189 passed; the one unrelated failure, `test_methodlist_repr`, is a pre-existing Python 3.14 repr flake).
- [x] `ruff check`, `ruff format --check`, `mypy` all clean.


bariod:
I modified `src/plum/_type.py` locally with these changes and my issue from #298 is fixed 🥳

src/plum/_type.py
```diff
@@ -329,24 +329,55 @@ def _substitute_any(hint: object, /) -> object:
...
 
 
-def _wrap_type_hint_uncached(hint: object, /) -> TypeHintWrapper:
-    return TypeHintWrapper(_substitute_any(hint))
+class _OpaqueHint:
+    """Fallback for a hint `beartype.door.TypeHint` refuses to wrap, e.g. a Sphinx
+    `autodoc_mock_imports` placeholder. Compares equal only to another opaque
+    wrapper around the same hint object, and is a subhint of nothing else. See GitHub
+    discussion #298.
+    """
```
wesselb, line 340: Docstring and `__eq__` disagree on equality.

```suggestion
class _OpaqueHint:
    """Fallback for a hint `beartype.door.TypeHint` refuses to wrap, e.g. a Sphinx
    `autodoc_mock_imports` placeholder. Compares equal only to another opaque
    wrapper around an equal hint, and is a subhint of nothing else. See GitHub
    discussion #298.

    Args:
        hint (object): Hint that `beartype.door.TypeHint` cannot wrap.
    """
```

The docstring says "the same hint object", but `__eq__` compares with `==`. I'd keep `==` and fix the docstring rather than the code: `beartype` also refuses real typing constructs such as `Callable[P, int]` with a `ParamSpec` or `Unpack[...]`, which compare equal by value, and `Signature.__eq__` compared hints with `==` before 2.5.8, so `==` restores the old behaviour for them. Identity could make two overloads spelled the same way look distinct. This also adds `Args:`, as the other classes in this module have.

src/plum/_type.py
```diff
@@ -329,24 +329,55 @@ def _substitute_any(hint: object, /) -> object:
...
+        return self == other
+
+
+def _wrap_type_hint_uncached(hint: object, /) -> TypeHintWrapper | _OpaqueHint:
+    # `Signature.__init__` already runs `is_faithful` on every type and warns there
+    # if `hint` can't be classified, so no second warning is needed here.
+    try:
+        return TypeHintWrapper(_substitute_any(hint))
+    except Exception:
+        return _OpaqueHint(hint)
```
wesselb, line 369: Warn when falling back.

```suggestion
def _wrap_type_hint_uncached(hint: object, /) -> TypeHintWrapper | _OpaqueHint:
    try:
        return TypeHintWrapper(_substitute_any(hint))
    except Exception:
        warnings.warn(
            f"Could not wrap the type hint `{hint}` for comparison. "
            f"I have fallen back to comparing it for equality only, so methods with "
            f"this type hint might not be ordered correctly. "
            f"Please open an issue at https://github.com/beartype/plum.",
            stacklevel=2,
        )
        return _OpaqueHint(hint)
```

`resolve_type_hint` and `is_faithful` warn where they give up, and this fallback should too: it silently changes how methods are ordered, which neither of the earlier warnings mentions, and `_type_hint_eq`/`_type_hint_le` are also reached from `_promotion` and `_parametric` without a `Signature` being built first. Since the wrapper is cached, the warning fires once per hashable hint.

src/plum/_type.py
```diff
@@ -329,24 +329,55 @@ def _substitute_any(hint: object, /) -> object:
...
+    object, is wrapped as an :class:`_OpaqueHint` instead of raising, since it is
+    still a hint that must be able to sit in a `Signature`.
 
     Args:
         hint (object): Already-resolved type hint.
 
     Returns:
-        TypeHintWrapper: Wrapped hint.
+        TypeHintWrapper | _OpaqueHint: Wrapped hint.
     """
```
wesselb, line 387: Duplicated documentation.

```suggestion
    """Wrap `hint` for comparison, replacing every `Any` by `object`.

    `Signature` comparison wraps the same fixed hints on every uncached dispatch, so
    the rewrite and `beartype`'s own wrapper lookup would otherwise repeat per call.

    Args:
        hint (object): Already-resolved type hint.

    Returns:
        TypeHintWrapper | _OpaqueHint: Wrapped hint, or an `_OpaqueHint` if `beartype`
            cannot wrap `hint`.
    """
```

The new paragraph repeats what the `_OpaqueHint` docstring already says (the Sphinx example, "instead of raising"); the fallback belongs in `Returns:`.

tests/test_type.py
```diff
@@ -370,3 +371,32 @@ def test_wrap_type_hint_unhashable_hint():
...
+    mock_b = _MockHint()
+
+    assert _wrap_type_hint(mock_a) == _OpaqueHint(mock_a)
+    assert _wrap_type_hint(mock_a) != _OpaqueHint(mock_b)
+
+    # Comparisons against an ordinary, wrappable hint must not raise either,
+    # regardless of which side is unwrappable.
+    assert not _type_hint_eq(mock_a, int)
+    assert not _type_hint_le(mock_a, int)
+    assert not _type_hint_le(int, mock_a)
```
wesselb: Simplify the stand-in and cover the positive case.

```suggestion
    hint_a = object()
    hint_b = object()

    with pytest.warns(Warning, match=r"(?i)could not wrap the type hint"):
        assert _wrap_type_hint(hint_a) == _OpaqueHint(hint_a)
    assert _wrap_type_hint(hint_a) != _OpaqueHint(hint_b)
    assert _type_hint_le(hint_a, hint_a)

    # Comparisons against an ordinary, wrappable hint must not raise either,
    # regardless of which side is unwrappable.
    assert not _type_hint_eq(hint_a, int)
    assert not _type_hint_le(hint_a, int)
    assert not _type_hint_le(int, hint_a)
```

The fallback catches every exception, so any object `beartype` rejects will do and the nested class can go. This also asserts that an opaque hint is a subhint of itself, which the class docstring promises but the test does not check, and that the fallback warns (see the comment on `_wrap_type_hint_uncached`; drop the `pytest.warns` lines if that is not adopted).

wesselb (approved):
Also thanks for this, @nstarman! This is great too. :) Happy to merge whenever you're happy too.

src/plum/_type.py
nstarman: Applied in 2fab74b, verbatim. Agreed on keeping `==`: confirmed `beartype.door.TypeHint` rejects `Callable[P, int]` with a `ParamSpec` (`BeartypeDoorPepUnsupportedException`), so real typing constructs do reach this fallback, not just mock objects.

src/plum/_type.py
nstarman: Applied in 2fab74b, verbatim — the wording matches `resolve_type_hint` and `is_faithful`. Confirmed it stays quiet across the rest of the suite: the full run still reports the one pre-existing `is_faithful` warning and nothing else.

src/plum/_type.py
nstarman: Applied in 2fab74b, verbatim.

tests/test_type.py
nstarman: Applied in 2fab74b, verbatim. `object()` is indeed rejected (`BeartypeDoorNonpepException`), so the mock class is gone. The `pytest.warns` block is in, since the warning was adopted; the later `_wrap_type_hint(hint_a)` calls are served from the `lru_cache` and so do not warn again. I also aligned the test docstring, which had the same identity-vs-equality wording you flagged on the class.

PR #327, opened 2026-09-15
docs(_type): correct the `Any` workaround rationale, and pin the `Union` case

nstarman:
## Context

[beartype@504602a](https://github.com/beartype/beartype/commit/504602a4cdb34a9bdd54bd5ce6ab515b11805e60) fixes the `TypeHint.__eq__`/`__hash__` inconsistency reported in beartype/beartype#687. I verified it against plum: **no plum behaviour needs to change**, but the comments explaining plum's `Any` workaround now point at a bug that no longer exists, and one real test gap turned up while checking.

## What I verified

Against `beartype` at 504602a (reports as `0.23.0rc2`), Python 3.13:

- **beartype/beartype#687 is fixed.** All three repro cases return consistent `eq`/`hash`; transitivity and set membership are restored, and it reaches further than the two containers filed there.
- **plum is green on it:** 206 passed, with only the 3 failures that fail identically on `beartype==0.22.9` in the same environment (2 autoreload, 1 `test_methodlist_repr` path-rendering artifact). No pin needed — [pyproject.toml](https://github.com/beartype/plum/blob/master/pyproject.toml) has a floor with no cap, and CI already runs `--pre beartype`.

I then tried deleting each piece of the workaround, to find what is still load-bearing:

| piece | removable? | what fails |
|---|---|---|
| `_substitute_any` | **no** | nested-`Any` ordering. `f(x: list[int])` + `f(x: list[Any])` → `f([1])` returns `"any"`; neither is strictly more specific, so registration order decides |
| root guard in `_type_hint_le` | **no** | `Any <= int` is still `True` upstream, by design |
| root guard in `_type_hint_eq` | **no** | `_substitute_any` rewrites root `Any` → `object`, so without the guard `_type_hint_eq(Any, object)` is `True` |

The third is why this PR exists. `beartype` reports `TypeHint(Any) != TypeHint(object)` perfectly well — it is plum's *own* rewrite that destroys the distinction, and the guard restores it. The docstring currently sends the reader to `_type_hint_le` for the reason, which is now the wrong reason entirely.

## The test gap

Replacing `_type_hint_eq` with a plain `TypeHint(x) == TypeHint(y)` left the suite **fully green** — which is not the same as the two being equivalent. A differential comparison over 1681 hint pairs found **12 disagreements, every one an `Any` inside a `Union`**:

```
x                  y                  plum   raw
int                Union[int,Any]     False  True
int                int|Any            False  True
Number             Union[int,Any]     False  True
Union[int,Any]     Union[int,object]  True   False
...
```

plum is right. `_substitute_any(Union[int, Any])` → `Union[int, object]`, which stays distinct from `int`; without the rewrite `Union[int, Any]` collapses onto every concrete type. With `_type_hint_eq` replaced by raw equality:

```python
@plum.dispatch
def f(x: int): return "int"
@plum.dispatch
def f(x: Union[int, Any]): return "anyish"

f(1)  # -> "anyish"    ← f(x: int) silently discarded as a "redefinition"
```

That is #296's bug 1 resurrected, and nothing in the suite caught it: `test_type_hint_eq_nested_any` covered `list`, `dict` and `tuple` but not `Union`, and the only `Union[int, Any]` in the suite asserts that the *rewrite* happens, never what it is for.

## What changed

- `_type_hint_le` — drop the stale "compare equal to" claim; note that this is a permanent difference in *relation* (`beartype` models assignability, where `Any` is deliberately both top and bottom, beartype/beartype#682; plum's specificity ordering needs `Any` purely as the top).
- `_type_hint_eq` — replace "See `_type_hint_le`" with what its guard actually defends against.
- `test_type_hint_eq_nested_any` — add the two `Union` assertions, and correct the docstring to name what actually fails without `_substitute_any`.

## Testing

Two assertions and comments; no behaviour change. 206 passed under **both** `beartype` 504602a and `0.22.9`. Confirmed the new assertions fail when `_type_hint_eq` is replaced by a plain `TypeHint` comparison — i.e. they now catch the regression that previously slipped through. `ruff check` / `ruff format --check` clean; pre-commit hooks pass.


src/plum/_type.py
```diff
@@ -360,14 +360,18 @@ def _wrap_type_hint(hint: object, /) -> TypeHintWrapper:
...
-    Since `beartype` 0.23, `is_subhint(Any, T)` is `True` for every `T`, which would
-    make an unannotated (`Any`-typed) parameter compare equal to, and as specific as,
-    any concrete type. See https://github.com/beartype/plum/issues/295. For Plum's
+    Since `beartype` 0.23, `is_subhint(Any, T)` is `True` for every `T`, which would make
+    an unannotated (`Any`-typed) parameter as specific as any concrete type. See
+    https://github.com/beartype/plum/issues/295. For Plum's
     signature bookkeeping, `Any` must instead be the unique least specific type. This
     check therefore differs from `beartype.door.TypeHint(x) <= TypeHint(y)` in exactly
     one respect: a root `Any` is only a subhint of `Any`, and a nested `Any` is
     rewritten to `object` by `_substitute_any`, which reproduces `beartype<0.23`.
```
wesselb: Line too long, and the paragraph wraps early.

```suggestion
    Since `beartype` 0.23, `is_subhint(Any, T)` is `True` for every `T`, which would
    make an unannotated (`Any`-typed) parameter as specific as any concrete type. See
    https://github.com/beartype/plum/issues/295. For Plum's signature bookkeeping, `Any`
    must instead be the unique least specific type. This check therefore differs from
    `beartype.door.TypeHint(x) <= TypeHint(y)` in exactly one respect: a root `Any` is
    only a subhint of `Any`, and a nested `Any` is rewritten to `object` by
    `_substitute_any`, which reproduces `beartype<0.23`.
```

The first line is 89 characters, which `ruff check` rejects (E501), so the pre-commit check in CI fails. The third line stops at column 60. This is the same text re-wrapped to the 88-column limit.

src/plum/_type.py
```diff
@@ -360,14 +360,18 @@ def _wrap_type_hint(hint: object, /) -> TypeHintWrapper:
...
 
+    Permanent, not a workaround for a bug: `beartype` intends `Any` to be both top and
+    bottom (beartype/beartype#682), Plum needs it purely top. Without the rewrite
+    neither `list[int]` nor `list[Any]` is more specific, so registration order wins.
```
wesselb: Wording and citation.

```suggestion
    The difference is permanent: `beartype` made `Any` a subhint of everything by
    design, see https://github.com/beartype/beartype/pull/616.
```

The paragraph above already says that `Any` must be the unique least specific type, so "top" and "bottom" restate that in lattice terms. The `list[int]` versus `list[Any]` sentence describes `_substitute_any`, whose docstring covers it. beartype/beartype#682 is an open bug report about `<`; the intent is in beartype/beartype#616. A full URL matches the rest of the code base and, unlike the shorthand, links from a docstring.

src/plum/_type.py
```diff
@@ -385,7 +389,9 @@ def _type_hint_le(x: object, y: object, /) -> bool:
 def _type_hint_eq(x: object, y: object, /) -> bool:
     """Check whether `x` and `y` are the same hint, where `Any` equals only itself.
 
-    See `_type_hint_le`.
+    The root-`Any` guard defends against Plum's own rewrite, not against `beartype`,
+    which compares a root `Any` correctly since beartype/beartype#687: `_substitute_any`
+    maps it to `object`, which would otherwise make the two equal.
```
wesselb: Implementation detail in the docstring, and the citation is off.

```suggestion
    See `_type_hint_le`.
```

The docstring describes the interface; why the guard exists belongs in a comment on the guard, e.g.

```python
    # `_substitute_any` maps a root `Any` to `object`, so `Any` must be compared here.
    if x is Any or y is Any:
```

`_type_hint_le` explains why `Any` must be the unique least specific type, which is the requirement behind this function's summary line, so the cross-reference stays. Also, `beartype` compares a root `Any` correctly since `0.23.0rc1`, before the change for beartype/beartype#687, which is about a nested `Any` in tuples and unions.

tests/test_type.py
```diff
@@ -325,15 +325,20 @@ def test_type_hint_le():
 def test_type_hint_eq_nested_any():
     """`Any` nested inside a parameterised hint must not collide either.
 
-    `beartype>=0.23` collapses `list[Any]` onto `list[int]` exactly as it collapses
-    `Any` onto `int`, so the rewrite in `_substitute_any` has to reach every
-    level, not just the root.
+    `beartype` compares a nested `Any` correctly since beartype/beartype#687, except
+    inside a `Union`, where `Union[int, Any]` still collapses onto every concrete type.
+    The `Union` cases below and every `_type_hint_le` case are what fail without
+    `_substitute_any`.
```
wesselb: The docstring is more specific than the implementation.

```suggestion
    The rewrite in `_substitute_any` has to reach every level, not just the root.
```

Which hint shapes `beartype` compares correctly changes between its releases, and this paragraph went stale once already. The claim is also not quite right: on `beartype@504602a`, `TypeHint(str) == TypeHint(Union[int, Any])` is `False`, so the collapse is onto supertypes of `int`, not every concrete type. State what Plum requires and leave `beartype`'s state to its tracker.

tests/test_type.py
```diff
@@ -325,15 +325,20 @@ def test_type_hint_le():
...
     """
     assert not _type_hint_eq(list[Any], list[int])
     assert not _type_hint_eq(dict[str, Any], dict[str, int])
     assert not _type_hint_eq(list[list[Any]], list[list[int]])
     assert not _type_hint_eq(tuple[int, Any], tuple[int, str])
 
+    # `Any` inside a `Union` must not collapse the `Union` onto a concrete type.
+    assert not _type_hint_eq(int, Union[int, Any])  # noqa: UP007
+    assert not _type_hint_eq(int, int | Any)
```
wesselb, line 336: One case is enough.

```suggestion
    # `Any` inside a `Union` must not collapse the `Union` onto a concrete type.
    assert not _type_hint_eq(int, int | Any)
```

`test_substitute_any_rebuilds_hints` already checks that both `Union[int, Any]` and `int | Any` are rewritten, so the second spelling retests `_substitute_any` through `_type_hint_eq`.

wesselb (approved):
Thanks, @nstarman! :) A few minor suggestions. Take or leave as you see fit. Happy to merge whenever you're happy too.

src/plum/_type.py
nstarman: Applied in a4a6082, verbatim. Confirmed: that line was 89 characters. `ruff check` and `ruff format --check` are clean now.

src/plum/_type.py
nstarman: Applied in a4a6082, verbatim.

src/plum/_type.py
nstarman: Applied in a4a6082, verbatim — docstring back to `See \`_type_hint_le\`.`, with your comment on the guard itself.

tests/test_type.py
nstarman: Applied in a4a6082, verbatim. Your correction checks out: on `beartype==0.23.0rc1`, `TypeHint(str) == TypeHint(Union[int, Any])` is `False`, so the collapse is onto supertypes of `int`, not every concrete type. Stating Plum's requirement instead is the right call.

tests/test_type.py
nstarman: Applied in a4a6082, verbatim. Checked before dropping it, since this is the one suggestion that removes an assertion — on `beartype==0.23.0rc1` / Python 3.13 the two spellings are indistinguishable here: `Union[int, Any] == int | Any` at the typing level, and a raw `TypeHint(int) ==` comparison returns `True` for both. The retained assertion still bites: raw says `True`, plum says `False`.

PR #328, opened 2026-09-16
Fix wrapped signatures in generated documentation on Python 3.13+

### LINKED ISSUE #276: `_document()` breaks RST docstrings on Python 3.13+ due to multi-line signature formatting from `pydoc`
The title and text of this issue were written by Claude Sonnet 4.6 while I attempted to debug why the Sphinx documentation build for a package I maintain failed when I changed Python version from 3.9 to 3.13 -- see CI runs at https://github.com/Qiskit/qiskit-addon-aqc-tensor/pull/178.  The model correctly identified that 3.12 is the latest release where the build would work without code changes.  I asked it to create a summary so I could open this issue.

---

When building Sphinx documentation for a project that uses plum-dispatch, the following RST warnings/errors are raised for `@dispatch.abstract` functions on Python 3.13+, but not on Python 3.12:

```
WARNING: Inline emphasis start-string without end-string. [docutils]
ERROR: Unexpected indentation. [docutils]
```

## Root cause

CPython PR [#112143](https://github.com/python/cpython/pull/112143) (merged Dec 2023, released in Python 3.13) changed `pydoc._getargspec()` to call `signature.format(max_width=80 - len(name))`. For functions with longer signatures, this wraps parameters across multiple lines.

`plum.resolver._document()` calls `pydoc._PlainTextDoc().document(f)` and then splits the result on `"\n"`, treating `parts[0]` as the title (the function declaration) and `parts[1:]` as the body. On Python 3.12 and earlier, the signature was always a single line so `parts[0]` was the complete declaration. On Python 3.13+, `parts[0]` is just the opening `funcname(`, and the remaining parameter lines end up in the body — mixed in with the actual docstring text. Any `*` characters in those parameter lines (e.g., from a keyword-only argument separator `*,` or `*args`) are then interpreted by docutils as RST emphasis markup.

## Minimal reproduction

Any `@dispatch` function with a signature long enough to wrap past 80 characters, containing keyword-only arguments (the `*,` separator), documented with Sphinx on Python 3.13+.

## Workaround

Set the environment variable `PLUM_SIMPLE_DOC=1` during the Sphinx build (suppresses appending overload signatures entirely) or pin to Python ≤ 3.12 for docs builds.

## Suggested fix

In `_document()`, after calling `pydoc._PlainTextDoc().document(f)`, join lines until the first one that ends with `)` or `) -> ...` before splitting into title vs. body — or use `inspect.signature(f)` directly to build the title rather than relying on `pydoc`.

xuu33030:
## Summary

Closes #276.

Python 3.13+ wraps long signatures in `pydoc` output. `_document()` assumed that the first line contained the whole signature, leaving parameter lines in the docstring body and producing invalid reStructuredText for Sphinx.

- Rejoin a wrapped signature through its unindented closing `)` line before separating the docstring body.
- Keep the existing `pydoc` rendering, including return annotations, and leave short signatures unchanged.
- Add regression coverage for Sphinx/plain output, with/without a docstring, and defaults containing parentheses and an arrow (which must not be mistaken for the signature boundary).

## Verification

On macOS, using the repository's locked dependencies:

- Python 3.13.5: the four new regression cases failed before the fix; `pytest -q tests/test_resolver.py` passes after it (19 passed).
- Python 3.13.5 and Python 3.10.20: `pytest -q` passes on each (434 passed, 63 skipped, 2 xfailed, 1 warning). Skips include version-specific union tests and optional documentation examples; the warning is from `test_is_faithful`.
- A temporary Sphinx 7.4.7 project using `sphinx.ext.autodoc` and a long `@dispatch.abstract`/`@dispatch` signature builds with `python -m sphinx -b html -W --keep-going`; the generated HTML contains the parameter and implementation documentation.
- `mypy --enable-error-code=no-redef src/plum tests/static`: passes.
- `pyright tests/static`: 0 errors, 0 warnings.
- Ruff 0.7.3 check and format check on both changed files: pass.
- `git diff --check`: passes.

The compiled mypyc wheel and the remaining CI Python/platform matrix were not run locally.

## AI assistance

This contribution was prepared and tested by an AI coding agent at the account owner's request. No human review is claimed.


src/plum/_resolver.py
```diff
@@ -194,6 +194,14 @@ def _document(f: Callable[..., object], f_name: str | None = None, /) -> str:
     # Separate out the function definition and the lines corresponding to the body.
     title = parts[0]
     body = parts[1:]
+    # Python 3.13+ wraps long signatures. Only the closing signature line is
+    # unindented; parentheses or arrows inside parameter defaults are not delimiters.
```
wesselb, line 198: Comment wording.

```suggestion
    body = parts[1:]

    # Python 3.13+ wraps long signatures over multiple lines: the parameters are
    # indented and the closing parenthesis starts the first unindented line. Join
    # these lines back into the title.
```

tests/test_resolver.py
```diff
@@ -91,6 +91,41 @@ def f(x):
...
+        "f(first_parameter: int, second_parameter: str, *, "
+        "keyword_parameter: tuple = ('a)', ' -> ')) -> bool"
+    )
+    if sphinx:
+        expected = ".. py:function:: " + signature + "\n   :noindex:"
+    else:
+        expected = "<separator>\n\n" + signature
+    if with_docstring:
+        expected += "\n\nSummary.\n\nMore details."
+    assert _document(f) == expected
```
wesselb, line 126: Test economy.

```suggestion
def test_document_wrapped_signature():
    """Test that a signature which `pydoc` wraps over multiple lines is rejoined."""

    # The signature must be long enough for `pydoc` to wrap it. The defaults contain
    # `)` and `->`, which must not be mistaken for the end of the signature.
    def f(
        first_parameter: int,
        second_parameter: str,
        *,
        keyword_parameter: tuple = ("a)", " -> "),
    ) -> bool:
        """Title.

        Hello.
        """

    signature = (
        "f(first_parameter: int, second_parameter: str, *, "
        "keyword_parameter: tuple = ('a)', ' -> ')) -> bool"
    )
    assert _document(f, "f") == f"<separator>\n\n{signature}\n\nTitle.\n\nHello."
```

The Sphinx prefix is independent of rejoining the signature and is already covered by the two tests above, and whether `f` has a docstring only changes whether the body is empty, which the rejoining does not touch. One case with a docstring is the one that reproduces #276 and is enough. The docstring and comment record why the signature is this long and why the defaults look odd; without them a future edit could shorten the names and quietly make the test vacuous on 3.13+.

wesselb (approved):
This looks great, @xuu33030! Thanks for the PR. :) I'm happy to have this merged after the comments. Perhaps @nstarman still wants to have a look?

src/plum/_resolver.py
```diff
@@ -194,6 +194,14 @@ def _document(f: Callable[..., object], f_name: str | None = None, /) -> str:
...
+    if title.endswith("("):
+        for i, line in enumerate(body):
+            if line.startswith(")"):
+                title += " ".join(part.strip() for part in body[:i]) + line
+                body = body[i + 1 :]
+                break
```
nstarman, line 204:
```suggestion
    if title.endswith("("):
        end = next(i for i, line in enumerate(body) if line.startswith(")"))
        title += " ".join(line.strip() for line in body[:end]) + body[end]
        body = body[end + 1 :]
```

nstarman: But I think there's a larger change can be made... See review.

nstarman:
*Posted by Claude on behalf of Nathaniel, at his request.*

---

The fix is correct, but the loop can go away entirely — along with the whole "re-parse `pydoc`'s wrapped output" step.

### Why the wrapping happens

`pydoc.TextDoc.docroutine` (3.13+) calls `pydoc._getargspec`, which does:

```python
max_width = (80 - len(name)) if name != '<lambda>' else None
return signature.format(max_width=max_width)
```

That `max_width` is the *only* source of the wrapping. `inspect.Signature.__str__` is `self.format()` with `max_width=None`, so it never wraps, on any version. Rather than un-wrapping `pydoc`'s output, we can avoid asking for the wrapping in the first place.

Relatedly, `inspect.getdoc(f)` already returns the dedented docstring — which is exactly what the `[line[4:] for line in body]` step below is undoing.

### Alternative

Net −15 lines vs. `master`, −25 vs. this PR:

```diff
-import pydoc
+import inspect
@@
-    if f_name is not None and getattr(f, "__name__", None) != f_name:
-        f = _change_function_name(f, f_name)
-    parts = pydoc._PlainTextDoc().document(f).rstrip().split("\n")
-    title = parts[0]
-    body = parts[1:]
-    body = [line[4:] for line in body]
+    # `inspect.signature` renders the signature on a single line, whereas `pydoc`
+    # wraps long ones since Python 3.13, breaking the reStructuredText below.
+    try:
+        title = (f_name or f.__name__) + str(inspect.signature(f))
+    except (TypeError, ValueError):  # pragma: no cover
+        title = (f_name or f.__name__) + "(...)"
+    if inspect.iscoroutinefunction(f) or inspect.isasyncgenfunction(f):
+        title = "async " + title
+    body = (inspect.getdoc(f) or "").rstrip().split("\n")
```

This removes the line split, the four-space dedent assumption, the new loop, and the `pydoc._PlainTextDoc` private-API access. `_change_function_name` also becomes dead code — only its own test in `tests/test_function.py` still references it — so that is another ~22 lines that could go.

**Verified:** applied on top of this branch, `pytest tests/` passes on 3.10, 3.12, 3.13 and 3.14, including the four new regression cases added here (the tuple default containing `"a)"` and `" -> "`). The one failure, `test_method.py::test_methodlist_repr`, is terminal-width-dependent and fails identically on the unmodified branch.

**Behavior deltas**, none of them reachable from `Resolver`, which only documents registered implementations: `pydoc`'s `" method of X instance"` / `" class method of X"` notes are dropped for bound methods, and a lambda renders as `<lambda>(x, y)` rather than `<lambda> lambda x, y`. The `async` prefix is preserved by the two lines above.

### If a minimal diff is preferred

Keeping this PR's approach, the manual index loop can at least become:

```python
if title.endswith("("):
    end = next(i for i, line in enumerate(body) if line.startswith(")"))
    title += " ".join(line.strip() for line in body[:end]) + body[end]
    body = body[end + 1 :]
```

That is cosmetic, though — it still depends on `pydoc`'s line layout and on "unindented `)`" remaining the delimiter across CPython versions. The `inspect` version deletes the fragility instead of tidying it.

xuu33030:
Thanks for the detailed alternative. Before I update the branch, would you prefer the broader `inspect.signature` / `inspect.getdoc` approach, or should I keep this PR focused and apply the smaller loop and test simplifications?

